Issue #64, opened 2020-12-31
Draft reimplementation of several GenoCAT examples


sehilyi:

sehilyi:
- Circos: http://circos.ca/intro/genomic_data/

![](http://circos.ca/intro/genomic_data/img/circos-conde-nast-large.png)

- Fuji plot (Circos): https://github.com/mkanai/fujiplot

![](https://camo.githubusercontent.com/581a3ef7b62d2b5c951e09bb6f8b09ebd79015295edc3759358b52f040bd5d4e/687474703a2f2f6d6b616e61692e6769746875622e696f2f696d672f4b616e6169323031385f466967312e737667)

- Other circular examples:
https://jokergoo.github.io/circlize_examples/

- CNVkit: https://cnvkit.readthedocs.io/en/stable/plots.html 

![](http://genocat.tools/assets/CNV.png)

- GIVE: http://genocat.tools/tools/give.html

![](https://media.springernature.com/full/springer-static/image/art%3A10.1186%2Fs13059-018-1465-6/MediaObjects/13059_2018_1465_Fig1_HTML.png)

- Oviz-Bio: http://genocat.tools/tools/oviz-bio.html

![](http://genocat.tools/assets/ovizbio.png)

- SWAV: http://swav.popgenetics.net/ 

![](http://genocat.tools/assets/swav.png)

sehilyi:
Eventually, use aggregated data for the performance.

sehilyi:
Some tools presented at TVCG.

- Gremlin: http://compbio.cs.brown.edu/projects/gremlin/
   - O'Brien et al. Gremlin: an interactive visualization model for analyzing genomic rearrangements. TVCG 2010

![Screen Shot 2021-01-07 at 12 11 16 PM](https://user-images.githubusercontent.com/9922882/103922345-9ace3100-50e1-11eb-8f19-e5b9d907c5cf.png)

- MizBee: http://www.cs.utah.edu/~miriah/mizbee/Overview.html
   - Data: http://www.cs.utah.edu/~miriah/mizbee/Executables_files/fish.txt
   - Meyer et al. MizBee: a multiscale synteny browser. TVCG 2019

![Screen Shot 2021-01-07 at 12 16 06 PM](https://user-images.githubusercontent.com/9922882/103922714-23e56800-50e2-11eb-9a1a-3631d110ec17.png)

- Variant View: https://www.cs.ubc.ca/labs/imager/tr/2013/VariantView/
[...]

PR #103, opened 2021-01-14
refactor: add a fuji plot

sehilyi:
Toward #64


sehilyi:
Closing this since we already have several circular examples. We can add this example later.

Issue #136, opened 2021-01-28
Allow multiple conditions in `visibility`


sehilyi:
Since the width of bars become large enough to show text labels when zoomed in too fast, texts are temporally being shown during zooming.

Should be able for users to use `zoomLevel` along with `width` measure in a single `visibility` option so that text can be entirely hidden when zoomed out.

![Screen Shot 2021-01-28 at 3 18 23 PM](https://user-images.githubusercontent.com/9922882/106194685-26852b80-617d-11eb-8304-34b3ddba4b92.png)

PR #170, opened 2021-02-02
docs: update README.md

### LINKED ISSUE #155: Update README.md

wangqianwen0418:
@sehilyi , any comments are welcome!

Also, we might need to store the logos (with/without names, different ratios) in one place.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,9 +41,9 @@ Using Gosling.js, users can easily create interactive and scalable genomic visua
 </pre>
 
 </td>
-<td>
+<td align="center">
 <img src="https://raw.githubusercontent.com/wiki/gosling-lang/gosling.js/images/demo.gif"  width="400"/>
-<center><a href="https://gosling.js.org/">Try Online</></center>
+<a href="https://gosling.js.org/">Try Online</>
 </td>
 </tr>
 </table>
--- a/README.md
+++ b/README.md
@@ -13,25 +13,27 @@
 <br/>
 <br/>
 
-# Introduction
+## Introduction
 
-Gosling is a declarative visualization grammar tailored for interactive genomic visualizations.   
-In Gosling, users can easily create interactive and scalable visualizations through writing a JSON configuration. 
+Gosling is a declarative visualization grammar tailored for interactive genomic visualizations. 
+Using Gosling.js, users can easily create interactive and scalable genomic visualizations through writing a JSON configuration. 
 
     
 ```json
-{"tracks": [{
-      "data": {
-        "url": "https://resgen.io/api/v1/tileset_info/?d=UvVPeLHuRDiYA3qwFlm7xQ",
-        "type": "tileset"
-      },
-      "metadata": {
-        "type": "higlass-multivec",
-        "row": "sample",
-        "column": "position",
-        "value": "peak",
-        "categories": ["sample 1", "sample 2", "sample 3", "sample 4"]
-      }
+{
+    // Simplest example with only data configuration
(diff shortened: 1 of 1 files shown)


sehilyi (commented):
Looks good, cool to see the GIF image! 

Since we are expecting to have more examples and resources very soon, let's keep this PR here for a while.

sehilyi:
@wangqianwen0418, I've updated README.md just a bit for the readability (e.g., used table to show example code and GIF, restructured little). I also removed our logo for now as it doesn't look good in the GitHub dark mode as you had mentioned before.

But I think it is ready to go, and I think we will be able to update this a bit more before v0.7.

Please feel free to give feedback on the my update and/or merge this into master branch.

sehilyi (approved):

sehilyi:
And yes, we will need to put png files in diverse sizes so that we can use that inside our editor and in other applications.

wangqianwen0418:
@sehilyi , many thanks!

I have one concern about the table layout. I like the idea and have tried before.
But it is really hard to modify the width of the table cell. (Because cell are inline-display. I have also tried to change the style to block-display but it doesn't work)
Below is a screenshot from the updateReadme branch.
image

wangqianwen0418:
For the logo, I think we can use the one I made for now? 
It has a white background so it will be okay (even though not prefect) for the dark mode.

sehilyi:
I also confronted that issue, but I guess I underestimated it by assuming that people will likely see the `README.md` file more frequently on their desktops/laptops rather than mobile devices (which is not necessarily true).

We could reduce the width of the left code by removing the data URL and adding new lines:

Screen Shot 2021-02-03 at 2 25 56 PM

But the issue still exists when opening on the mobile (i.e., with narrower screen).

I used the table just because width of the code and the GIF is relatively small, and arranging them vertically would make the README.md needlessly longer.

However, I am on the fence now considering the mobile issue, so it's your call!

(We could later put another and more wider example that better represent the strength of our grammar.)

sehilyi:
For the logo, I personally think that white rectangular boundary of the logo and black background of the document in a dark mode is making it not looking great. And I also prefer to have the title as a pure text.

wangqianwen0418:
Cool, I like the way you modify the code part to reduce the width. 
Let's use the table layout!

wangqianwen0418:
I am working on the data configuration part, so the example needs to update after that. I will merge the readme when the new example is ready.

wangqianwen0418:
https://github.community/t/support-theme-context-for-images-in-light-vs-dark-mode/147981
Seems there is not a good way to detect theme mode now

sehilyi:
When you update the examples for `data/metadata` in the docs, could you please also update `visibility` options? Recently, I changed its type from `object` to `array of object` (#136), and if there is any example in docs that contains `visibility` options, it would not work properly. You just need to change `visibility: { ... }` to `visibility: [{ ... }]`.

https://github.com/gosling-lang/gosling.js/blob/9d166be53d848c318e58beada2aef910a1f13b19/src/core/gosling.schema.ts#L221

wangqianwen0418:
> When you update the examples for `data/metadata` in the docs, could you please also update `visibility` options? Recently, I changed its type from `object` to `array of object` (#136), and if there is any example in docs that contains `visibility` options, it would not work properly. You just need to change `visibility: { ... }` to `visibility: [{ ... }]`.
> 

I am so sorry for missing this message! Please feel free to ping me at slack if something similar happen in the future

Issue #50, opened 2020-12-25
Improve rendering performance


sehilyi:
### Some Initial Thoughts
- Using HiGlass' data structure largely allows us to render visualization w/o too much cost because it pre-aggregates data and send Gosling the manageable amount of data values to display. But, still, improving rendering performance is really important (e.g., using many number of tracks w/ non-higlass data will significantly degrade the performance).
- Performance can be largely improved If we render texture (sprite), instead of individual graphical elements, and stretch them upon zooming in/out until getting new tiles. Some challenges:
  - Not straightforward to apply this concept in circular layouts
  - Some marks (text and circles) would result in ugly or even unreadable appearance
  - We still need to update subset of graphical elements if mark-level visibility options are being used (i.e., `visibility: [{..., target: "mark" }]`)
- Sharing the same texture for identical-looking visual marks would be another option ([Example](https://github.com/e-/ScatterplotAnime/blob/68ac4959f7dfa88c0f5c26409e3678409bb2b749/1webgl.js#L37))
  - e.g., use the same texture for the same color `bars` and just resize them.
- Linking multiple higlass views looks to significantly decrease the zooming/panning performance
- This rendering performance is especially important for dense matrix visualization (#38)
- A constant/related problem that we encounter: https://github.com/gosling-lang/gosling.js/issues/230

### Resources
- web worker
- SpriteBatches: http://rbwhitaker.wikidot.com/spritebatch-basics
- https://mattdesl.svbtle.com/drawing-lines-is-hard

sehilyi:
Especially, linking multiple tracks seem to significantly decrease the zooming/panning performance.

sehilyi:
First, we could partly use `Sprite` to stretch tiles along x and y axes, rather than re-rendering upon zoom/pan.

sehilyi:
Using `sprite` would be the simplest try. For example, we avoid error messages shown in the bottom of #175.

sehilyi:
Another warning message related to performance:
<img width="408" alt="Screen Shot 2021-02-10 at 7 55 14 PM" src="https://user-images.githubusercontent.com/9922882/107592053-f5274980-6bd9-11eb-9029-8d61b6842847.png">

etowahadams:
[...]

PR #175, opened 2021-02-04
feat: add a data preview panel in the editor

### LINKED ISSUE #144: Add a data table in the editor

sehilyi:
Fix #144 

Screen Shot 2021-02-03 at 10 25 40 PM


sehilyi:
@wangqianwen0418 Can you review whether a newly add data preview panel properly works in your computer and as you expected?

wangqianwen0418:
@sehilyi , the data table looks great!
I only have a few small comments 
- Maybe we can change from "tileset" to "bed",  "vector", or "multivec"?
- do you think we can also specify the tracks using the data? something like a track id
- Sometimes, the data table has no data. The problem is resolved after refreshing the page. I am not sure whether this is caused by the internet connection or my computer memory
- maybe adding a visual cue to tell people about the existence of this table?

sehilyi:
@wangqianwen0418 Thank you for reviewing this and sharing feedback, this is very helpful!

> Maybe we can change from "tileset" to "bed", "vector", or "multivec"?

This is a good suggestion, I will update this!

> do you think we can also specify the tracks using the data? something like a track id

We can specify the `track id`, but it will be unreadable and arbitrary strings. I think allowing users to easily map between individual tracks and data table would be really useful, but it would take some more time to implement. I will add this as a separate issue.

> Sometimes, the data table has no data. The problem is resolved after refreshing the page. I am not sure whether this is caused by the internet connection or my computer memory

This is odd, I will investigate more.

> maybe adding a visual cue to tell people about the existence of this table?

Good idea, perhaps we can put a hidden (green) circle by the title (i.e., "Data Preview") and show it when data is being updated. Or, do you have any suggestion?

wangqianwen0418:
image

Not sure whether missing value in the table is related to this error

sehilyi:
I think it is related to the visualization part, can you tell me which example resulted in the error?

#50

wangqianwen0418:
Sometimes, the basic-linking-views has this problem

wangqianwen0418:
image

Like this, but the problem not always happen

sehilyi:
@wangqianwen0418 Thanks a lot for sharing this bug. 

I will need to investigate more, but I think rendering too many PIXI elements sometimes making such errors.

PR #178, opened 2021-02-04
feat: support genome builds

### LINKED ISSUE #141: Support genome builds
> That is important, in particular for the matrix view, which might use two different genomes/genome builds.

We have an option for this, but not supported yet:
https://github.com/gosling-lang/gosling.js/blob/9606e6bd783f69b6d1412aa9a0ce0e5dc71dc55e/src/core/gosling.schema.ts#L7

We will need to add this to track-level

sehilyi:
Fix #141 

Supported:
- hg38
- hg19
- hg17
- hg16
- mm10
- mm9

### DIFF
--- a/schema/gosling.schema.json
+++ b/schema/gosling.schema.json
@@ -81,6 +81,18 @@
       },
       "type": "object"
     },
+    "Assembly": {
+      "enum": [
+        "hg38",
+        "hg19",
+        "hg18",
+        "hg17",
+        "hg16",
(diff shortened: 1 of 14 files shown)


sehilyi:
@wangqianwen0418 Would you be able to add this into our docs?

wangqianwen0418:
Sure! I will update the doc later today

sehilyi:
Thanks!

PR #263, opened 2021-02-26
feat: loading specs from gist

flekschas:
Allows loading specs from GitHub Gist

1. To load a spec you first have to create a gist with a file named `gosling.js*` that specifies the spec.
2. You can additionally specify a `readme.md` file to describe your spec
3. Also be sure to give your gist a fabulous title. It'll be shown in the gosling editor

Finally open your data as follows:

http://localhost:8080/?gist=flekschas/e6e388332814886d4d714efd0e88093b

The `gist` param needs to follow the `<username>/<id>` pattern.


sehilyi (approved):
Thank you for this PR, I think this will be a really awesome feature!

There were some compatibility issues between the latest `react-monaco-editor` and our code, so I updated the code a little bit to make the editor fully functioning (e.g., semantic highlighting, keyword auto-completion, finding keywords).

As a minor update, I made the description panel float on top of the main interface (making it more like a side/extra view), so resizing it does not affect the size of the left (editor) and right (visualization) panels.

I think the gist works really well, and we are ready to merge this.

sehilyi:
Oh, it would be really helpful if you could try this again locally. I updated the `package.json` and wonder if the error messages related to `react-monaco-editor` appear again from your side.

flekschas:
Works for me! Maybe we should update the `README.md` at some point to tell people to exclusively use Yarn given that there seem to be some difference between npm and yarn.

Not sure I prefer the _floating-ness_ of the content. I think I would either like to see both in full (to be able to understand the visualization as I read the text) or hide the text all together. However, I see the point in treating the textual description as a supplement and the _floating-ness_ supports that. In general, I do not feel strongly about this, so pick whichever version you like more :)

sehilyi:
Thank you for checking it!

I think it is a good idea to clarify that we exclusively support `yarn`. Reflecting it, I updated `README.md` and also removed `package-lock.json`.

My main rationale was to have only one vertical separator to keep the interface and interaction straightforward until we find an optimal layout. I feel it is sometimes difficult to interactive with multiple (more than two) views juxtaposed along a single direction since resizing or maximizing views may not work as how I wanted/expected.

I am not sure how difficult it would be to implement, but since we currently have many panels that people may want to see at once (i.e., Gosling spec, visualization, data preview, HiGlass viewConfig, textual description), perhaps having flexibility in changing the position of panels (e.g., using drag-and-drop) and/or changing the layout of panels would be useful, just like in many tab-based interfaces (e.g., VS Code).

PR #296, opened 2021-03-02
refactor: better names for triangle

wangqianwen0418:
change `triangle-r` to `triangleRight`, `triangle-l` to `triangleLeft`,  `triangle-d` to `triangleBottom`

### DIFF
--- a/src/editor/editor.css
+++ b/src/editor/editor.css
@@ -65,7 +65,7 @@
     outline: none;
 }
 
-/* Drop down menu's triangle-down icon */
+/* Drop down menu's triangleBottomown icon */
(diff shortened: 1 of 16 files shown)


src/editor/editor.css
```diff
@@ -65,7 +65,7 @@
     outline: none;
 }
 
-/* Drop down menu's triangle-down icon */
```
sehilyi: Can you restore the comment?

sehilyi (changes requested):
Thank you for this PR. I have one minor comment, and I think it is good to go.

sehilyi (approved):

PR #335, opened 2021-04-06
docs: add link to preprint and url+doi to the bibtex

flekschas:


sehilyi (approved):
Thanks for the update! Looks good.

Issue #342, opened 2021-04-08
Circos-like cancer variant visualization


sehilyi:
**Gosling visualization**
![Screen Shot 2021-04-07 at 6 18 38 PM](https://user-images.githubusercontent.com/9922882/113942027-326d1980-97ce-11eb-961b-b845ef5bdf57.png)

**Original Image**
![Screen Shot 2021-04-07 at 6 22 09 PM](https://user-images.githubusercontent.com/9922882/113942039-3862fa80-97ce-11eb-95ad-7215efb3d9fd.png)

Reference: Fig 3 in https://data.mendeley.com/datasets/2mn4ctdpxp/1

**To-Do**
- [x] Add data transformation for `log`
- [ ] Add data transformation for concatenating string values (e.g., join `"A"` and `"G"` with `">"`, creating `"A > G"`).
- [ ] Not sure what data is used in the middle (i.e., genomic intervals in the original image)
- [ ] Show multiple color legends simultaneously in a circular view
- [ ] Visually distinguish chromosomes (e.g., outline for each chromosome)?

PR #344, opened 2021-04-08
feat: support log transformation

sehilyi:
Towards #342

### DIFF
--- a/src/core/gosling.schema.ts
+++ b/src/core/gosling.schema.ts
@@ -403,8 +403,10 @@ export interface MatrixData {
     url: string;
 }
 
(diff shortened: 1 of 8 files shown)


src/core/gosling.schema.ts
```diff
@@ -403,8 +403,10 @@ export interface MatrixData {
     url: string;
 }
 
+// !!! Transformation is applied in the same order (i.e., filter, log, and then stack)
```
sehilyi, line 406: I think we need to make this more flexible (e.g., allow users to determine the order of data transformation) in the future.

PR #367, opened 2021-04-22
feat: select example through url

wangqianwen0418:
select example through url for displaying examples in the static website

### DIFF
--- a/src/editor/editor.tsx
+++ b/src/editor/editor.tsx
@@ -135,13 +135,14 @@ function Editor(props: any) {
     const urlParams = qs.parse(props.location.search, { ignoreQueryPrefix: true });
     const urlSpec = urlParams?.spec ? JSONUncrush(urlParams.spec as string) : null;
     const urlGist = urlParams?.gist ?? null;
+    const urlExampleIndex = urlParams?.example ? examples.map(d => d.id).indexOf(urlParams.example as string) : null;
 
     const defaultCode = urlGist ? emptySpec() : stringify(urlSpec ?? (examples[INIT_DEMO_INDEX].spec as GoslingSpec));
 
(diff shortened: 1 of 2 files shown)


src/editor/editor.tsx
```diff
@@ -135,13 +135,14 @@ function Editor(props: any) {
     const urlParams = qs.parse(props.location.search, { ignoreQueryPrefix: true });
     const urlSpec = urlParams?.spec ? JSONUncrush(urlParams.spec as string) : null;
     const urlGist = urlParams?.gist ?? null;
+    const urlExampleIndex = urlParams?.example ? examples.map(d => d.id).indexOf(urlParams.example as string) : null;
```
sehilyi: Currently, when a wrong example id is used, the site becomes entirely empty. How about changing this part to show a default (e.g., first) example if `example` does not exist?

sehilyi (changes requested):
Thank you for this update, this looks good! I just left one comment about preventing from showing an empty page when a given `id` does not exist.

Also, just curious how `id`s will be managed in the `gosling-website`. I guess we will be having a JSON array that stores example ids and other related information (e.g., thumbnail), and we just need to update the array to add/remove certain examples to show in the `gosling-website`?

wangqianwen0418:
Thanks! I have updated the code

wangqianwen0418:
> I guess we will be having a JSON array that stores example ids and other related information (e.g., thumbnail), and we just need to update the array to add/remove certain examples to show in the gosling-website?

Yes, you are correct. The examples in gosling-website are currently documented in assets/example-list

sehilyi (approved):
Tested locally, and works well! I will merge this.

Issue #29, opened 2020-12-14
Support grids


sehilyi:
Show rules for showing the chromosome boundaries

PR #373, opened 2021-04-27
feat: support y-axis

### LINKED ISSUE #78: Show y-axis labels

### LINKED ISSUE #92: label of y axis
**problem:**
the label of the y axis has wrong orientation

image

**reproduce: **
```js
{
  "layout": "linear",
  "arrangement": {"direction": "vertical", "columnSizes": 800, "rowSizes": 200},
  "tracks": [
    {
      "data": {
        "url": "https://resgen.io/api/v1/tileset_info/?d=UvVPeLHuRDiYA3qwFlm7xQ",
        "type": "tileset"
      },
      "metadata": {
        "type": "higlass-multivec",
        "row": "sample",
        "column": "position",
        "value": "peak",
        "categories": ["sample 1", "sample 2", "sample 3"]
      },
      "mark": "line",
      "x": {
        "field": "position",
        "type": "genomic",
        "domain": {"chromosome": "2"},
        "axis": "top"
      },
      "y": {"field": "peak", "type": "quantitative", "axis": "left"},
      "color": {"field": "sample", "type": "nominal", "legend": true}
    }
  ]
}
```

sehilyi:
## Update Details
- Show linear scale y-axis on the left or right depending on the `y` channel's `"axis"` property (default `"left"`).
- Show the axis only when the assigned axis height is more than `20px`.
- When a `row` channel is mapped to a nominal field, a single track contains multiple (shared) y-scales. In this case, only the first (top-most) y-axis is shown to prevent making a track too bush (refer to screenshots below).

Fixes #78, Fixes #92

## Open Issues
- How to better display a y-axis, a track title, color/row legends at once w/o visual occlusion? We currently show such visual components inside a track, making the track busy.

Screen Shot 2021-04-28 at 9 55 39 AM

## Screenshots
Screen Shot 2021-04-27 at 5 49 56 PM

Screen Shot 2021-04-27 at 9 44 45 PM

Screen Shot 2021-04-27 at 9 43 19 PM

Screen Shot 2021-04-27 at 9 42 33 PM

### DIFF
--- a/src/core/gosling-track-model.test.ts
+++ b/src/core/gosling-track-model.test.ts
@@ -130,6 +130,11 @@ describe('Gosling track model should be properly generated with data', () => {
         const rowDomain = IsChannelDeep(spec.row) ? (spec.row.domain as string[]) : [];
         const yDomain = IsChannelDeep(spec.y) ? (spec.y.domain as number[]) : [];
 
+        // model properties
+        expect(model.getChannelDomainArray('color')).toHaveLength(3);
+        expect(model.getChannelRangeArray('color')).not.toBeUndefined();
+        expect(model.isShowYAxis()).toBe(true);
+
         // domain
         expect(colorDomain).not.toBeUndefined();
         expect(colorDomain[0]).toBe(1);
--- a/src/core/gosling-track-model.ts
+++ b/src/core/gosling-track-model.ts
@@ -28,7 +28,8 @@ import {
     IsShallowMark,
     IsStackedChannel,
     IsDomainArray,
-    PREDEFINED_COLOR_STR_MAP
+    PREDEFINED_COLOR_STR_MAP,
+    IsRangeArray
 } from './gosling.schema.guards';
 import { CHANNEL_DEFAULTS } from './channel';
 import { getTheme, Theme } from './utils/theme';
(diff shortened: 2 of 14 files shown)


sehilyi:
Hi @wangqianwen0418, I added a PR to display the y-axis (#92). Can you review this PR and take a look at whether the y-axis is properly supported as how you imagined?

wangqianwen0418:
@sehilyi , thanks for the updates! Overall it looks great!
I have just several comments:
- what do you think about putting the labels and ticks outside of the chart box? For example, when `y: {axis: 'left'}`, labels and ticks are put at the left side of the axis. By doing so, we can avoid the overlapping between y labels and charts. Also, this is consistent with how we layout the x labels. 
- I am not sure whether this is done by purpose. When `row` is specified, the y axis is only shown in the second row.


- Adding y axis to the circular layout is quite challenging. Maybe adding a grid background can help? Below is an example 


- A minor issue. It seems that the y label is reversed in the circular layout? I suppose that 0 should be at the bottom?

sehilyi:
@wangqianwen0418, thank you for giving feedback!

> what do you think about putting the labels and ticks outside of the chart box? For example, when y: {axis: 'left'}, labels and ticks are put at the left side of the axis. By doing so, we can avoid the overlapping between y labels and charts. Also, this is consistent with how we layout the x labels.

I agree! I think displaying the axis outside the 'chart box' is ideal, and I would like to support that eventually. There was a technical challenge around our underlying library, i.e., HiGlass, which looked to require me to spend much more time to support 'outward-axes.' So I instead supported the axis similar to how HiGlass currently supports for 1D tracks (e.g., http://higlass.io/l/?d=PuyG-OywTmSRlpsYk1LDFg). I will get back to this issue later.

> I am not sure whether this is done by purpose. When row is specified, the y axis is only shown in the second row.

Yes, it was intended, but I figured it would be confusing to users. So, I have changed to show axes of all rows as can be found below. (I think we might need to consider adding small padding between rows and put a baseline so that axes are more readable.)

Screen Shot 2021-04-29 at 9 40 24 PM

> Adding y axis to the circular layout is quite challenging. Maybe adding a grid background can help? Below is an example

I totally agree, and I will support the grid in a separate PR to address #29.

btw, which visualization library is it in your figure? EChart? Looks worth investigating for me to improve our circular visualizations :)

> A minor issue. It seems that the y label is reversed in the circular layout? I suppose that 0 should be at the bottom?

Ah, 🤦, thanks for catching this issue! I've reversed it, and it should look okay now as follows:

Screen Shot 2021-04-29 at 9 34 47 PM

PR #398, opened 2021-05-29
docs: update url

wangqianwen0418:
change the url from https://gosling-lang.github.io/gosling-website/ to http://gosling-lang.org/


sehilyi (approved):
Thank you for the update! Looks great

PR #409, opened 2021-06-17
chore: split examples

### LINKED ISSUE #368: split examples
@sehilyi , what do you think about splitting some current examples (e.g., semantic zoom) into several small examples.
Some examples are quite dense and may not friendly for new users.

sehilyi:
Made the semantic zoom example into two examples: (1) sequence track and (2) Clinvar lollipop plots.

Fix #368


wangqianwen0418 (approved):
looks great. Thanks for the update

PR #419, opened 2021-07-02
feat: allow specifying `id` and `className` in `GoslingComponent`

sehilyi:
Towards https://github.com/gosling-lang/gosling-react/issues/1


wangqianwen0418 (commented):
I feel it might help to always keep a default className. 
Feel free to modify my commit

sehilyi:
Cool! I like it, thanks!

Issue #69, opened 2021-01-02
Better way to allow zoom/pan in circular tracks?


sehilyi:
To make the zoom/pan interactions useful in circular layouts, the origin of the zoom and pan interactions (i.e., mouse position) should be correctly handled. However, since zoom/pan is solely handled by `higlass` currently, it uses the mouse position as if the tracks are linear. To solve this issue, we need to somehow adjust the mouse position in `higlass`.

sehilyi:
- value scale zoom (zoom/pan along the coordinate encoding non-genomic positions): https://github.com/higlass/higlass/blob/923ed43a08aa8d3ffbb0e1569bf55c5de37f1fef/app/scripts/TrackRenderer.js#L1123

sehilyi:
Perhaps, need to update scales here:
https://github.com/higlass/higlass/blob/923ed43a08aa8d3ffbb0e1569bf55c5de37f1fef/app/scripts/TrackRenderer.js#L1308

PR #422, opened 2021-07-06
feat: allow parallel and serial arrangements of multiple circular views with different domains

### LINKED ISSUE #203: Allow showing different genomic intervals in parallelized and serialized circular views
This is currently impossible since those tracks are contained in a single "higlass view" which can have exactly one `xDomain`. 

To allow this, we should either (1) support multiple `xDomains` in the higlass view or (2) allow overlaying multiple "higlass views" on top of each other (this is not supported by `react-grid-layout` which is the main library being used in HiGlass to layout views).

For the second option above, i.e., overlaying multiple views on top of each other, we could either overlay multiple RGL grids (Related issue: https://github.com/react-grid-layout/react-grid-layout/discussions/1421) or allow higlass to overlay multiple views in a single RGL grid.

sehilyi:
# Overview

This PR allows using different domains among multiple circular views if they are arranged in parallel or serial. Previously, due to technical limitations, such circular views were not able to show different domains. 

The main reason was that to show circular views inside the same rectangular area, such views had to be assigned to the same HiGlass view (i.e., `center.combined`). Since a HiGlass view only has a single domain, this resulted in forcing to show the identical domain across such views.

Given that `react-grid-layout` will allow overlaps (`allowOverlap: true`) in the [next release](https://github.com/react-grid-layout/react-grid-layout/issues/1491), we can render views in separate RGL's views and superpose them (screenshot).

# Example
## Parallel
Multiple ideograms that show different chromosomes are arranged in `parallel`.

gosling-visualization

## Serial
Two chromosomes, chr1 and chr2, are juxtaposed and mirrored (`serial` arrangement).
Screen Shot 2021-07-06 at 11 55 53 AM

# To-Do
- [ ] Since the `allowOverlap` of RGL is not released yet, this PR uses manually built RGL and HiGlass packages. Once the new version of RGL is released, we can upgrade RGL in HiGlass and include an `allowOverlap` prop in the HiGlass view config.

# Open Issues
- Supporting interactions (Related to #69): How to better support zooming and panning in circular views? Currently, only the view on the top is affected by mouse drag and wheel interactions.

Fix #203

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## [0.8.4](https://github.com/gosling-lang/gosling.js/compare/v0.8.3...v0.8.4) (2021-07-02)
+## [0.8.4](https://github.com/gosling-lang/gosling.js/compare/v0.8.3...v0.8.4) (2021-07-06)
 
 
 ### Bug Fixes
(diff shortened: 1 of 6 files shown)


sehilyi:
This has been merged to another PR, #459.

PR #448, opened 2021-08-02
chore: fix types path in package.json

manzt:
Types are emitted in `dist/src` and not `dist/`.


sehilyi (approved):
Thanks for this update!

PR #452, opened 2021-08-04
feat(api): getCanvas()

### LINKED ISSUE #449: add `exportCanvas` to `GoslingComponent`
The `GoslingComponent` takes an optional `ref` that exposes a public API. The `exportPNG` and `exportPDF` functions are good, but ultimately prompt the end-user.

Is there a way to expose a `exportCanvas` or `renderCanvas` function to this API that hooks into the PIXI renderer? Ultimately I'd like to embed a static rendering of a gosling visualization in a jupyter notebook for offline use.

sehilyi:
This PR adds a public API that provides a canvas (`HTMLCanvasElement`) that displays a static gosling visualization (Fix #449):

```js
const { 
   canvas, 
   resolution, 
   width, 
   height 
} = gosRef.current.api.getCanvas({ 
   resolution: 1, // default: 4
   transparentBackground: true // default: false
});
```

An example can be found on the following line of code: https://github.com/gosling-lang/gosling.js/blob/57649bca8531c9d995e4457dfcb9ce808a8bf21b/src/editor/editor.tsx#L407

### DIFF
--- a/src/core/gosling-component.tsx
+++ b/src/core/gosling-component.tsx
@@ -120,6 +120,41 @@ export const GoslingComponent = forwardRef((props: GoslingCompProps, ref: any) =
                     });
                     return ids;
                 },
+                getCanvas: (options: { resolution: number; transparentBackground: boolean }) => {
+                    const resolution = options?.resolution ?? 4;
+                    const transparentBackground = options?.transparentBackground ?? false;
+
+                    const renderer = hgRef.current.pixiRenderer;
+                    const renderTexture = PIXI.RenderTexture.create({
+                        width: renderer.width / 2,
+                        height: renderer.height / 2,
+                        resolution
+                    });
+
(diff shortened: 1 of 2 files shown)


manzt (approved):
Looks great! Thanks for implementing. Curious if you could remove some remove some duplicated code by reusing this method in `exportPNG`.

would

```typescript
exportPNG: (transparentBackground?: boolean) => {
  const { canvas } = this.getCanvas({ resolution: 4, transparentBackground });
  canvas.toBlob((blob: any) => {
    const a = document.createElement('a');
    document.body.append(a);
    a.download = 'gosling-visualization';
    a.href = URL.createObjectURL(blob);
    a.click();
    a.remove();
  }, 'image/png');
},
```

work?

sehilyi:
Thank you for the review, and that is a good catch! I will update the code accordingly and merge it.

PR #454, opened 2021-08-10
chore: expose react component instance api; add types

manzt:
Inspired from some of the work in `ipygosling`...

Adds stronger types for the forwarded `ref` that `GoslingComponent` accepts as a prop. Since `embed` is likely meant to be called in a non-React context, the function now returns a promise for the `GoslingComponentAPI` after `GoslingComponent` has been rendered and after `ref.current` is set to the API.

```javascript
import { embed } from 'gosling.js';

embed(/* ... */)
	.then(api => {
      // do something with the API
	})
    .catch(err => {
      // handle error
    })

```

### DIFF
--- a/src/core/gosling-embed.tsx
+++ b/src/core/gosling-embed.tsx
@@ -22,7 +22,7 @@ export function embed(
         className?: string;
         theme?: Theme;
     }
-) {
+): Promise<GoslingComponentApi> {
     const ref = React.createRef<{ api: GoslingComponentApi }>();
     return new Promise((resolve, reject) => {
         ReactDOM.render(React.createElement(GoslingComponent, { ref, spec, ...options }), element, () => {
--- a/src/core/gosling-embed.tsx
+++ b/src/core/gosling-embed.tsx
@@ -4,7 +4,7 @@ import { Theme } from '..';
 import { GoslingComponent, GoslingComponentApi } from './gosling-component';
 import { GoslingSpec } from './gosling.schema';
 
-const MAX_TRIES = 1000;
+const MAX_TRIES = 120;
 
 /**
  * Embed a Gosling component to a given HTMLElement.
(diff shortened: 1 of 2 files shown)


src/core/gosling-embed.tsx
```diff
@@ -1,9 +1,11 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
 import { Theme } from '..';
-import { GoslingComponent } from './gosling-component';
+import { GoslingComponent, GoslingComponentApi } from './gosling-component';
 import { GoslingSpec } from './gosling.schema';
 
+const MAX_TRIES = 1000;
```
sehilyi:
```suggestion
const MAX_TRIES = 120;
```

Maybe we can use a smaller number not to wait too long if something is wrong (e.g., 1 min)?

sehilyi (approved):
Thank you for the update on the promise and the types! Tested locally, and it works well. I left only one very minor comment.

manzt:
Thanks for the suggestion!

Hmm, I'm wondering if there is some issue with returning `ref.current.api` rather than `ref`. I just tried...

```javascript
embed(/* ... */).then(api => api.on('mousemove', (payload) => console.log(payload));
```

and it don't see anything in the console when I interact with the viewer. Either this is an issue with something like passing a value rather than a reference, or I I'm not triggering the event handler correctly.

sehilyi:
Oh, yes. I think the issue is related to the 'mousemove' API. This API is still experimental and works only in certain cases (e.g., `tooltip` should be specified in the gosling.js spec).

I tested with the `exportPNG()` API, and it worked well. So, I think `ref.current.api` is working as expected.

manzt:
Ok, that makes things a lot more simple. I'll tidy things up and merge!

sehilyi:
Thank you for testing it before merging this.

I just tested with a different example spec that should work with the `mousemove` API, and it is working as expected.

PR #456, opened 2021-08-11
refactor: create higlass-wrapper-component

manzt:
Creates a re-useable `HiGlassComponentWrapper` and `createApi` utils to be used in reactive `GoslingComponent` contenxt and `embed`.


src/core/api.ts
```diff
@@ -1,14 +1,161 @@
...
+    return {
+        on: (type, callback) => {
+            switch (type) {
+                case 'mouseover':
+                    return PubSub.subscribe(type, callback);
+                default: {
+                    console.warn('Subsriber not recognized.');
+                }
+            }
+        },
```
manzt: Rather than using another ref, we can have the end users of the API handle "unsubscribing" from the global PubSub event bus.

src/core/init.ts
```diff
@@ -68,4 +72,6 @@ export function init() {
     higlassRegister({ dataFetcher: CSVDataFetcher, config: CSVDataFetcher.config }, { pluginType: 'dataFetcher' });
     higlassRegister({ dataFetcher: RawDataFetcher, config: RawDataFetcher.config }, { pluginType: 'dataFetcher' });
     higlassRegister({ dataFetcher: BBIDataFetcher, config: BBIDataFetcher.config }, { pluginType: 'dataFetcher' });
+
+    once = true;
```
manzt, line 76: I thought it made sense to call this function once, unless there is some reason this is not desired ...

manzt:
~Not ready for review yet. Still fixing some bugs...~

EDIT: ready to go!

src/core/init.ts
sehilyi: I agree!

src/core/gosling-embed.ts
```diff
@@ -0,0 +1,68 @@
...
+    const component = React.createElement(HiGlassComponentWrapper, {
+        ref,
+        viewConfig,
+        size,
+        id: opts.id,
+        className: opts.className,
+        options: opts
+    });
+    ReactDOM.render(component, element);
+    return ref.current!;
```
sehilyi: Oh, this non-null assertion operator looks useful!

src/core/api.ts
```diff
@@ -1,14 +1,168 @@
...
+type EventMap = PubSubEvent<'mouseover', CommonEventData>;
+// & PubSubEvent<'my-event', { hello: 'world' }> & PubSubEvent<'foo', "bar">;
+
+export interface GoslingApi {
+    on<EventName extends keyof EventMap>(type: EventName, callback: (payload: EventMap[EventName]) => void): void;
+    zoomTo(viewId: string, position: string, duration?: number): void;
+    zoomToExtent(viewId: string, duration?: number): void;
+    zoomToGene(viewId: string, gene: string, duration?: number): void;
+    getViewIds(): string[];
+    exportPNG(transparentBackground?: boolean): void;
```
sehilyi, line 35: Not directly related to this PR, but worth asking your opinion here. I realized I used inconsistent letter cases with `exportPNG` and `exportPDF`. Do you think it makes more sense for the end users if we change these to `exportPng` and `exportPdf`, just like  `getViewIds` and `linkingId`, so that users can more easily remember these?

src/core/api.ts
```diff
@@ -1,14 +1,168 @@
...
+//
+// - Two different events ('mouseover' & 'my-event') with the same payload
+//
+// PubSubEvent<'mouseover' | 'my-event', { same: 'payload' }>
+type PubSubEvent<EventName extends string, Payload> = {
+    [Key in EventName]: Payload;
+};
+
+// New `PubSubEvent`s should be added to the `EventMap`...
+type EventMap = PubSubEvent<'mouseover', CommonEventData>;
```
sehilyi, line 26: Thank you for defining this!

src/core/api.ts
```diff
@@ -1,14 +1,168 @@
...
+}
+
+export function createApi(
+    hgRef: React.RefObject<HiGlassApi | undefined> | HiGlassApi,
+    hgSpec: HiGlassSpec | undefined,
+    theme: Required<CompleteThemeDeep>
+): GoslingApi {
+    const getHg = () => {
+        // Safely get higlass API
+        if ('api' in hgRef) return hgRef;
```
sehilyi, line 55: It looks like calling API functions after using `embed()` cause an error on my computer (API functions worked properly when I tested with `<GoslingComponent ... />`):

```js
embed(
    document.getElementById('example')!,
    editedGos,
    {}
).then(d => {
    // I can visually see the gosling visualization rendered on the screen.
    d.zoomTo('view-1', 'chr3'); // "error TypeError: Cannot use 'in' operator to search for 'api' in null"
}).catch(e => {
    console.log('error', e);
});
```

Looks like the return value of `launchHiglass()` (i.e., `ref.current`) is `null`. This is odd since the codes in this function are similar to that in higlass' `launch()`.

sehilyi (commented):
Thanks a lot for this update!

Code-wise it looks great, but when I tested them locally, calling API functions caused an error after using `embed()`. I left a comment about this in detail below. Would you be able to check this issue?

src/core/api.ts
sehilyi: I agree. Do you think we can add an `off()` function for that?

src/core/gosling-embed.ts
manzt: It is useful, but dangerous for the exact reason you've found  ;)

src/core/api.ts
manzt: Personally, I don't have a huge preference! Now that the API is strongly typed, end-users will get auto-completion in their editor when interacting with the API..

image

So I'd base your decision on aesthetically what your preference is.

src/core/api.ts
manzt: definitely! I did something similar in regl-scatterplot and it worked quite nicely. It's really user-friendly to get auto-completion for both the available events _and_ then the corresponding payload.

image

image

A nice way to "explore" the available API for a developer, IMO.

src/core/api.ts
manzt: Let me take a closer look. We might want to make the return type of `launchHiglass` be a `Promise` (like before) for the _higlass_ API, and poll the ref in a `setTimeout` until the API is defined? I also find this odd, since the code is so similar to `launch`, and that seems to be defined consistently.

manzt: can you share the gosling config so I can explore?

sehilyi: Sure thing. I used the following spec (also defined in [/src/editor/example/visual-encoding.ts](https://github.com/gosling-lang/gosling.js/blob/master/src/editor/example/visual-encoding.ts)):

```js
{
  "title": "Visual Encoding",
  "subtitle": "Gosling provides diverse visual encoding methods",
  "layout": "linear",
  "arrangement": "vertical",
  "centerRadius": 0.8,
  "xDomain": {"chromosome": "1", "interval": [1, 3000500]},
  "views": [
    {
      "id": "view-1",
      "tracks": [
        {
          "data": {
            "type": "multivec",
            "row": "sample",
            "column": "position",
            "value": "peak",
            "categories": ["sample 1", "sample 2", "sample 3", "sample 4"],
            "binSize": 4
          },
          "mark": "rect",
          "x": {"field": "start", "type": "genomic", "axis": "top"},
          "row": {"field": "sample", "type": "nominal", "legend": true},
          "color": {"field": "peak", "type": "quantitative", "legend": true},
          "tooltip": [
            {"field": "start", "type": "genomic", "alt": "Start Position"},
            {"field": "end", "type": "genomic", "alt": "End Position"},
            {
              "field": "peak",
              "type": "quantitative",
              "alt": "Value",
              "format": ".2"
            },
            {"field": "sample", "type": "nominal", "alt": "Sample"}
          ],
          "width": 600,
          "height": 130
        }
      ]
    },
    {
      "tracks": [
        {
          "data": {
            "type": "multivec",
            "row": "sample",
            "column": "position",
            "value": "peak",
            "categories": ["sample 1", "sample 2", "sample 3", "sample 4"]
          },
          "mark": "bar",
          "x": {"field": "position", "type": "genomic", "axis": "top"},
          "y": {"field": "peak", "type": "quantitative"},
          "color": {"field": "sample", "type": "nominal", "legend": true},
          "tooltip": [
            {"field": "position", "type": "genomic", "alt": "Position"},
            {
              "field": "peak",
              "type": "quantitative",
              "alt": "Value",
              "format": ".2"
            },
            {"field": "sample", "type": "nominal", "alt": "Sample"}
          ],
          "width": 600,
          "height": 130
        }
      ]
    },
    {
      "tracks": [
        {
          "data": {
            "type": "multivec",
            "row": "sample",
            "column": "position",
            "value": "peak",
            "categories": ["sample 1", "sample 2", "sample 3", "sample 4"]
          },
          "mark": "bar",
          "x": {"field": "position", "type": "genomic", "axis": "top"},
          "y": {"field": "peak", "type": "quantitative", "grid": true},
          "color": {"field": "sample", "type": "nominal", "legend": true},
          "tooltip": [
            {"field": "position", "type": "genomic", "alt": "Position"},
            {
              "field": "peak",
              "type": "quantitative",
              "alt": "Value",
              "format": ".2"
            },
            {"field": "sample", "type": "nominal", "alt": "Sample"}
          ],
          "width": 600,
          "height": 130
        }
      ]
    },
    {
      "alignment": "overlay",
      "data": {
        "type": "multivec",
        "row": "sample",
        "column": "position",
        "value": "peak",
        "categories": ["sample 1", "sample 2", "sample 3", "sample 4"]
      },
      "x": {"field": "position", "type": "genomic", "axis": "top"},
      "y": {"field": "peak", "type": "quantitative"},
      "color": {"field": "sample", "type": "nominal", "legend": true},
      "tracks": [
        {"mark": "line"},
        {
          "mark": "point",
          "size": {"field": "peak", "type": "quantitative", "range": [0, 2]}
        }
      ],
      "tooltip": [
        {"field": "position", "type": "genomic", "alt": "Position"},
        {
          "field": "peak",
          "type": "quantitative",
          "alt": "Value",
          "format": ".2"
        },
        {"field": "sample", "type": "nominal", "alt": "Sample"}
      ],
      "width": 600,
      "height": 130
    },
    {
      "tracks": [
        {
          "data": {
            "type": "multivec",
            "row": "sample",
            "column": "position",
            "value": "peak",
            "categories": ["sample 1", "sample 2", "sample 3", "sample 4"]
          },
          "mark": "point",
          "x": {"field": "position", "type": "genomic", "axis": "top"},
          "y": {"field": "peak", "type": "quantitative"},
          "size": {"field": "peak", "type": "quantitative"},
          "color": {"field": "sample", "type": "nominal", "legend": true},
          "opacity": {"value": 0.5},
          "tooltip": [
            {"field": "position", "type": "genomic", "alt": "Position"},
            {
              "field": "peak",
              "type": "quantitative",
              "alt": "Value",
              "format": ".2"
            },
            {"field": "sample", "type": "nominal", "alt": "Sample"}
          ],
          "width": 600,
          "height": 130
        }
      ]
    },
    {
      "tracks": [
        {
          "data": {
            "type": "multivec",
            "row": "sample",
            "column": "position",
            "value": "peak",
            "categories": ["sample 1", "sample 2", "sample 3", "sample 4"]
          },
          "mark": "point",
          "x": {"field": "position", "type": "genomic", "axis": "top"},
          "y": {"field": "peak", "type": "quantitative", "grid": true},
          "size": {"field": "peak", "type": "quantitative"},
          "color": {"field": "sample", "type": "nominal", "legend": true},
          "opacity": {"value": 0.5},
          "tooltip": [
            {"field": "position", "type": "genomic", "alt": "Position"},
            {
              "field": "peak",
              "type": "quantitative",
              "alt": "Value",
              "format": ".2"
            },
            {"field": "sample", "type": "nominal", "alt": "Sample"}
          ],
          "width": 600,
          "height": 130
        }
      ]
    },
    {
      "tracks": [
        {
          "data": {
            "type": "multivec",
            "row": "sample",
            "column": "position",
            "value": "peak",
            "categories": ["sample 1", "sample 2", "sample 3", "sample 4"]
          },
          "mark": "area",
          "x": {"field": "position", "type": "genomic", "axis": "top"},
          "y": {"field": "peak", "type": "quantitative"},
          "color": {"field": "sample", "type": "nominal", "legend": true},
          "stroke": {"value": "white"},
          "strokeWidth": {"value": 0.5},
          "tooltip": [
            {"field": "position", "type": "genomic", "alt": "Position"},
            {
              "field": "peak",
              "type": "quantitative",
              "alt": "Value",
              "format": ".2"
            },
            {"field": "sample", "type": "nominal", "alt": "Sample"}
          ],
          "width": 600,
          "height": 130
        }
      ]
    },
    {
      "tracks": [
        {
          "data": {
            "url": "https://raw.githubusercontent.com/sehilyi/gemini-datasets/master/data/circos-segdup-edited.txt",
            "type": "csv",
            "chromosomeField": "c2",
            "genomicFields": ["s1", "e1", "s2", "e2"]
          },
          "mark": "withinLink",
          "x": {
            "field": "s1",
            "type": "genomic",
            "domain": {"chromosome": "1", "interval": [103900000, 104100000]}
          },
          "xe": {"field": "e1", "type": "genomic"},
          "x1": {
            "field": "s2",
            "type": "genomic",
            "domain": {"chromosome": "1"}
          },
          "x1e": {"field": "e2", "type": "genomic"},
          "color": {"field": "s1", "type": "nominal"},
          "stroke": {"value": "black"},
          "strokeWidth": {"value": 0.5},
          "opacity": {"value": 0.2},
          "width": 600,
          "height": 130
        }
      ]
    }
  ]
}
```

src/core/api.ts
sehilyi: That's true! I also heavily rely on auto-completion so that I do not have to memorize everything.

I guess, in our case, we can try using consistent letter cases (so, `exportPng` and `exportPdf`) in the API, as well as the grammar, to make sure that they look more organized and unified.

src/core/api.ts
manzt: Hmm I can't seem to reproduce on my machine. Confusing. I wonder if it has something to do with `HiGlassComponent` being a class component and higlass has more granular control over when the ref is initialized. 

I don't think we need to wait that many ticks, pushed the latest with these changes if you're able to try it out.

src/core/api.ts
manzt: That's an interesting idea. Currently end-users could interact with this like:

```javascript
const api = await embed(/* ... */);
const token = api.on('mousover', console.log) // log the event payload 

// do something ...

PubSub.unsubscribe(token);
```

I would be comfortable calling this global, but it might be a bit confusing to call something outside of the API. I don't know if I've seen an `on`/`off` api, which could also be confusing since they are just aliases for `subscribe`/`unsubscribe`, but that would look like:

```javascript
const api = await embed(/* ... */);
const token = api.on('mousover', console.log) // log the event payload 

// do something ...

api.off(token); // just wraps PubSub.unsubscribe
```

Personally, I think the most idiomatic approach would be to use `subscribe`/`unsubscribe` on the API since this pattern is common:

```javascript
const api = await embed(/* ... */);
const token = api.subscribe('mousover', console.log) // log the event payload 

// do something ...

api.unsubscribe(token); // just wraps PubSub.unsubscribe
```

src/core/api.ts
sehilyi: Sorry, do you mean you pushed new changes to this branch?

manzt: Forgot to push :)

sehilyi: Ah, this works on my machine!

src/core/api.ts
sehilyi: Agree, should be really useful for end-users!

src/core/api.ts
sehilyi: Thanks for the examples! I think I prefer the latter two slightly more than the first one as subscribe/unsubscribe can be done within the API. I saw `on`/`off` in HiGlass (https://docs.higlass.io/javascript_api.html#public.off), but since we are basically using the PubSub's `subscribe`/`unsubscribe`, I agree that the last one seems to be the most idiomatic approach.

manzt: Yes, I think higlass `on`/`off` hooks into more than just `pub-sub-es`. In our case we are just wrapping `PubSubJS`, which lets you unsubscribe in many ways https://github.com/mroderick/PubSubJS: e.g. `token`, `function` or `topic`. 

```javascript
const api = await embed(/* .. */);

const cb = (data) => console.log(data);
const cb2 = (data) => console.log({ data });

const token = api.subscribe('mouseover', cb
const token2 = api.subscribe('mouseover', cb2);
const token3 = api.subscribe('mouseover', cb);

PubSub.unsubscribe(token); // unsubscribe specific listener
PubSub.unsubscribe(cb); // unsubscribe line 1 & 3 ( same cb )
PubSub.unsubscribe('mouseover'); // unsubscribe all
```

By aliasing `PubSub.unsubscribe` in our API, we have the freedom to change implementations in the future as well as allow/restrict what patterns of unsubscribe are allowed.

manzt: I'm going to leave this as is (the clunkly `on` + `PubSub.unsubscribe`), and we can open a follow-up PR for deciding on the new naming API. Does that sound ok?

sehilyi: Yes, that sounds good. This is something that we can revisit when we want to include more diverse APIs.

sehilyi: And I like the current flexibility as I see the last example you provided.

src/core/api.ts
manzt: I've made these changes.

PR #460, opened 2021-08-13
chore: rename  `api.export{PNG,PDF}` -> `api.export{Png,Pdf}`

manzt:


sehilyi (approved):
Thank you for the update!

PR #461, opened 2021-08-13
chore: deprecate `api.on` in favor of `api.subscribe`/`api.unsubscribe`

### LINKED ISSUE #458: deprecate `api.on` in favor of `api.subscribe`/`api.unsubscribe`
From #456

----

**Trevor** 

Currently... 

```javascript
const api = await embed(/* ... */);
const token = api.on('mousover', console.log) // log the event payload 

// do something ...

PubSub.unsubscribe(token);
```

I would be comfortable calling this global, but it might be a bit confusing to call something outside of the API. I don't know if I've seen an `on`/`off` api, which could also be confusing since they are just aliases for `subscribe`/`unsubscribe`, but that would look like:

```javascript
const api = await embed(/* ... */);
const token = api.on('mousover', console.log) // log the event payload 

// do something ...

api.off(token); // just wraps PubSub.unsubscribe
```

Personally, I think the most idiomatic approach would be to use `subscribe`/`unsubscribe` on the API since this pattern is common:

```javascript
const api = await embed(/* ... */);
const token = api.subscribe('mousover', console.log) // log the event payload 

// do something ...

api.unsubscribe(token); // just wraps PubSub.unsubscribe
```

---- 

**Sehi**

Thanks for the examples! I think I prefer the latter two slightly more than the first one as subscribe/unsubscribe can be done within the API. I saw `on`/`off` in HiGlass (https://docs.higlass.io/javascript_api.html#public.off), but since we are basically using the PubSub's `subscribe`/`unsubscribe`, I agree that the last one seems to be the most idiomatic approach.

----

**Trevor**

Yes, I think higlass `on`/`off` hooks into more than just `pub-sub-es`. In our case we are just wrapping `PubSubJS`, which lets you unsubscribe in many ways https://github.com/mroderick/PubSubJS: e.g. `token`, `function` or `topic`. 

```javascript
const api = await embed(/* .. */);

const cb = (data) => console.log(data);
const cb2 = (data) => console.log({ data });

const token = api.subscribe('mouseover', cb
const token2 = api.subscribe('mouseover', cb2);
const token3 = api.subscribe('mouseover', cb);

PubSub.unsubscribe(token); // unsubscribe specific listener
PubSub.unsubscribe(cb); // unsubscribe line 1 & 3 ( same cb )
PubSub.unsubscribe('mouseover'); // unsubscribe all
```

By aliasing `PubSub.unsubscribe` in our API, we have the freedom to change implementations in the future as well as allow/restrict what patterns of unsubscribe are allowed.

_Originally posted by @manzt in https://github.com/gosling-lang/gosling.js/pull/456#discussion_r687942498_

manzt:
Fixes #458

### DIFF
--- a/src/core/api.ts
+++ b/src/core/api.ts
@@ -27,7 +27,11 @@ type EventMap = PubSubEvent<'mouseover', CommonEventData>;
 // & PubSubEvent<'my-event', { hello: 'world' }> & PubSubEvent<'foo', "bar">;
 
 export interface GoslingApi {
-    on<EventName extends keyof EventMap>(type: EventName, callback: (payload: EventMap[EventName]) => void): void;
+    subscribe<EventName extends keyof EventMap>(
+        type: EventName,
+        callback: (payload: EventMap[EventName]) => void
+    ): void;
+    unsubscribe(tokenOrFunction: string | ((...args: unknown[]) => unknown)): void;
     zoomTo(viewId: string, position: string, duration?: number): void;
     zoomToExtent(viewId: string, duration?: number): void;
     zoomToGene(viewId: string, gene: string, duration?: number): void;
@@ -93,7 +97,7 @@ export function createApi(
         };
     };
     return {
-        on: (type, callback) => {
+        subscribe: (type, callback) => {
             switch (type) {
                 case 'mouseover':
                     return PubSub.subscribe(type, callback);
@@ -102,6 +106,7 @@ export function createApi(
                 }
             }
         },
+        unsubscribe: tokenOrFunction => PubSub.unsubscribe(tokenOrFunction),
         // TODO: Support assemblies (we can infer this from the spec)
         zoomTo: (viewId, position, duration = 1000) => {
             // Accepted input: 'chr1' or 'chr1:1-1000'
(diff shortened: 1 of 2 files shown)


sehilyi (approved):
Thank you! Looks great!

sehilyi:
~~Removed ` characters in the titles since it looks commit messages are not properly parsed/generated when merged.~~

~~For example,~~
~~chore: rename `api.export{PNG,PDF}` -> `api.export{Png,Pdf}`~~

~~becomes~~

~~chore: rename ->~~

~~Hopefully, this issue will be fixed by the GitHub team in the future.~~

Changed back since I found that other commit messages with ` characters were generated properly. Unsure why the recent commit message was not parsed properly.

manzt:
That's my fault. I made the commit message in my terminal and zsh I interpreted it as an (empty) command. Sorry about that... but with a squash and merge, the weird commit should go away. Will avoid in the future!

sehilyi:
Ah, didn't know that a commit message is used as is with squash and merge when there is only one commit for a PR. I thought the title of a PR is used always.

No worries at all. I was just curious about the reason 🙂

manzt:
To be clear, I titled the PR on GitHub via the web UI, but the commit "chore: rename ->" came from my shell interpreting \` \` incorrectly.

My comment regarding squash/merge just meant that the odd commit ("chore: rename -->") will not end up in the commit history because it will be squashed and renamed. I also found it confusing :)

Issue #493, opened 2021-09-02
Improve build + packaging; make higlass external?


manzt:
## Motivation

Currently the npm distribution for gosling.js includes:

```
dist/
├── 1.gosling.js
├── 1.gosling.js.map
├── 2.gosling.js
├── 2.gosling.js.map
├── 3.gosling.js
├── 3.gosling.js.map
├── 4.gosling.js
├── 4.gosling.js.map
├── gosling.js
├── gosling.js.map
└── worker.js
```

All of which are minified, UMD-like exports. It seems `{1,2,3,4}.gosling.js` are chunks for dynamic imports that are aliased by `__webpack_require__` in `dist/gosling.js`, and `dist/worker.js` is a separate build entirely that ends up being referenced in the source code via `raw-loader!../../dist/worker.js`.

Ideally, an npm package should provide an ES module entrypoint (which is easier to statically analyze and treeshake) rather than a pre-built & minified UMD. In my experience, webpack isn't a great choice for a library because you can't target ESM as an output and the generated bundles are difficult to load with by bundlers (e.g. `__wekpack_require__` statements cannot be resolved, so any project using `gosling.js` as a dependency will likely be unable to load `{1,2,3,4}.gosling.js`).

I have experimented with a new build for `gosling.js` on my vacation and wanted to share some ideas.

## Approach

- Bundle `gosling.js` the _library_ (`src/index.ts`) with `esbuild` or `rollup` and target esm output (as well as UMD bundle like before for use in script tags)

- Consume the _library_ `gosling.js` in the `editor` (rather than importing from `src/`). Since the editor is really an "app" and not distributed in the npm module, it would make sense to configure this separately from the library. This is similar to what we do in Viv, where `src/` corresponds to the library and `avivator/` is an app that consumes `src/` as a module. This enforces that the editor only use the public API from gosling.js.

```bash
src/ # bundled with rollup/esbuild
editor/ # built with webpack, imports generated bundle
```

- Make `higlass` an external dependency. Since the CSS for higlass is already required via a script tag, I think it would also make sense to also externalize higlass. I'm not sure this is something that has been discussed, but since higlass is distributed as a UMD bundle, we could similarly add it as script tag along side `react`, `react-dom`, and `pixi.js` for use with our UMD bundle.

[...]

PR #494, opened 2021-09-03
feat: replace webpack with esbuild for packaging

manzt:
An attempt at #493


tsconfig.json
```diff
@@ -20,16 +20,27 @@
...
+    "paths": {
+      "generic-filehandle": ["./src/alias/generic-filehandle.ts"],
+      "zlib": ["./src/alias/zlib.ts"],
+      "vm": ["./src/alias/vm.ts"],
+      "@gmod/bgzf-filehandle": ["./src/alias/bgzf-filehandle.ts"],
+      "gosling.js": ["./src/index.ts"],
+      "@gosling.schema": ["./src/core/gosling.schema.ts"],
+      "@higlass.schema": ["./src/core/higlass.schema.ts"]
+    },
+    "baseUrl": "."
```
manzt, line 33: This is the key to supporting the custom resolution. esbuild respects these paths when resolve module names, even in dependencies.

manzt: Typescript will also respect the `gosling.js` & `@*` aliases with the editor source code. This allows us to point the compiler to the correct place when when typechecking the code, but ultimately `webpack.config.js` looks in `dist/gosling.es.js` for the entrypoint.

editor/index.tsx
```diff
@@ -3,6 +3,7 @@ import ReactDOM from 'react-dom';
 import { BrowserRouter, Route } from 'react-router-dom';
 import Editor from './editor';
 import './index.css';
+import 'higlass/dist/hglib.css';
```
manzt, line 6: Can just import the CSS here which `css-loader` takes care of..

public/index.html
```diff
@@ -40,18 +40,11 @@
...
     </head>
 
     <body>
         <noscript>You need to enable JavaScript to run this app.</noscript>
         <div id="root"></div>
-        <script crossorigin type="text/javascript" src="https://unpkg.com/react@16/umd/react.development.js"></script>
-        <script crossorigin type="text/javascript" src="https://unpkg.com/react-dom@16/umd/react-dom.development.js"></script>
-        <script crossorigin type="text/javascript" src="https://unpkg.com/pixi.js@5/dist/pixi.js"></script>
-        <script crossorigin type="text/javascript" src="https://unpkg.com/react-bootstrap@0.32.1/dist/react-bootstrap.js"></script>
-        <script crossorigin type="text/javascript" src="https://unpkg.com/higlass@1.11.7/dist/hglib.min.js"></script>
```
manzt, line 55: None of these modules were externalized in the `webpack.config.editor.js`, so the amount of javascript downloaded when navigating to `gosling.js.org` is ~2x currently (since the modules are bundled in the output produced by webpack as well as imported separately here and not used)

sehilyi: Oh, this was a huge mistake..

editor/index.tsx
sehilyi: Sounds good!

tsconfig.json
sehilyi: As I am not that familiar with aliases, may I ask how you determine to select those aliases (e.g., zlib, vm, ...)?  I can only think of an inefficient way that manually reads `@gmod` codes and including any dependent modules to this list if they require implementation on our own for the esbuild.

manzt: Yes this is what I had to do currently. Ideally very few dependencies for gosling include usage of node builtins. All the "problem" modules for bundling are isolated to `@gmod/*` and it took sometime to find the error boundaries. I can make some PRs to these dependencies. The use of `Proxy` in for the node-builtins makes it so that we get a runtime error if any other methods try to be accessed by other modules:

```typescript
// alias/zlib.ts
export default new Proxy(
    {},
    {
        get(key) {
            if (key === 'inflateSync') return (src: Uint8Array) => Buffer.from(unzlibSync(src));
            throw new Error(
                'Module "zlib" has been externalized for browser compatibility and cannot be accessed in client code.'
            );
        }
    }
);
```

```javascript
import zlib from './alias/zlib.js';
zlib.inflateSync(...) // ok
zlib.flush() // Error! Module "zlib" has been exernalized for browser....
```

manzt: https://github.com/GMOD/binary-parser/pull/23

manzt: https://github.com/GMOD/bgzf-filehandle/pull/50

manzt: https://github.com/GMOD/generic-filehandle/pull/88

manzt:
closing in favor of #522

PR #495, opened 2021-09-03
feat: replace webpack/webpack-dev-server with vite

manzt:
Builds off of #494, but replaces the webpack-dev-server/build with `vite`. I think ideally we could just use Vite to unify the build, but the `build.js` script gives us better control over the "package".


sehilyi:
Thanks a lot for these efforts. I think it is a good idea to use `vite` consistently.

One issue I found while playing with Editor was getting errors with BAM examples (e.g., "BAM file pileup tracks"), which seems to be related to the `threads` package. Will take a look at the issues.

Screen Shot 2021-09-17 at 10 30 20 AM

manzt:
I'll take a closer look! Do you have BAM example i could look at?

manzt:
Ok I think I was able to fix the issue (fixed in #494), and have rebased the changes here. (It wasn't related to `threads` but actually `@gmod` again...) However, now the issue seems to be that the BAM example is invalid (the `chromsizes` url is undefined): 

image

sehilyi:
> Do you have BAM example i could look at?

I was using the following example files:

```js
"url": "https://s3.amazonaws.com/gosling-lang.org/data/example_higlass.bam",
"indexUrl": "https://s3.amazonaws.com/gosling-lang.org/data/example_higlass.bam.bai"
```

The error/warning messages you shared seem to be the same as what we currently get in `gosling.js.org` and seem to be not related to the reasons why the BAM examples do not work.

Screen Shot 2021-09-19 at 4 37 08 PM

> It wasn't related to threads but actually `@gmod` again...

Was there a nice way to find the origin of the issues other than manually checking individual packages to find unimplemented parts in our alias?

sehilyi:
Somehow, it looks to be working on my machine on the latest `manzt/esbuild-vite` branch:

Screen Shot 2021-09-19 at 5 12 26 PM

manzt:
> Was there a nice way to find the origin of the issues other than manually checking individual packages to find unimplemented parts in our alias?

Not that I'm aware of. The issue is that `@gmod` modules rely on runtime checks for webpack constructs (e.g. `typeof __webpack_require__ !== 'function' ? require('fs') : null;`. Webpack <5 will polyfill node built-ins automatically (like `vm`, `fs`, `zlib`) but modern front-end bundlers don't polyfill automatically, and the reality is that some of these these polyfills just bloat gosling.js. For example, `generic-filehandle/LocalFile.js` will never be used in our code and relies on `node:fs`, so there is a large polyfill that ends up in the resulting code for no reason.

We can probably make some PRs to `@gmod` to suggest some changes, but the aliases provide a way to hook into the build tool and typescript type-checking to inject our modified modules.

manzt:
@sehilyi was there an issue previously with monaco? I noticed you changed the imports in 32c12d0. Your changes work fine, but _all_ language bindings are generated for the final bundle, which is unnecessary and takes longer to build as a result...

image

The configuration I had before was designed to only add the `json` feature (which is the only thing the editor supports): 

image

Also the largest asset is about 1KB smaller ...

sehilyi:
> was there an issue previously with monaco? 

@manzt So sorry, I should've explained the issues before editing codes. The issues are that some built-in features in the editor are not working anymore, such as auto-complete, pop-up warning messages, search feature, resizing fonts.

**Search box that appears upon pressing command + f:**
Screen Shot 2021-09-20 at 11 07 59 AM

**Pop-up dialog:**
Screen Shot 2021-09-20 at 11 07 52 AM

**Auto-complete:**
Screen Shot 2021-09-20 at 11 07 41 AM

**Clicking on the resize button on the left-most panel leads to app crash**:
Screen Shot 2021-09-20 at 11 10 22 AM

manzt:
@sehilyi. Thanks for all feedback. Would you be able to try out the latest commit and see if the features you have noted are working correctly? The entrypoint for `monaco-editor` points to `node_modules/monaco-editor/esm/vs/editor/editor.main.js`, which includes.

```javascript
import '../language/typescript/monaco.contribution';
import '../language/css/monaco.contribution';
import '../language/json/monaco.contribution';
import '../language/html/monaco.contribution';
import '../basic-languages/monaco.contribution';

export * from './edcore.main';
```

I created a similar import in `editor/monaco.ts`, but left off the extra languages.

sehilyi:
Tested locally and works with all the Monaco features we previously had.

manzt:
Closing in favor of #522

PR #503, opened 2021-09-13
feat: enable auto doc generate

wangqianwen0418:
This pull request modifies the current gosling.schema, aiming to enable the automatic (or semi-automatic)  doc generation from the typescript code. 

The comment `/** experimental */` can be used to ignore properties that we do not want to show in the documentation.

Some initial results about the data doc is available at https://github.com/gosling-lang/gosling-docs/tree/docs-from-tsjson/docs_auto

**the ts code**
image

**the generated markdown doc**
image


wangqianwen0418:
This pull request enables better table generation for the doc website.
https://github.com/gosling-lang/gosling-website/pull/14

block comments (\/\*\* \* \/) will be converted into descriptions in schema.json, then used in the property tables as markdown string.
@xxx in block comments will be excluded in the description.
Including keywords (['deprecated', 'experimental', 'not supported', 'internal']) will ignore this property row in the property table.
If possible, please use (/** */) rather than (//) for comments.
line comments (//) will not be converted into descriptions.

Please refer to ts-json-schema-generator and vega-lite for more details about how the block comments are parsed.
Below is a comment example 
image

src/core/gosling.schema.ts
```diff
@@ -89,23 +117,39 @@ export interface CommonTrackDef extends CommonViewDef, CommonRequiredTrackDef {
...
+    /** If defined, will show the textual label on the left-top corner of a track */
+    title?: string;
     subtitle?: string; // Being used only for a title track (i.e., 'text-track')
 
     // Arrangement
     overlayOnPreviousTrack?: boolean;
 
     // Circular Layout
+    /**
+     * Specify the outer radius of tracks when `{"layout": "circular"}`
```
sehilyi: I notice that the letter case is inconsistent across options. Should we use the upper case consistently for the readability?

src/core/gosling.schema.ts
```diff
@@ -39,7 +39,12 @@ export interface OverlaidTracks extends CommonViewDef, Partial<SingleTrack> {
 }
 
 export interface MultipleViews extends CommonViewDef {
+    /**
+     * Specify how multive views are arranged.
```
sehilyi:
```suggestion
     * Specify how multiple views are arranged.
```

src/core/gosling.schema.ts
```diff
@@ -39,7 +39,12 @@ export interface OverlaidTracks extends CommonViewDef, Partial<SingleTrack> {
...
+     * One of `"parallel"`,  `"serial"`, `"horizontal"`, `"vertical"`.
```
sehilyi: Do you think it is possible/realistic to infer the possible options by checking typescript types instead of manually adding the duplicated descriptions as a comment?

src/core/gosling.schema.ts
```diff
@@ -49,38 +54,61 @@ export type Assembly = 'hg38' | 'hg19' | 'hg18' | 'hg17' | 'hg16' | 'mm10' | 'mm
 export type ZoomLimits = [number | null, number | null];
 
 export interface CommonViewDef {
+    /** specify the layout type of all tracks, either `"linear"` or `"circular"` */
     layout?: Layout;
+    /** either `"horizontal"` or `"vertical"` */
```
sehilyi: Likewise, I am curious if these (possible properties) can be auto-generated?

src/core/gosling.schema.ts
```diff
@@ -49,38 +54,61 @@ export type Assembly = 'hg38' | 'hg19' | 'hg18' | 'hg17' | 'hg16' | 'mm10' | 'mm
...
     orientation?: Orientation;
 
+    /**
+     * - if `{"layout": "linear"}`, specify the space between tracks in pixels;
+     *
+     * - if `{"layout": "circular"}`, specify the space between tracks in percentage ranging from 0 to 100.
+     */
     spacing?: number;
+    /** whether to disable [Zooming and Panning](http://gosling-lang.org/docs/user-interaction#zooming-and-panning),
+     * __Default:__ false.
```
sehilyi: Do you think it is better to use code blocks like the below for all descriptions that are related to actual properties in the grammar (e.g., all default values)?

```suggestion
     * __Default:__ `false`.
```

src/core/gosling.schema.ts
```diff
@@ -49,38 +54,61 @@ export type Assembly = 'hg38' | 'hg19' | 'hg18' | 'hg17' | 'hg16' | 'mm10' | 'mm
...
+     */
     static?: boolean;
     zoomLimits?: ZoomLimits; // limits of zoom levels. default: [1, null]
 
     // offsets
```
sehilyi: Can remove this since we added more detailed comments.

```suggestion
```

src/core/gosling.schema.ts
```diff
@@ -49,38 +54,61 @@ export type Assembly = 'hg38' | 'hg19' | 'hg18' | 'hg17' | 'hg16' | 'mm10' | 'mm
...
+    /**specify an ID for [linking multiple views](http://gosling-lang.org/docs/user-interaction#linking-views) */
     linkingId?: string;
+    /** not supported  */
     xAxis?: AxisPosition; // not supported currently
 
     /**
      * Proportion of the radius of the center white space.
+     *
+     * __Default:__ 0.3
+     * @Range [0, 1]
```
sehilyi, line 97: Curious if this tag (i.e., `Range`) is used in the docs?

src/core/gosling.schema.ts
```diff
@@ -513,16 +833,26 @@ export interface DisplaceTransform {
...
+        /** The name of a nominal field to group rows by in prior to piling-up */
+        groupField?: string;
     };
+    /** A string that specifies the type of diseplancement, one of `"pile"` and `"spread"`  */
     method: DisplacementType;
     newField: string;
 
-    // "pile" specific parameters (TODO: make this a separate interface)
+    /** Specify maximum rows to be generated (default has no limit) */
     maxRows?: number; // Specify maximum rows to be generated (default: `undefined` meaning no limit)
```
sehilyi: Should we remove the original comment which is now duplicated?

```suggestion
    maxRows?: number;
```

src/core/gosling.schema.ts
```diff
@@ -358,100 +511,265 @@ export type Aggregate = 'max' | 'min' | 'mean' | 'bin' | 'count';
...
+
+    /**
+     * specify the name of genomic data fields
+     */
     genomicFields?: string[];
+
+    /**
+     * specify the number of rows loaded from the url.
+     *
+     * _Default:__ 1000
```
sehilyi:
```suggestion
     * __Default:__ `1000`
```

Missing `_` character broke the style of `Default`

sehilyi (commented):
This is great! 👍

I left several minor comments and have some high-level thoughts:

- Was just curious if some additional special tags can be used for frequently used categories, something like `@default` for default values and `@options` for the list of possible options. Not something that is essential since the current way already works.
- Just so that I can properly add comment statements next time when I update the schema -- Do you have any general guideline on which kinds of properties require comment statements and which does not? Perhaps, good to add comment statements to all properties that I add?
- Do you mind adding descriptions about adding comment statements for docs to `CONTRIBUTING.md`? (just like your explanation above)
- I think we can use the consistent letter cases (e.g., first letter upper case) and styles (e.g., code blocks for default values, like "__Default:__ `[0, 100]`")

Also, could you make sure to run `yarn format`? I notice the code styles of many example files are changed unnecessarily.

wangqianwen0418:
@sehilyi, thanks a lot for these helpful comments!

> Just so that I can properly add comment statements next time when I update the schema -- Do you have any general guideline on which kinds of properties require comment statements and which does not? Perhaps, good to add comment statements to all properties that I add?

Thanks! That would be great. Generally speaking, if a property accepts user-assigned values, it is better to provide some descriptions for it.
You already have some very useful comments in the code. It would be super helpful if you can use block comments when possible in the future.

> Do you mind adding descriptions about adding comment statements for docs to CONTRIBUTING.md? (just like your explanation above)

Sure, will do that shortly.

> I think we can use the consistent letter cases (e.g., first letter upper case) and styles (e.g., code blocks for default values, like "Default: [0, 100]")

Totally agree. I will go through to check the first letter case

> Also, could you make sure to run yarn format? I notice the code styles of many example files are changed unnecessarily.

Sure, of course! Will make sure to run `yarn format` next time.

src/core/gosling.schema.ts
wangqianwen0418: Sure, will go through the file to check.

src/core/gosling.schema.ts
wangqianwen0418: > Likewise, I am curious if these (possible properties) can be auto-generated?

Good point. I simply follow the practice in vega-lite but I do believe this can be auto-generated.
Will address this in a PR in `gosling-website`

src/core/gosling.schema.ts
wangqianwen0418: Good point. I simply follow the practice in vega-lite but I do believe this can be auto-generated.
Will address this in a PR in `gosling-website`

src/core/gosling.schema.ts
wangqianwen0418: No, the @tag is not used

src/core/gosling.schema.ts
wangqianwen0418: Agree. Will go through the code to check

PR #524, opened 2021-09-23
feat: add `gosling.js/embed` package export

manzt:
Adds https://github.com/manzt/gosling-embed/issues/2 to repo

### DIFF
--- a/vite.config.js
+++ b/vite.config.js
@@ -74,7 +74,16 @@ const editorConfig = defineConfig({
     },
     plugins: [
         bundleWebWorker(),
-        reactRefresh(),
+        // reactRefresh(),
+        {
+          name: 'inject-global',
+          apply: 'serve',
+          transform(code, id) {
+            if (/node_modules/.test(id)) {
+              return 'const global = globalThis;\n' + code;
+            }
+          }
+        }
     ],
 });
 
--- a/vite.config.js
+++ b/vite.config.js
@@ -10,27 +10,41 @@ import pkg from './package.json';
  * see: https://github.com/hms-dbmi/viv/pull/469#issuecomment-877276110
  * @returns {import('vite').Plugin}
  */
-const bundleWebWorker = () => {
-    return {
-        name: 'bundle-web-worker',
-        apply: 'serve', // plugin only applied with dev-server
-        async transform(_, id) {
-            if (/\?worker_file$/.test(id)) {
-                // just use esbuild to bundle the worker dependencies
-                const bundle = await esbuild.build({
-                    entryPoints: [id],
-                    inject: ['./src/alias/buffer-shim.js'],
-                    format: 'esm',
-                    bundle: true,
-                    write: false,
-                });
-                if (bundle.outputFiles.length !== 1) {
-                    throw new Error('Worker must be a single module.');
-                }
-                return bundle.outputFiles[0].text;
+const bundleWebWorker = {
+    name: 'bundle-web-worker',
+    apply: 'serve', // plugin only applied with dev-server
+    async transform(_, id) {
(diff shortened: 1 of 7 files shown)


manzt:
@sehilyi This is still a WIP, but I just rebased the changes on `master` so it is easier to diff.

manzt:
@sehilyi should be ready for review. A cool "feature" I've added in this PR is the ability to test out the embed module using the `vite` dev server. I had to use the changes from #526 to allow for this, but essentially you can start the dev server.

Since `vite` is a pure ESM, any entrypoint can be imported in the browser via a dynamic `import(...)`:

```
vite # start the vite dev server
```

image


You can edit any of the contents in `src/` and refresh the page to apply changes. You might need to refresh the page a couple times since observable does some caching, but regardless this feature is useful for trying out this module.

vite.config.js
```diff
@@ -10,27 +10,41 @@ import pkg from './package.json';
...
+ * Transforms /node_modules/.vite/* to include an alias for global.
+ * Only enabled during development.
+ *
+ * @type {import('vite').Plugin}
+ */
+const injectGlobal = {
+    name: 'inject-global',
+    apply: 'serve', // only runs during
+    transform(code, id) {
+        if (/node_modules/.test(id)) {
```
sehilyi: Should this be something like the following?

```suggestion
        if (/node_modules\/\.vite\//.test(id)) {
```

It looks `global` is added to files outside `/node_modules/.vite/` (e.g., `/node_modules/monaco-editor/esm/vs/editor/standalone/browser/quickInput/standaloneQuickInput.css`), breaking styles of editor:

Screen Shot 2021-09-24 at 5 20 49 PM

vite.config.js
```diff
@@ -66,19 +78,21 @@ const esm = defineConfig({
 });
 
 const editorConfig = defineConfig({
```
sehilyi: Perhaps, we can change this to something like `dev` to make it clear that this config is used when developing either Editor or `embed`.

```suggestion
const dev = defineConfig({
```

vite.config.js
```diff
@@ -66,19 +78,21 @@ const esm = defineConfig({
...
-        bundleWebWorker(),
-        reactRefresh(),
-    ],
+    plugins: [bundleWebWorker, injectGlobal],
 });
 
 export default ({ command, mode }) => {
     if (command === 'build' && mode === 'lib') return esm;
+    if (mode === 'editor') {
+        editorConfig.plugins.push(reactRefresh());
```
sehilyi: Perhaps, `dev`?

```suggestion
        dev.plugins.push(reactRefresh());
```

vite.config.js
```diff
@@ -66,19 +78,21 @@ const esm = defineConfig({
...
+    }
     return editorConfig;
```
sehilyi:
```suggestion
    return dev;
```

sehilyi (commented):
Looks great! I especially like that this allows developing `embed` more conveniently. I just left small comments.

Also, would you mind updating [CONTRIBUTING.md](https://github.com/gosling-lang/gosling.js/blob/master/CONTRIBUTING.md#getting-started) by adding instructions on how to develop `embed` (just like your comment above https://github.com/gosling-lang/gosling.js/pull/524#issuecomment-926846584)?

We could also add an explicit script in `package.json` like `"start-embed": "vite"`.

vite.config.js
manzt: I updated the regex to retrict to only `/node_modules/.vite/*.js` imports

manzt:
@sehilyi I've addressed your comments. Should be ready to merge!

sehilyi (approved):
@manzt Thank you for the update, looks good to me!

PR #526, opened 2021-09-24
feat: remove aliases & update @gmod/bbi w deps

manzt:
re-locks the yarn.lock with changes from: https://github.com/GMOD/binary-parser/pull/23


manzt:
Not sure why the 10.x tests are still running/preventing a merge ... but otherwise this should be good to go!

sehilyi (approved):
Tested locally and looks good to me!

I was able to remove 10.x status checks from Settings.

PR #527, opened 2021-09-25
feat: migrate `uuid` &  `box-intersect` to browser-compatible deps

manzt:
The hack in #524 to assign `global` in the dev-server isn't very robust as you've found. Digging into the issue dependencies, I found that `uuid` and `box-intersect` include node-isms which are difficult to polyfill. 

I found that the latest version of `uuid` includes a browser-friendly export, so fixing that dependency is easy. However, [`box-intersect`](https://github.com/mikolalysenko/box-intersect) is a bit of a challenge. I found a [benchmark](https://github.com/mikolalysenko/box-intersect-benchmark) on the README for that project that included a reference for `rbush` as an alternative. I wonder if we could use that library instead as an alternative? Something to think about...


manzt:
I'm going to close this for now because it's not essential. We can open in the future.

src/gosling-genomic-axis/axis-track.ts
```diff
@@ -549,36 +559,33 @@ function AxisTrack(HGC: any, ...args: any[]): any {
...
-                    allTexts[j].text.visible = false;
-                    if (this.options.layout === 'circular' && allTexts[j].rope) {
-                        this.pTicksCircular.removeChild(allTexts[j].rope);
-                    }
-                } else {
-                    allTexts[i].text.visible = false;
-                    if (this.options.layout === 'circular' && allTexts[i].rope) {
-                        this.pTicksCircular.removeChild(allTexts[i].rope);
+            // using bounding boxes of the text objects, calculate overlaps
+            allTexts
```
sehilyi, line 572: Given that we are calculating overlaps of a small number of labels, I think it's okay to use another library (i.e., rbush). Using rbush, I've implemented a part that calculates overlaps and hides labels. We can consider removing `box-intersect` from the dependency in the future.

manzt:
Great! I was able to remove the `injectGlobal` plugin as well. Should be ready to merge.

sehilyi (approved):
Looks great!

This also seems to address errors regarding `Buffer` in BAM examples (#535) when tested with `yarn start`.

PR #530, opened 2021-09-27
chore: Fix TSC compilation errors; add type-checking to CI

manzt:
Vite doesn't perform any type-checking, so running `tsc` is no long a part of the CI. This PR fixes a compiler error with the gosling API, and adds a GH Action to perform typechecking as a part of our CI.


sehilyi (approved):
Looks 👍! Great to include type-checking as a part of our CI.

Issue #506, opened 2021-09-14
More precise encoding types


manzt:
Right now all the track encoding fields (`x`, `y`, `color`, etc) reuse the `Channel` definition; however, not all possible Channel definitions are supported for each encoding field.

Some examples of definitions that are completely valid by the schema but have no meaning and aren't supported.

```
"x": { "type": "quantitative", ... }
"x": { "value": 10 },
"color: { "type": "genomic", "aggregate": ..., ... }
```

Right now we implicitly rely on _convention_ to ensure that users don't make these mistakes. But we could could make unique types for each encoding field that capture the boundaries of what is allowed.
For example, if "x" and "xe" only allow for `"type": "genomic"` then there is no point in forcing users to specify a type: it can only be genomic. Similar for "y": a user just needs to specify the field name if only one "type" is supported.

This also lends itself to being more expressive per definition, and refining types for the domain use case. e.g.

```typescript
type X = {
 start: string,
 end?: string,
 axis?: Axis,
}
```

```
"x": { "start": "position", "axis": "top" },
// or
"x": { "start": "position", "end": "positionEnd", "axis": "top" },
```

in **gos** this would end up looking something like:

```python
gos.Track(...).encode(x=gos.X(start="position", end="positionEnd"))
```

and we could even make shorthand parsing that turns this into:

```python
gos.Track(...).encode(x="position:positionEnd")
```

sehilyi:
We definitely need to use precise types (i.e., `X`, `Y`, `Size`, instead of a union `Channel` type)!

Your second idea on merging `x` and `xe` is interesting and might be a good idea. One issue I found with the current separate channels (i.e., `x: {...}, xe: {...}`) is that some properties (e.g., `axis`, `domain`, `range`) that should be specified only once can be specified multiple times:

```js
x: { field: "start", axis: "top" },
xe: { field: "end", axis: "none" }, 
// we should pick only one of two axis properties internally
```

This uncertain and wrong spec can be prevented by using your idea, i.e.,
```js
x: { start: "start", end: "end", axis: "top" } // axis can be defined only once
```

We use at most four different field names for a single axis—`x`, `xe`, `x1`, `x1e`—for band connection, so we could allow something like

```js
x: { start: "start", end: "end", start2: "start2", end2: "end" }
```

Users can still specify only one field as is:

```js
x: { field: "position", ... }
```

I also like how it could end up in **gos** (i.e., `x="start:end"`).

[...]

PR #533, opened 2021-09-29
feat: more precise channel types (e.g., `X`, `Y`, and `Color` instead of `Channel`)

sehilyi:
This PR uses more precise types of individual channels, like `X`, `Y`, and `Color`. This allows checking gosling.js specs more precisely, e.g., `row` cannot be encoded with a `genomic` field:

```js
row: { ..., type: "genomic" } // Error
```

or `x` is always mapped to a `genomic` field:

```js
x: { ..., type: "quantitative" } // Error
```

Genomic `domain` is used for only x-axis channels:

```js
y: { ..., domain: { chromosome: '1' } } // Error
```

Towards #506


src/core/gosling-to-higlass.ts
```diff
@@ -186,7 +186,13 @@ export function goslingToHiGlass(
         // check whether to show axis
         ['x', 'y'].forEach(c => {
             const channel = (firstResolvedSpec as any)[c];
-            if (IsChannelDeep(channel) && channel.axis && channel.axis !== 'none' && channel.type === 'genomic') {
+            if (
+                IsChannelDeep(channel) &&
+                'axis' in channel &&
```
sehilyi, line 191: Since not every channel has an `'axis'` property, this addition was necessary.

src/core/gosling-track-model.test.ts
```diff
@@ -40,21 +40,6 @@ describe('gosling track model should properly validate the original spec', () =>
         const model2 = new GoslingTrackModel({ ...track, x: { field: 'x', type: 'genomic' } }, [], getTheme());
         expect(model2.validateSpec().valid).toBe(true);
     });
-
-    it('genomic coordinate cannot be encoded with a color channel', () => {
```
sehilyi, line 44: I think we can remove this test (and other related tests) since typescript/jsonschema now checks this.

src/core/gosling-track-model.ts
```diff
@@ -162,13 +162,8 @@ export class GoslingTrackModel {
         // zero baseline
         SUPPORTED_CHANNELS.forEach(channelKey => {
             const channel = spec[channelKey];
-            if (
-                IsChannelDeep(channel) &&
-                typeof channel.zeroBaseline === 'undefined' &&
-                channel.type !== 'genomic' &&
-                ['x', 'y'].includes(channelKey)
```
sehilyi, line 169: I think using `zeroBaseline` by default for all quantitative channels consistently (e.g., `size`, `color`, `y`) makes more sense. So, updated this part accordingly.

sehilyi:
@manzt Can you review this PR and see if this works for `gos`? 

`*.json` files are automatically updated ones, so you can ignore them.

manzt:
@sehilyi I'll give this a look over. This will require some changes to gos since all uses of `gos.Channel` will need to be replaced with `gos.<Encoding>`. Hopefully I'll be able to try out soon.

src/core/gosling.schema.ts
```diff
@@ -305,33 +305,112 @@ export type ChannelType = keyof typeof ChannelTypes | string;
...
     legend?: boolean;
     baseline?: string | number;
     zeroBaseline?: boolean; // We could remove this and use the `baseline` option instead
     mirrored?: boolean; // Show baseline on the top or right instead of bottom or left
+    aggregate?: Aggregate;
     grid?: boolean;
     linkingId?: string;
     flip?: boolean; // Flip a track vertically or horizontally?
-    stack?: boolean; // Experimental: We could use this option to stack visual marks, addressing the visual overlap (e.g., stacked bar).
     padding?: number; // Experimental: Used in `row` and `column` for vertical and horizontal padding.
```
manzt, line 339: Is there a reason why all the fields remain optional? I think a benefit of splitting out the different types is that the typing gets much stronger. For example `{}` is currently valid for all types. By being more restrictive, we can get compilation or validation errors instead of runtime errors checking for fields that may or may not exist.

For example, by setting:

```typescript
type ChannelDeepCommon = { field: string } // all types _must_ have a field
```

anything that derives from this type _must_ have `field` defined (e.g. `{ }` is now invalid for any derived type). 

Another benefit is that in Typescript we can narrow union types very effectively when they are defined. For example, you can't capture 

```typescript
// currently need to check if type is defined 
if ('type' in encoding && encoding.type === 'nominal') {
   // only encodings where type is defined and is nominal
  // could capture y: { type: 'nominal' }
} else {
  // could still capture y: { } | y: { field: genomic } | y : { field: quantitative }
}

// if required...
if (encoding.type === 'nominal') {
  // captures y: { type: 'nominal' }
} else {
  // no need to handle undefined case
}

```

src/core/gosling.schema.ts
```diff
@@ -154,26 +154,26 @@ export interface SingleTrack extends CommonTrackDef {
...
+    color?: Color | ChannelValue;
+    size?: Size | ChannelValue;
+    text?: Text | ChannelValue;
 
-    stroke?: Channel;
-    strokeWidth?: Channel;
-    opacity?: Channel;
+    stroke?: Stroke | ChannelValue;
+    strokeWidth?: StrokeWidth | ChannelValue;
+    opacity?: Opacity | ChannelValue;
```
manzt: Somewhat related to https://github.com/gosling-lang/gosling.js/issues/482, but since these fields aren't exported included on an explicit type, the code to auto-generate the utilities in gos needs to inspect `SingleTrack` deeply and then conditionally handle fields with are detected to be visual encodings. This works, but isn't very future proof. 

Ideally we could at least define the "encoding" properties for the track as its own type and then intersect that type with the `SingleTrack`.

```typescript
interface TrackBase extends CommonTrackDef { ... }
export interface VisualEncoding { ... }
export type SingleTrack = TrackBase & VisualEncoding;
```

This way we can autogenerate the custom wrappers in gos by inspecting the `VisualEncoding` type rather than `SingleTrack` which contains more than just the visual encodings.

manzt:
@sehilyi, I'm just realizing that didn't submit my comments. Sorry!

src/core/gosling.schema.ts
sehilyi: That's a good question! One of the reasons that made me reluctant to make `type` and `field` required options was that `brush` mark does not necessarily use those fields:

```js
{
  ...,
  "alignment": "overlay",
  "tracks": [
    {"mark": "bar", "x": {"field": "start", "type": "genomic"}},
    {"mark": "brush", "x": {"linkingId": "detail"}} // <--  
  ]
}
```

I agree that making those two fields required is more ideal for multiple reasons including what you mentioned above. 

While there are many possible ways to address this, I was thinking of adding `xLinkingId` inside `SingleTrack` and remove `linkingId` inside `x` channel. In this way, users can use brush without any channels:

```diff
{
  ...,
  "alignment": "overlay",
  "tracks": [
     {"mark": "bar", "x": {"field": "start", "type": "genomic"}},
-    {"mark": "brush", "x": {"linkingId": "detail"}} // <--  
+    {"mark": "brush", "xLinkingId": "detail"} // <--  
  ]
}
```

I think this is better than using channels without `field` and `type` w/ `brush` since `ChannelDeep` is supposed to be bound to actual data, which means no-field or no-type is a bit odd.

@manzt To make this PR less large, I can update this (grammar, compiler, and examples) in a separate PR. What do you think?

src/core/gosling.schema.ts
```diff
@@ -141,7 +141,9 @@ export type MarkType =
     | 'header';
 
 /* ----------------------------- TRACK ----------------------------- */
-export interface SingleTrack extends CommonTrackDef {
+type SingleTrack = SingleTrackBase & Encoding;
```
sehilyi: `Encoding` type added here

src/core/gosling.schema.ts
sehilyi: Thanks for the suggestion! I have updated the code accordingly.

I suppose this address #482 only partly since, for example, `x` can still be specified in either `gos.Track(...)` or `gos.Track().properties(...)`.

If so, I will work on fixing this in a separate PR this by adding an explicit `encoding: Encoding` property inside `SingleTrack`:

```js
{  // SingleTrack
   mark: Mark,
   encoding: Encoding, // <--
   width: number,
   height: number
}
```

sehilyi:
@manzt Thank you so much for reviewing this PR! I've added an explicit `Encoding` type as you suggested and left several questions. Let me know what do you think!

> I'm just realizing that didn't submit my comments. Sorry!

Sorry, do you mean some of your additional comments that should have been submitted are lost?

src/core/gosling.schema.ts
manzt: An alternative would be to just add another type to the union of `x` the field, no? Similar to `ChannelValue`.

```typescript
interface Encoding {
  x: X | ChannelValue | LinkingId;
  // ...
}
```

Then the type can easily be distinguished in our code, and no additional fields are valid when a `linkingId` is provided:

```typescript
let track: SingleTrack;

if ('linkingId' in track.x) {
  // x === { linkingId: string }
} else if ('value' in track.x) {
  // x === { value: number | string }
} else {
  // x == { field ... }
}
```

I think this should be just fine in gos, but we can also just make a util like `gos.value` for this use case:

```python
import gosling as gos

track = gos.Track(...).encode(x=gos.link('view1'))
```

manzt: > @manzt To make this PR less large, I can update this (grammar, compiler, and examples) in a separate PR. What do you think?

Yes, completely agree. I'll be out of town for the rest of the week just FYI, but feel free to add me to PRs to review.

src/core/gosling.schema.ts
manzt: > I suppose this address #482 only partly since, for example, x can still be specified in either gos.Track(...) or gos.Track().properties(...).

Yes exactly, but creating an explicit exported type is a great first step. I know creating an `encoding` property might be a lot of work, and just having this as is stands is quite valuable.

manzt:
> Sorry, do you mean some of your additional comments that should have been submitted are lost?

Nothing to worry here. I had just reviewed this earlier last week and forgot to press submit until yesterday.

PR #536, opened 2021-09-30
chore: encourage users to ask questions on BioStars

sehilyi:
It might be better if we can provide a link that creates a new post with a predefined tag (i.e., `gosling.js`), like the `labels` in GitHub issue templates. But, it looks there is no such feature.

Also, there seems to be no subscription feature in BioStars & Stack Overflow (i.e., notify when new posts with specific tags are created by someone). I would need to manually visit those sites frequently.


manzt (approved):
Great! Thanks for adding this. I don't have further ideas at the moment.

PR #537, opened 2021-10-01
fix: inject Buffer polyfill for BamWorker

### LINKED ISSUE #535: Errors with BAM data examples
When opening a demo example with a BAM file at gosling.js.org, the error message says `Buffer is not defined`. The BAM data is not rendered in the visualization.

Screen Shot 2021-09-30 at 9 17 40 PM

manzt:
Fixes #535. The issue is that the worker code is in a different environment than the main script, so `Buffer` isn't exposed as a global (which is required in the @gmod/bam dependency tree). This should fix this for the time being, but ideally any module using `Buffer` uses an import rather than accessing a global.

```javascript
Buffer.from(...) // Reference error; buffer is not defined 
```

```javascript
import { Buffer } from 'buffer'; // bundler knows it needs a polyfill for Buffer
Buffer.from(...) 
```

I haven't had the time to look into where this occurs in the dependency tree to be fixed. If/when that is changed, we can remove this hack.

### DIFF
--- a/vite.config.js
+++ b/vite.config.js
@@ -19,7 +19,7 @@ const bundleWebWorker = {
             const bundle = await esbuild.build({
                 entryPoints: [id],
                 inject: ['./src/alias/buffer-shim.js'],
-                format: 'esm',
+                format: 'iife',
                 bundle: true,
                 write: false,
             });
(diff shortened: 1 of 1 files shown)


sehilyi (approved):
Great! BAM examples are working, I will merge this.

PR #540, opened 2021-10-04
feat(editor): support using relative CSV file URLs for Gist specs

sehilyi:
This PR allows Editor users to specify relative file URLs of CSV data in Gist specs. This will work when files are located in the same Gist where the spec (i.e., `gosling.js`) is located.

- Example URL for Editor: http://localhost:3000/?gist=sehilyi/c7d3494ef35366e9ee77b570ffd9d184
- Example Gist URL: https://gist.github.com/sehilyi/c7d3494ef35366e9ee77b570ffd9d184

File structure:
```js
GitHub Gist
ㄴ gosling.js    // spec
ㄴ example.csv   // file to be used
```

Original spec in Gist (`gosling.js`):
```js
data: { ..., type: 'csv', 
   url: './example.csv'
}
```

The spec that is updated in the Editor before compiling:
```js
data: { ..., type: 'csv', 
   url: 'https://gist.githubusercontent.com/sehilyi/c7d3494ef35366e9ee77b570ffd9d184/raw/NA_summits.bed'
}
```

### DIFF
--- a/editor/editor.tsx
+++ b/editor/editor.tsx
@@ -13,6 +13,7 @@ import SplitPane from 'react-split-pane';
 import ErrorBoundary from './error-boundary';
 import { debounce, isEqual } from 'lodash-es';
 import { examples } from './example';
+import { traverseTracksAndViews } from '../src/core/utils/spec-preprocess';
 import stripJsonComments from 'strip-json-comments';
 import * as qs from 'qs';
 import JSONCrush from 'jsoncrush';
@@ -86,6 +87,23 @@ const emptySpec = (message?: string) => (message !== undefined ? `{\n\t// ${mess
 
 const getDescPanelDefultWidth = () => Math.min(500, window.innerWidth);
 
+/**
+ * Convert relative CSV data URLs to absolute URLs.
+ * (e.g., './example.csv' => 'https://gist.githubusercontent.com/{urlGist}/raw/example.csv')
+ */
+function resolveRelativeCsvUrls(spec: string, importMeta: URL) {
+    const newSpec = JSON.parse(spec);
+    // https://regex101.com/r/l87Q5q/1
+    // eslint-disable-next-line
+    const relativePathRegex = /^[.\/]|^\.[.\/]|^\.\.[^\/]/;
+    traverseTracksAndViews(newSpec as gosling.GoslingSpec, (tv: any) => {
+        if (tv.data && tv.data.type === 'csv' && relativePathRegex.test(tv.data.url)) {
+            tv.data.url = new URL(tv.data.url, importMeta).href;
+        }
+    });
+    return stringify(newSpec);
+}
+
 const fetchSpecFromGist = async (gist: string) => {
     let metadata: any = null;
     try {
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -14,3 +14,6 @@ export { compile } from './core/compile';
 export { validateGoslingSpec } from './core/utils/validate';
 export { GoslingComponent } from './core/gosling-component';
 export { embed } from './core/gosling-embed';
+
+// Utils
+export { traverseTracksAndViews, traverseViewArrangements } from './core/utils/spec-preprocess';
```
sehilyi: We could expose more util functions here later.

src/core/utils/spec-preprocess.ts
```diff
@@ -34,12 +37,15 @@ import { getStyleOverridden } from '../utils/style';
...
  */
 export function traverseTracks(
-    spec: GoslingSpec,
+    spec: GoslingSpec | View | PartialTrack,
     callback: (t: Partial<Track>, i: number, ts: Partial<Track>[]) => void
 ) {
     if ('tracks' in spec) {
-        spec.tracks.forEach((...args) => callback(...args));
-    } else {
+        spec.tracks.forEach((t, i, ts) => {
```
sehilyi, line 44: Updates on this file are not directly related to this PR but are to just handle some cases that were not handled correctly (i.e., `tracks` in a track). These two functions were not properly updated previously after we had changed the grammar.

src/index.ts
manzt: Good to know! I wonder if we need to expose these currently to users.. We could treat these as private functions by importing from src:

```typescript
// editor.tsx
import { traverseTracksAndView } from 'gosling.js/core/utils/spec-preprocess';
```

editor/editor.tsx
```diff
@@ -235,6 +235,20 @@ function Editor(props: any) {
...
+    function resolveRelativeCsvUrls(spec: string) {
+        const newSpec = JSON.parse(spec);
+        gosling.traverseTracksAndViews(newSpec as gosling.GoslingSpec, (tv: any) => {
+            if (tv.data && tv.data.type === 'csv' && tv.data.url.indexOf('./') === 0) {
+                tv.data.url = tv.data.url.replace('./', `https://gist.githubusercontent.com/${urlGist}/raw/`);
+            }
+        });
+        return stringify(newSpec);
+    }
+
```
manzt: I wonder if we could make this an even more reusable:

```typescript
/**
 * Convert relative CSV data URLs to absolute URLs.
 * (e.g., './example.csv' => 'https://gist.githubusercontent.com/{urlGist}/raw/example.csv')
 */
function resolveRelativeCsvUrls(spec: string, importMeta: URL) {
	const newSpec = JSON.parse(spec);
	// https://regex101.com/r/l87Q5q/1
	const relativePathRegex = /^[.\/]|^\.[.\/]|^\.\.[^\/]/
	gosling.traverseTracksAndViews(newSpec as gosling.GoslingSpec, (tv: any) => {
		if (tv.data&& tv.data.type === 'csv' && relativePathRegex.test(tv.data.url)) {
			tv.data.url = new URL(tv.data.url, importMeta).href;
		}
	});
	return stringify(newSpec);
}
```

and then:

```typescript
    const specUrl = new URL(`https://gist.githubusercontent.com/${gist}/raw/${dataFile}`);
    const whenCode = fetch(specUrl.href).then(async response =>
        response.status === 200 ? resolveRelativeCsvUrls(await response.text(), specUrl) : null
    );
```

the second argument to `URL` is a [`base`](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL) to resolve a path relative to.

manzt (approved):
Nice! just a couple of minor comments.

src/index.ts
sehilyi: I assumed those functions would be useful for users to have, but I think it might be better to expose them when we find/hear actual needs from users. Let me make changes reflecting your suggestion!

editor/editor.tsx
sehilyi: Thank you for this suggestion, I really like it! Good to know that `URL` supports the relative URL using the second parameter. I will make the changes reflecting your comment.

Issue #541, opened 2021-10-08
path.js: process is not defined 


sehilyi:
When `gosling.js` is used with a [`vite-react-ts`](https://github.com/fabien-ml/react-ts-vite-template) template, multiple error messages appear:

<img width="763" alt="Screen Shot 2021-10-08 at 12 59 51 PM" src="https://user-images.githubusercontent.com/9922882/136496492-77184af2-e7c9-4741-90ac-5cafb046076a.png">

Adding the following lines in the base HTML files address the issue:

```html
<script>
   var process = { platform: '', env: '' }
</script>
```

Would need to include this in the build so that users do not have to care about these details.

sehilyi:
Another error message:
<img width="760" alt="Screen Shot 2021-10-08 at 1 14 38 PM" src="https://user-images.githubusercontent.com/9922882/136497400-4b440e7c-29c3-425d-b2d2-6ec6e931f346.png">

PR #554, opened 2021-10-15
fix: fix errors related to `process` and `global`

### LINKED ISSUE #553: `Bezier` is not exported from `bezier-js` error
This seems to occur when the latest `bezier-js` (4.1.0~) is used with `create-react-app` (https://github.com/vasturiano/react-force-graph/issues/282).

Screen Shot 2021-10-15 at 4 12 57 PM

### LINKED ISSUE #555: rng-browser.js:7 Uncaught ReferenceError: global is not defined
This issue occurs from an old `uuid`, which is being used in an old `slugid`, a higlass dependency.

Need to specify the resolution of `slugid` in `package.json` so that the recent `uuid` is being used.

Refer to https://github.com/uuidjs/uuid/issues/565

sehilyi:
Fix #553 
Fix #555

#541 is still not fixed, so the following addition is still required with `vite`.

```html
<script> var process = { platform: '', env: '' } </script>
```

### DIFF
--- a/index.html
+++ b/index.html
@@ -46,14 +46,6 @@
     <body>
         <noscript>You need to enable JavaScript to run this app.</noscript>
         <div id="root"></div>
-        <script src="https://bundle.run/buffer@6.0.3"></script>
-        <script>
-          // Many dependencies reference `Buffer` without the use of an
-          // import. Since Vite relies on ESM, these references aren't resolved
-          // and we get runtime errors. We could use something like `@rollup/plugin-inject`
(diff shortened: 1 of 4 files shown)


index.html
```diff
@@ -46,14 +46,6 @@
     <body>
         <noscript>You need to enable JavaScript to run this app.</noscript>
         <div id="root"></div>
-        <script src="https://bundle.run/buffer@6.0.3"></script>
```
sehilyi, line 49: Might need to revert this chnage

PR #556, opened 2021-10-15
feat: infer SV types and find mates from `bam` files; add an API to get raw data in JSON format

sehilyi:
gosling-visualization (1)

# Updates
- Put clipping representations back
- Highlight reads with potential SVs
- Highlight reads with no mates found

# Issues
- When `opts.pairAcrossChr` is set (https://github.com/GMOD/bam-js), I see the following error message from both `Gosling.js` and `higlass/higlass-pileup`:
Screen Shot 2021-11-05 at 6 19 46 AM

### DIFF
--- a/editor/editor.tsx
+++ b/editor/editor.tsx
@@ -229,9 +229,9 @@ function Editor(props: any) {
     // publish event listeners to Gosling.js
     useEffect(() => {
         // if (gosRef.current) {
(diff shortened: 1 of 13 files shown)


sehilyi:
See how the JBrowse2 team determined SVs: https://github.com/GMOD/jbrowse-components/blob/81209864b7b8e92eb8f2230074d95ce871a7aee6/plugins/alignments/src/PileupRenderer/PileupRenderer.ts#L407

PR #558, opened 2021-10-17
feat: allow access to HiGlass `ref` from JS API

### LINKED ISSUE #559: Access HiGlass ref from api
I really appreciate the awesome work you put into this repo, great work!

We are using gosling.js to visualize self-defined genomic sequence formats. One of our features requires to access the HiGlass track object and view config, and I figure this can only be accessed via HiGlass Ref. 

Can the HiGlass Ref be exposed in the API? Or is there a better way to achieve this?

I made a fork to expose HiGlass Ref in API, and created a PR here
https://github.com/gosling-lang/gosling.js/pull/558

seanli3:
To the best of my knowledge, there is currently no way to access higlass ref from the gosling react component.

This PR adds the higlass ref to the exposed api object. 

Fixes #559


sehilyi (approved):
Thank you so much for creating this PR and being the first external contributor to `gosling-lang/gosling.js`! 🎉 🥰

This will be a very useful thing to have!

I tested locally and worked well, e.g., 

```js
gosRef.current.hgRef.current.api.zoomTo('my-id', 1, 1000, 1, 1000);
```

I will also update our documentation (http://gosling-lang.org/docs/js-api) accordingly.

PR #574, opened 2021-10-29
chore: add encoding comments

wangqianwen0418:


wangqianwen0418:
@sehilyi, are properties like `field`, `range`, `domain` required for the encoding types? If not, how do we make sure there is a correct data binding? Are there any default values?

 Also, for constant values, can we do something like `"color": "black"` rather than `"color": {"value": "black"}`? The former is just shorter

sehilyi:
@wangqianwen0418 good questions.

Sometimes, `field` may not be used (e.g., `brush` mark uses `linkingId` only). But in other cases, if `field` is not provided, we do not use the corresponding incorrect channel since it is neither bound to data nor constant `value`. We do not provide a smart default value for `field`.

The `range` and `domain` are optional and we set proper default values if missing (e.g., data extent of a quantitative field for `range` and `[0, width]` to `domain` for axis channels). [[code]](https://github.com/gosling-lang/gosling.js/blob/5115859d1d46eaa0a0c278e11ef1baec51d27044/src/core/gosling-track-model.ts#L504)

Regarding `value`, I chose to use the latter due to some reason that I forgot (e.g., something like for the type-checking in the compiling part). I could revisit this at some point.

sehilyi:
Also, I will add the descriptions you mentioned previously in Slack now.

sehilyi:
@wangqianwen0418, I added the comment statements to the properties you mentioned (e.g., `y.flip` or `row.range`) as well as some other properties. Let me know if this works or need more clarifications.

wangqianwen0418:
@sehilyi , thanks a lot for the updates!
One quick question, why do visual channels like `strokeWidth`, `Opacity` have `baseline`, `zeroBaseline` properties? Since you comment the  `baseline` as "custom baseline of the y-axis", it is a little confusing to me

sehilyi:
@wangqianwen0418 Sorry for the confusion, I copy-and-pasted the comment without considering that it includes explanations about the y-axis. The `baseline` can be used for other channels, e.g., if `baseline` is `3` and `domain` is undefined, then, the `domain` of a `size` would be set to `[3, max]` instead of `[min, max]`. 

That being said, I find `baseline` is currently only used for the `y` channel and is not managed for other channels. What do you think of removing `baseline` in all other channels and adding them when we find proper use cases in the future.

wangqianwen0418:
@sehilyi, thanks for the explanation :)
Yes, I would agree with removing the `baseline` property for visual channels other than y. 
As you mentioned, its function can be achieved by `domain`.

I will remove it if it sounds good to you

sehilyi:
@wangqianwen0418 Sounds good to me as well, thanks!

Issue #565, opened 2021-10-24
`Encoding` type not included in generated JSON schema


manzt:
Although we created a new `Encoding` type in #533, it doesn't seem to end up as a definition in the generated JSON schema. I'm not sure if there is away to force this behavior, but it would be useful for

- https://github.com/gosling-lang/gos/pull/71

sehilyi:
~~Just realized we have not released `gosling.js` after we made the update on the `Encoding`. I can release the latest version so that the updated schema can be visible to the `gos` package.~~

@manzt I just released v0.9.9, and the updated `Encoding` should be accessible. Apologize that it took some time to release the new version.

manzt:
No worries! I just checked out https://github.com/gosling-lang/gosling.js/blob/v0.9.9/schema/gosling.schema.json, but still can't seem to find `Encoding` unfortunately.

sehilyi:
As part of the post-v1 release, we are cleaning up GitHub issues. Please feel free to reopen this issue if you believe it is still relevant!

PR #575, opened 2021-11-01
feat!: merge axis channels (e.g., `x` and `xe`); group channels as an `encoding` property

### LINKED ISSUE #482: Group channels as `encoding`
... FWIW, I don't think visibility exactly belong in `.encode(...)` and rather `.properties` would be more appropriate usage currently. 

The ambiguity arises because encoding fields (for simplicity, fields with type `Channel`) are not nested under a field themselves but alongside `"mark"`, `"style"`, `"dataTransforms"`, `visibility`, etc.

For this reason, `.encode(...)` must accept any key-word argument. For example, you could equivalently pass `width`, `height`, `data` to `.encode` as well as any `gos.Channel` to `.properties(...)`. 

I worry that this could become confusing for users, but adding this additional `mixin` should certainly prevent users from specifying `visibility` in `.encode` or `.properties`.

As a side note, if gosling were to group channels within a key (e.g. `"channel"` or `"encoding"`), then it would be easier to restrict the usage.

```
{
  data: { ... },
  width: { ... },
  height: { ... },
  mark: "bar",
  channels: { x: ..., y: ... },
  visibility: [ ... ],
  dataTransform: [ ... ],
}
```

In this case, you can't set `x` directly in `gos.Track(...)` or `gos.Track(...).properties(...)`. I know this would be a somewhat substantial schema change, but something worth considering in my opinion.

_Originally posted by @manzt in https://github.com/manzt/gos/issues/34#issuecomment-904798525_

Recently, I actually also ran into an issue where I can get benefit from grouping channels (e.g., `encoding: { x... }`). Also, the only reason for not grouping channels I think was to avoid deeper hierarchy, but I don't think this is a critical issue here.

I think I can change the schema (and other related codes and examples) in the Gosling and open a PR for this.

_Originally posted by @sehilyi in https://github.com/manzt/gos/issues/34#issuecomment-904857604_

### LINKED ISSUE #506: More precise encoding types

sehilyi:
Fix #506
Fix #482

## Summary of Changes
In addition to #533, this PR changes grammar around visual channels.

This PR contains a relatively large amount of changes, but many parts are changes on the examples (`src/example/*.ts`) and tests (`*.test.ts`) reflecting the grammatical changes that have been introduced in this PR. This PR introduces the following major grammatical changes.

### C1. Axis channels are combined
from:

```js
x: { field: 'start', type: 'genomic', axis: 'top', linkingId: 'overview' },
xe: { field: 'end', type: 'genomic', axis: 'top' }
```

to:

```js
x: { startField: 'start', endField: 'end', type: 'genomic', axis: 'top', linkingId: 'overview' }
```

This way, axis channels convey a more clear meaning: 
- multiple axes do not exist for the x-axis (i.e., `x` and `xe`) or the y-axis (i.e., `y` and `ye`). 
- Instead, multiple fields can be used for a single axis (i.e., `startField` and `endField` in an `x` channel).

More importantly, this prevents users from making uncertain specs, such as
- using different types for the single axis (e.g., `nominal` field for `y` and then `quantitative` field for `ye`)
- defining properties multiple times for a single axis that should be defined only once (e.g., `axis: 'top'` or `linkingId: 'overview'`):

```js
x: { field: 'start', type: 'genomic', axis: 'top', linkingId: 'overview' },
xe: { field: 'end', type: 'genomic', axis: 'bottom', linkingId: 'ab' }
```

### C2. The `encoding` property is added
Track now has a `encoding` property that wraps all channels:

```js
tracks: [{
   mark: 'line',
   encoding: {
      x: { startField: 'start', endField: 'end', type: 'genomic', axis: 'top', linkingId: 'overview' },
      color: { value: 'black' }
   },
   width: 100,
   height: 30
}]
```

This is consistent with the Vega-Lite and is beneficial in **gos** (https://github.com/gosling-lang/gos/issues/34#issuecomment-904798525).

### C3. `TemplateTrack` modified accordingly
As we made a grammatical change to channels (i.e., `x` and `xe` --> `x: { startField: ...., endField: ..., ...}`), the template mapper logic was changed as well:

```js
// part of a TemplateTrackDef spec
x: { base: { startField: 'startPosition', endField: 'endPosition', ... }
```

### C4. `DataTrack` is removed entirely
We did not support `DataTrack` anymore, but it was still included in our code. I removed all related codes. We could support a similar feature with Track Templates in Gosling in the future:

```js
{ template: 'vector-track', data: { ... }, width: 1000, height: 30}
```

## To Test Locally
Run the following command to update your local `gosling.schema.json` since this PR updates the grammar.
```sh
yarn schema
```

## BREAKING CHANGE
This PR introduces breaking changes. To change your existing specification for previous versions (~v0.9.9), you need to change your spec following the instructions described below.

If you are using multiple fields for a single axis in your spec:

```js
x: { field: 'start', type: 'genomic', axis: 'top', linkingId: 'overview' },
xe: { field: 'end', type: 'genomic', axis: 'top' }
```

You need to combine them to a single axis using up to four fields (i.e., `startField`, `endField`, `startField2`, `endField2`):

```js
x: { startField: 'start', endField: 'end', type: 'genomic', axis: 'top', linkingId: 'overview' }
```

Single-field axes are not changed (i.e., `field` properties are used):

```js
x: { field: 'position', type: 'genomic', axis: 'top', linkingId: 'overview' }
```

Add an `encoding` property to all tracks.

From:

```js
tracks: [{
   data: ...,
   mark: ...,
   x: ...,
   y: ...,
   ...
}]
```

To: 
```js
tracks: [{
   data: ...,
   mark: ...,
   encoding: {
      x: ...,
      y: ...,
      ...
   },
   ...
}]
```

**TODO**
- [x] Update all example specs that use multiple fields for a single axis
- [x] Update encoding parts of all marks
- [x] Update all tests
- [ ] Rename `'|xe - x|'` threshold to `'assignedWidth'`?


src/index.ts
```diff
@@ -2,7 +2,7 @@ export { name, version } from '../package.json';
 import GoslingSchema from '../schema/gosling.schema.json';
 import ThemeSchema from '../schema/theme.schema.json';
 
-export type { GoslingSpec, TemplateTrackDef } from './core/gosling.schema';
+export type { GoslingSpec, Encoding, TemplateTrackDef } from './core/gosling.schema';
```
sehilyi, line 5: Would need to ensure to have `Encoding` in the `schema.json` file.

Reference: https://github.com/gosling-lang/gosling.js/issues/565#issuecomment-951922456

src/core/gosling.schema.guards.ts
```diff
@@ -211,17 +219,43 @@ export function IsStackedChannel(track: SingleTrack, channelKey: keyof typeof Ch
...
     }
+    // Otherwise, just return `undefined`
     return undefined;
 }
 
 export function getChannelKeysByAggregateFnc(spec: SingleTrack) {
     const keys: (keyof typeof ChannelTypes)[] = [];
     SUPPORTED_CHANNELS.forEach(k => {
-        const c = spec[k];
+        const c = spec.encoding[k];
```
sehilyi, line 247: Aggregating `startField` and `endField` separately should be still possible even when we group `x` and `xe`.

Reference: An ideogram example with alternating backgrounds by chromosomes.

src/core/gosling-track-model.ts
```diff
@@ -599,6 +600,7 @@ export class GoslingTrackModel {
                     }
                 } else if (IsChannelDeep(channel) && (channel.type === 'quantitative' || channel.type === 'genomic')) {
                     if (channel.domain === undefined && field) {
+                        // TOOD: do this for all fields
```
sehilyi, line 603: `domain` should be the union of `startField`, `endField`, `startField2`, `endField2` if multiple fields are used.

src/core/compile.ts
```diff
@@ -15,9 +15,6 @@ export function compile(
     // Make sure to keep the original spec as is
     const _spec = JSON.parse(JSON.stringify(spec));
 
-    // Override default visual encoding (i.e., `DataTrack` => `BasicSingleTrack`)
-    overrideDataTemplates(_spec);
```
sehilyi, line 19: Removed this since `DataTrack` had been deprecated.

src/core/compile.ts
```diff
@@ -15,9 +15,6 @@ export function compile(
...
-
     // Replace track templates with raw gosling specs (i.e., `TemplateTrack` => `SingleTrack | OverlaidTrack`)
     replaceTrackTemplates(_spec, templates);
```
sehilyi, line 19: TODO: Templates should use startField and endField instead of x and xe

src/core/gosling.schema.ts
```diff
@@ -459,9 +436,11 @@ export interface ChannelDeepCommon {
     field?: string;
 }
 
-export interface X extends ChannelDeepCommon {
```
sehilyi, line 462: Key changes can be found in this file.

sehilyi:
@manzt, sorry that this update was delayed. It turns out to be more significant updates than I expected. 

Can you review this PR? I would be interested in hearing from you whether the grammatical changes look okay and these updates work for you to continue working on https://github.com/gosling-lang/gos/pull/71. Please review code bases as deeply as you would like to.

manzt:
No problem! Won't be able to look today but probably tomorrow.

manzt:
@sehilyi I completely lost track of this. Will review this weekend -- sorry! Do you have any thoughts regarding how to reserve these merge conflict?

sehilyi:
@manzt No worries! I was basically thinking of manually resolving conflicts which I hope to be manageable.

Also, please let me know if all grammatical changes introduced in this PR make sense to you. Since this PR introduces breaking changes, I would like to merge this carefully.

sehilyi:
@manzt Since this will introduce breaking changes and is too far behind the current version, I am reluctant to resolve conflicts and merge this.

I think we can revisit this (at least making an `encoding` property in a `SingleTrack`) later when we plan to make a major version bump.

manzt:
Agreed!

manzt:
@sehilyi I just wanted to apologize that your hard work here never got a thorough review. I think this PR had several different aims and touched many internals of Gosling that at the time I struggled to understand.

sehilyi:
@manzt No worries. This was such a huge PR. And I am sure we can revisit this and repurpose it for future PRs.

PR #580, opened 2021-11-09
chore: add doc examples

### LINKED ISSUE #578: Move doc examples from gist to main repo
moving the basic doc examples (e.g., bar charts, pie charts) from gist to the main repo to facilitate versioning

wangqianwen0418:
move doc examples from gist to the main repo to facilitate the doc versioning
fix #578


wangqianwen0418:
In this PR:
- we move doc examples fro gist to the main gosling repo
- we change example from a list to an object, enable users to access hidden demos from an URL (e.g., https://gosling.js.org/?example=MATRIX_HFFC6)
- the spec of doc examples is a string rather than a goslingSpec object so that the comments in the spec can be reserved.

sehilyi:
@wangqianwen0418 Thank you for opening this PR!

Before I am able to look into details, I want to share several high-level questions/concerns:

- I was not sure if we want to include these examples to `gosling-lang/gosling.js` or to another repo (e.g., `gosling-lang/gosling-website`). One of the pain is that we will need to change those examples every time we change the grammar. Multiple grammar-wise changes can happen for a single release, so we will need to update examples multiple times as well. However, if these examples are placed somewhere, we can just update those examples once for a new release (e.g., `v0.10.0`) which I think is sufficient for the version control.
- I see the benefit of showing comments by defining the spec in a string format, but it makes it also difficult to edit the examples based on the grammatical change since IDE does not type-check the examples.. I wonder if we have a better solution for this..
- To "enable users to access hidden demos from an URL", I think we could just filter the example list (`.filter(d => !d.hidden)`) in the `editor/editor.tsx` instead of `editor/example/index.ts`, without changing the structure of the example (_flat JSON array_ --> _nested JSON array_). Is there any other reason that made you change the structure?

sehilyi (commented):
@wangqianwen0418 Thank you for opening this PR!

Before I get a chance to look into details, I want to share several high-level questions/concerns:

- I was not sure if we want to include these examples to `gosling-lang/gosling.js` or to another repo (e.g., `gosling-lang/gosling-website`). One of the pain is that we will need to change those examples every time we change the grammar. Multiple grammar-wise changes can happen for a single release, so we will need to update examples multiple times as well. However, if these examples are placed somewhere, we can just update those examples once for a new release (e.g., `v0.9.9` --> `v0.10.0`) which I think is sufficient for the version control.
- I see the benefit of showing comments by defining the spec in a string format, but it makes it also difficult to edit the examples based on the grammatical change since IDE does not type-check the examples.. I wonder if we have a better solution for this..
- To "enable users to access hidden demos from an URL", I think we could just filter the example list (`.filter(d => !d.hidden)`) in the `editor/editor.tsx` instead of `editor/example/index.ts`, without changing the structure of the example (_flat JSON array_ --> _nested JSON array_). Is there any other reason that made you change the structure?

wangqianwen0418:
Hi @sehilyi, many thanks for the quick reply!

- About the location of the examples, my main motivation of putting them in the main gosling repo is to keep them updated with the online gosling editor. The current doc directs users to the gosling online editor to open these examples. Basically, if you update the online editor and other examples in the example folder, you also need to update the doc examples. I agree with the pain point you mentioned, maybe we can find a way to only update the online editor and all the examples only when there is a new release?
- yes, I see the drawbacks of missing type-checking. The main reason I am using the string specs is a) to keep the comments; b) these examples are previously stored in gist and more like strings. Maybe I can run an additional spec validation function for these string specs to address your concerns? 
- About the example structure, well, I am okay to change it back. Just feel a dict is easier to access examples compared with an array.

schema/gosling.schema.json
```diff
@@ -3150,7 +3150,7 @@
           "$ref": "#/definitions/AxisPosition",
           "description": "not supported"
         },
-        "xDomain": {
```
sehilyi: Can you check updates on this file? Since `gosling.schema.ts` is not edited, I think this file should not be updated as well.

editor/editor.tsx
```diff
@@ -461,15 +480,15 @@ function Editor(props: any) {
...
-                <span className="demo-dropdown" hidden={urlSpec !== null || urlGist !== null}>
+                <span className="demo-dropdown" hidden={urlSpec !== null || urlGist !== null || urlExampleId !== null}>
                     <select
                         style={{ maxWidth: IS_SMALL_SCREEN ? window.innerWidth - 180 : 'none' }}
                         onChange={e => {
-                            setDemo(examples.find(d => d.id === e.target.value) as any);
+                            setDemo(examples[e.target.value] as any);
                         }}
-                        defaultValue={demo.id}
+                        value={demo.id}
```
sehilyi, line 486: > About the example structure, well, I am okay to change it back. Just feel a dict is easier to access examples compared with an array.

I am okay with the new structure, and I think it has an advantage that it prevents us from accidentally assigning the same `id` to multiple examples. 

Since `id` and `key` are duplicated currently, how about removing `id` properties? I think `demo.id` in this line can be replaced with `urlExampleId` and we could change the following

```js
const SHOWN_EXAMPLE_LIST = Object.values(examples).filter(d => !d.hidden);
```

to

```js
const SHOWN_EXAMPLE_ENTRIES = Object.entries(examples).filter([k, d] => !d.hidden);
```

to keep the keys ([Object.entries()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries)).

editor/example/doc-examples/area.ts
```diff
@@ -0,0 +1,44 @@
+export const AREA = `{
```
sehilyi: > yes, I see the drawbacks of missing type-checking. The main reason I am using the string specs is a) to keep the comments; b) these examples are previously stored in gist and more like strings. Maybe I can run an additional spec validation function for these string specs to address your concerns?

One thing I really like in my IDE is that renaming properties in the schema is very simple, i.e., when I rename something, all properties used in other files are updated at once. But, having string-format examples will make it difficult to find corresponding changes in the examples and to update them.

If you think comments are something that is really important for describing examples, I think we can just keep them as-is (i.e., string) and revisit this in the future. But, if not, let's change these examples to JSON objects so that we can more efficiently maintain examples in this repo. It's your call!

editor/editor.tsx
```diff
@@ -154,18 +164,23 @@ function Editor(props: any) {
...
-        ? examples.map(d => d.id).indexOf(urlParams.example as string)
-        : INIT_DEMO_INDEX;
+    const urlExampleId = urlParams?.example as string;
+
+    const defaultCode = urlGist ? emptySpec() : stringify(urlSpec ?? (INIT_DEMO.spec as gosling.GoslingSpec));
 
-    const defaultCode = urlGist
-        ? emptySpec()
-        : stringify(urlSpec ?? (examples[INIT_DEMO_INDEX].spec as gosling.GoslingSpec));
+    // let defaultCode = stringify(INIT_DEMO.spec as gosling.GoslingSpec);
```
sehilyi: I was unsure if this is left by mistake, but we can remove this comment block?

editor/example/index.ts
```diff
@@ -20,149 +20,154 @@ import { EX_SPEC_BAND } from './vertical-band';
...
-    {
+    TEMPLATE: {
         name: 'Track Template',
         id: 'TEMPLATE',
         spec: EX_SPEC_TEMPLATE,
         underDevelopment: true
-    }
-].filter(d => !d.hidden);
+    },
+    ...DOC_EXAMPLES
```
sehilyi: What do you think of adding doc-related examples (i.e., examples in `DOC_EXAMPLES`) directly here so that we can maintain examples in a single file? Perhaps we can add a property, like `doc: boolean` if needed.

sehilyi (commented):
> About the location of the examples, my main motivation of putting them in the main gosling repo is to keep them updated with the online gosling editor. The current doc directs users to the gosling online editor to open these examples. Basically, if you update the online editor and other examples in the example folder, you also need to update the doc examples. I agree with the pain point you mentioned, maybe we can find a way to only update the online editor and all the examples only when there is a new release?

Thanks for explaining this which allowed me to understand the motivation clearly! I agree that since these examples will be opened in the editor, we would want to have this here. We can move examples once we separate the editor from gosling.js. 

I've added more detailed comments that are mostly related to maintenance. Let me know what you think about that.

In addition, it looks the example dropdown menu does not appear. Can you double-check that and let me know once it is fixed so that I can test the Editor again?

Screen Shot 2021-11-11 at 1 04 32 PM

Lastly, what do you think of merging some examples? It looks many doc-related examples are duplicated with existing examples (e.g., basic marks and semantic zooming), so was wondering if we would want to encapsulate them for better maintenance (e.g., define `function getBarTrack (width: number, ...) {}` and use them in an editor example and documentation). (if you agree to make examples to actual JSON objects).

editor/editor.tsx
wangqianwen0418: Sorry, the comment is left here by mistake and I will remove it.

editor/example/doc-examples/area.ts
wangqianwen0418: Well, in this case, I think an object is better than a string for code maintenance.  No problem, I will change them to JSON objects

editor/editor.tsx
wangqianwen0418: Sounds great! Will change it accordingly

schema/gosling.schema.json
wangqianwen0418: Sure!

wangqianwen0418:
@sehilyi, thanks so much for your feedback! They are very helpful!
Sure, I will check the drop-down menu and other things you mentioned in the code review.

> what do you think of merging some examples? It looks many doc-related examples are duplicated with existing examples (e.g., basic marks and semantic zooming), so was wondering if we would want to encapsulate them for better maintenance (e.g., define function getBarTrack (width: number, ...) {} and use them in an editor example and documentation). (if you agree to make examples to actual JSON objects).

I think this is a great idea. Actually, there is not a huge difference between examples and doc examples (doc examples might be much simpler though). Merging and encapsulating them will definitely help the code maintenance. I will first change the doc examples to JSON objects first and then see how can we better maintain these examples

editor/example/index.ts
wangqianwen0418: Sounds great~ Maybe no need to add another property, I can just assign their keys as `doc_xxx` and set `hidden: true`

sehilyi:
@wangqianwen0418 Sounds great to me!

editor/example/doc-examples/index.ts
```diff
@@ -0,0 +1,13 @@
...
+export { BRUSH } from './brush';
+export { LINK } from './link';
+export { LINKING_TRACKS } from './linking-tracks';
+export { OVERLAY_TRACKS_BAR_POINT } from './overlay-tracks-bar-point';
+export { OVERLAY_TRACKS_LINE_POINT } from './overlay-tracks-line-point';
+export { OVERLAY_TRACKS_RECT_TEXT } from './overlay-tracks-rect-text';
+
+export { RECT } from './rect';
+export { SEMANTIC_ZOOM_SEQUENCE, SEMANTIC_ZOOM_CYTO } from './semantic-zoom';
+// export { SEMANTIC_ZOOM_CYTO } from './semantic-zoom-cyto';
```
sehilyi: I think we can remove this line
```suggestion
```

editor/example/index.ts
```diff
@@ -20,149 +20,203 @@ import { EX_SPEC_BAND } from './vertical-band';
...
         underDevelopment: true
     },
-    {
+    TEMPLATE: {
         name: 'Track Template',
-        id: 'TEMPLATE',
         spec: EX_SPEC_TEMPLATE,
         underDevelopment: true
+    },
+    // ...DOC_EXAMPLES
```
sehilyi: We could add a descriptive comment, replacing the previous code left here.

```suggestion
    // Followings are doc examples that are only accessible via URLs using the key, such as
    // https://gosling.js.org/?example=doc_area
```

schema/gosling.schema.json
```diff
@@ -428,7 +428,7 @@
           "description": "Determine the colors that should be bound to data value. Default properties are determined considering the field type."
         },
         "type": {
-          "description": "Data type",
+          "description": "Specify the data type",
```
sehilyi: Could you double-check this file again? I think this file should be unchanged in this PR.

sehilyi (commented):
@wangqianwen0418, thank you for the updates! I just left minor comments regarding comment statements and an unnecessary change on `gosling.schem.json`, but other than that this looks great! I also tested locally with diverse types of URLs (i.e., `example=`, `gist=`).

sehilyi:
FYI, I made a small change (https://github.com/gosling-lang/gosling.js/pull/580/commits/2fd34986c2bf2447310e13725232b77925649581, https://github.com/gosling-lang/gosling.js/pull/580/commits/9943eb5141aa2ee507ed38fac1892ceb95d95b17) to address conflict on the `editor/example/index.ts` that was introduced by #586 and to remove a `circularLink` use in the `Link.ts` example since the property was deprecated.

schema/gosling.schema.json
wangqianwen0418: Not very sure about what caused this problem. But the updated description "Specify the data type" is the right description based on the comments in the gosling.schema.ts in this PR.

wangqianwen0418:
Thanks a lot, @sehilyi! 
Also, maybe we should start thinking about separating the online editor from the main gosling repo to better version it as well as these examples

PR #591, opened 2021-11-23
feat: allow aggregation functions (`mean` and `sum`) for vector, bigwig, and multivec

### LINKED ISSUE #587: Support diverse aggregation functions, such as `sum` and `mean`, for HiGlass files (e.g., `vector`)
Calculating average over tileUnitSize should be removed when the data type is vector because higlass has done the calculation once somehow.  Doing it twice causes the y-axis domain to show incorrect values.
The changes I've made to the code is shown in the following figure.

image

sehilyi:
Fix #587 

```js
"data": {
  "type": "multivec", // or "vector" and "bigwig"
  ...,
  "aggregation": "sum" // or "mean" which is default
},
```

### DIFF
--- a/schema/gosling.schema.json
+++ b/schema/gosling.schema.json
@@ -230,9 +230,7 @@
     "BinAggregate": {
       "enum": [
         "mean",
-        "sum",
-        "min",
-        "max"
+        "sum"
       ],
       "type": "string"
     },
--- a/schema/gosling.schema.json
+++ b/schema/gosling.schema.json
@@ -186,6 +186,10 @@
     "BIGWIGData": {
       "additionalProperties": false,
       "properties": {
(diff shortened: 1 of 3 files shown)


zhangzhen:
@sehilyi this fix hasn't been merged into 0.9.11?

Zhen

sehilyi:
@zhangzhen, sorry for the late response.

Since we want to allow using either `mean` or `sum` values, I added an `aggregation` property to `bigwig`, `vector`, and `multivec` data types. While `mean` is the default value (i.e., the current value we use in the editor), you can use `aggregation: "sum"` to prevent calculating mean values. This will be merged into the master branch and will be released today. Let me know if this change works or not for your use case.

PR #612, opened 2021-12-15
feat: support native matrix visualization

sehilyi:
This PR adds native support of `matrix` data. Example of drawing matrix using `matrix` format data:

```ts
{
     data: {
          url: GOSLING_PUBLIC_DATA.matrixMicroC,
          type: 'matrix'
      },
      mark: 'bar',
      x: { field: 'xs', type: 'genomic', axis: 'none' },
      xe: { field: 'xe', type: 'genomic', axis: 'none' },
      y: { field: 'ys', type: 'genomic', axis: 'none' },
      ye: { field: 'ye', type: 'genomic', axis: 'none' },
      color: { field: 'value', type: 'quantitative', range: 'warm' },
      width: 600,
      height: 600
}
```

## Screenshots

Screen Shot 2021-12-16 at 5 48 01 PM

https://user-images.githubusercontent.com/9922882/146459349-a7f10f91-2476-40f6-8f3e-e99b488946bc.mov

**There are multiple follow up things to do:**
- [ ] Axis on 2D tracks seem to be broken. This needs to be fixed.
- [ ] Improving the rendering performance. Zooming is slow with dense Hi-C data.
- [ ] Supporting log scale color (e.g., `color: { ..., scale: 'log' }`)
- [ ] Supporting horizontal rules. Currently, we support vertical rules (see the `MATRIX` example).
- [ ] Somehow, the right-top part along the diagonal is zero values. Need to confirm if data itself only contains the left-bottom part to reduce performance. In this case, we need to properly manipulate data when loaded.


editor/example/index.ts
```diff
@@ -53,6 +54,11 @@ export const examples: {
         name: 'Basic Example: Rule Mark',
         spec: EX_SPEC_RULE
     },
+    MATRIX: {
+        name: 'Basic Example: Hi-C Matrix',
+        spec: EX_SPEC_MATRIX,
+        forceShow: true
```
sehilyi:
```suggestion
        forceShow: false
```

I will turn this off after the review so that our public editor does not show this example by default.

sehilyi:
@manzt May I assign you as a reviewer of this PR? 

I think the `cooler` + `rule` examples would be nice use cases for `gos`, and this PR will be one of the core updates toward that.

There are multiple things that I would need to explore/fix (which I listed above), but I would like to make this PR small and incrementally address them with follow-up PRs.

manzt:
Yes! Add me as a reviewer and I'll get to it tomorrow.

editor/example/matrix-hffc6.ts
```diff
@@ -242,9 +242,11 @@ export const EX_SPEC_MATRIX_HFFC6: GoslingSpec = {
...
                                         type: 'matrix'
                                     },
-                                    mark: 'rect',
-                                    x: { field: 'position1', type: 'genomic', axis: 'none' },
-                                    y: { field: 'position2', type: 'genomic', axis: 'none' },
+                                    mark: 'bar',
+                                    x: { field: 'xs', type: 'genomic', axis: 'none' },
+                                    xe: { field: 'xe', type: 'genomic', axis: 'none' },
+                                    y: { field: 'ys', type: 'genomic', axis: 'none' },
+                                    ye: { field: 'ye', type: 'genomic', axis: 'none' },
```
manzt, line 249: Just so I understand, are these fields (`xs`, `xe`, `ys,` `ye`, and `value`) now implicitly defined for `matrix` data-type?

src/gosling-track/data-abstraction.ts
```diff
@@ -199,6 +201,68 @@ export function getTabularData(
...
             });
         });
+    } else if (spec.data.type === 'matrix') {
+        if (!data.dense || typeof data.tileY === 'undefined' || typeof data.tileHeight === 'undefined') {
+            // we do not have sufficient data.
+            return;
+        }
+
+        // width and height of the tile
+        const numBins = Math.sqrt(data.dense.length);
```
manzt, line 211: same a tile width/height or no?

manzt: op, nevermind... seeing below...

manzt: is this ever not the same as the tilewidth, though?

src/gosling-track/data-abstraction.ts
```diff
@@ -199,6 +201,68 @@ export function getTabularData(
...
+                for (let r = 0; r < aggLen; r++) {
+                    const curVal = numericValues[(yIndex + r) * numBins + (xIndex + c)];
+                    if (!isNaN(+curVal)) {
+                        if (isNaN(value)) {
+                            value = 0;
+                        }
+                        value += curVal;
+                    }
+                }
+            }
```
manzt: Could you leave a comment on the logic here. It took me a while to parse what was going on. My understanding is.

Calculate the _sum_ of a 2D `aggLen x aggLen`. If and only if every item in the region is `NaN`, `value===NaN` in the end. Otherwise treat `NaN` values as `0` and calculate the sum with defined numbers in the region? If so, is summation always the aggregation that makes sense?

src/gosling-track/data-abstraction.ts
```diff
@@ -199,6 +201,68 @@ export function getTabularData(
...
+
+            if (isNaN(value)) {
+                // this is a missing value
+                continue;
+            }
+
+            value /= aggSize;
```
manzt: Oh i see. So its the mean?

src/gosling-track/data-abstraction.ts
```diff
@@ -199,6 +201,68 @@ export function getTabularData(
...
+            const y = (ys + ye) / 2.0;
+            tabularData.push({
+                value,
+                [columnField]: x,
+                [`${columnField}s`]: xs,
+                [`${columnField}e`]: xe,
+                [rowField]: y,
+                [`${rowField}s`]: ys,
+                [`${rowField}e`]: ye
+            });
```
manzt, line 265: This is an interesting tradeoff. So we transforms a dense numeric array into an array of objects? While I understand that this provides much more flexibility with regard to rendering, converting to this data-structure and then rendering is probably very expensive, and the default use case I assume suffers somewhat with regard to performance.

manzt (approved):
I think this is coming along very nicely. One thing to keep in mind is that `matrix` / tileset data is handled _very_ well by HiGlass, and it seems adding this feature to Gosling comes at the expense of that functionality.

Generally, I wonder if we can keep the HiGlass rendering of `matrix` data sources while somehow exposing the ability to overlay over marks `y` & `x` marks. Converting dense `matrix` data (higlass server tileset) to tabular data is an expensive operation and noticeable in the UI.

I think the tension here is that Gosling juggles both columnar data-types _and_ dense arrays in the special case of the `matrix` data source. I'm curious why we couldn't some how keep this special handling of `matrix` data (via HiGlass) and then allow columnar-based datasets to overlaid on the `x` and `y` axes, since both are type `genomic`

manzt:
@sehilyi sorry, I mean to "comment" not approve these changes without further discussion.

editor/example/matrix-hffc6.ts
sehilyi: Yes, that's true!

I think I would need to add proper documentation for this.

src/gosling-track/data-abstraction.ts
sehilyi: Yes, I did not fully understand why this happens, but this is sometimes different from `tileWidth`.

src/gosling-track/data-abstraction.ts
sehilyi: Yes, it is

src/gosling-track/data-abstraction.ts
sehilyi: Yes, your understanding is correct and it calculates the mean. 

Actually, I think this should be summation instead of mean since cooler / matrix data calculates summation, as far as I understand correctly.

We can also support other types of aggregation functions in case a user uses matrix data with different aggregation functions.

src/gosling-track/data-abstraction.ts
sehilyi: This is a great question. There is certainly a strong tradeoff between flexible encoding and efficient rendering.

In this PR, I went with flexibility since the core feature of our grammar is to support expressiveness. And to support expressive encoding, some kinds of data abstraction (here, converting to columnar formats) is necessary. In this way, one can fully use the concept of our grammar, like using other mark types (e.g., point w/ varying size and color), data transformation (e.g., filter data points with very strong signals), or applying semantic zooming (e.g., showing text labels when zoomed in very far).

I think there is some space to improve the performance (especially on the rendering pipeline), but I don't currently have a concrete idea on how to make it happen, both in the data abstraction and rendering, without sacrificing the flexibility. I would need to test this in-depth. One thing I would need to refer to is the rendering pipeline of GenomeSpy since it looks to work very smoothly with columnar data (CSV files).

Additionaly, I think, we can support the fast HiGlass-based matrix rendering with a `template` (e.g., `{ template: "matrix", data: { /* Cooler data */ } }`). Since this is a template (not a grammar-based approach), we do not need to worry about whether a user change encoding that is not defined by the template (e.g., changing rect to point). We will be able to allow overlaying annotations as well (e.g., `alignment: "overlay", tracks: [ { /* matrix template */ }, { /* annotation rules */ }]`).

sehilyi:
@manzt Thanks a lot for reviewing this PR! 

Regarding your last comment, which I think is really important, I wonder if we can support more efficient rendering without sacrificing the expressivity of our grammar. I would need to investigate further in the next few weeks. Let me know also if you have any idea about that!

src/gosling-track/data-abstraction.ts
sehilyi: To add, my main idea on this PR was to perform the data abstraction for the expressive encoding (since it is necessary for the grammar), but improve the performance (especially on the rendering) as a follow-up PRs, to make the zooming/panning smooth. When I tested locally and informally, the rendering pipeline seemed to be the main bottleneck.

Let me know if it makes more sense for you to make this PR pending until we improve the performance, or you have any other better ideas to workaround this issue.

manzt: Thanks for the detailed response! I agree (mostly) with the prioritization of flexibility over efficiency in this regard. I guess my point, specifically, is questioning why we map the dense matrix data to `tabularData`. 

I understand why having this mapping is important for columnar data (that you want to map to track with `x` === `genomic` and `y` === `genomic`, but I'm struggling to see what flexibility this provides in the context of dense matrix data. 
 
I guess now you could do something like,

```javascript
{
     data: {
          url: GOSLING_PUBLIC_DATA.matrixMicroC,
          type: 'matrix'
      },
      mark: 'point',
      x: { field: 'xs', type: 'genomic', axis: 'none' },
      y: { field: 'ys', type: 'genomic', axis: 'none' },
      color: { field: 'value', type: 'quantitative', range: 'warm' },
}
```

but it isn't clear to me how this is useful. 

I bet I'm conceptually missing the vision of this PR, but I think ideally we should make the mapping of matrix -> tabular data "lazy" (e.g. only perform this conversion if we need to render something other than a standard heatmap). For example, if the configuration below is supplied, then we should just let HiGlass load the tiles natively,

```javascript
{
     data: {
          url: GOSLING_PUBLIC_DATA.matrixMicroC,
          type: 'matrix'
      },
      mark: 'bar',
      x: { field: 'xs', type: 'genomic', axis: 'none' },
      xe: { field: 'xe', type: 'genomic', axis: 'none' },
      y: { field: 'ys', type: 'genomic', axis: 'none' },
      ye: { field: 'ye', type: 'genomic', axis: 'none' },
      color: { field: 'value', type: 'quantitative', range: 'warm' },
      width: 600,
      height: 600
}
```

In fact, the higlass server is already doing some work to convert columnar like data (sparse matrices) to dense arrays and we are in a sense undo-ing that on the client.

I think to your point, it's important to remember that we can make the grammar expressive (allowing a wide variety of visualizations), but inherently _some_ of the possible visualizations will be more inefficient to render from others. In the implementation of the renderer, however, I think we should be careful to consider/optimize for common visualizations if possible.

manzt:
> Regarding your last comment, which I think is really important, I wonder if we can support more efficient rendering without sacrificing the expressivity of our grammar. I would need to investigate further in the next few weeks. Let me know also if you have any idea about that!

I agree. I think the key here is that that grammar can (and should) appear flexible and consistent to end-users, but we are free to implement the rendering however. The point of the declarative grammar is that the _how_ of creating a visualization is abstracted away. 

From a Gosling spec, we know _what_ data are being used and exactly which mappings are desired. Because the `matrix` data type is designed to be rendered directly as a dense array, if the user wants this type of visualization, we shouldn't try to convert the data to some tabular representation. Instead, we should handle this special (but very common) case.

src/gosling-track/data-abstraction.ts
sehilyi: @manzt Thanks for the reply. Having a conversation on this issue is really helpful! 

The vision of this PR is to fully support the grammar approach. The encoding supported in HiGlass `matrix` is a very small space in the entire design space in our grammar (i.e., rectangular shape mark + quantitative color + two genomic axes), and we will need to figure out how to support the remaining spaces. For example, if one just adds stroke colors of marks or adjusts their sizes, we will not be able to use HiGlass due to its limited rendering expressiveness. I think there are many other encodings that users can try with using our grammar other than the regular matrix. One could try displaying arcs with colors in 1D tracks (e.g., filter to show strong interactions) [^1], a bubble matrix [^2], or the superposition of two matrices with varying sizes for pair-wise comparison [^3]. Supporting semantic zooming will be another example (e.g., when dense, show the regular matrix representation, but when sparse, change representations and show details with labels).

Having said that, I agree with you that the regular matrix visualization in HiGlass will be the common encoding with the `matrix` data, and we will need to handle this nicely. I think, instead of using `tabularData` all the time, we can use the HiGlass rendering when a user uses the following spec. Until we find an efficient way to render the matrix directly on Gosling (e.g., using the lazy data processing), we can keep this logic in our grammar.

```
{
     data: {
          url: GOSLING_PUBLIC_DATA.matrixMicroC,
          type: 'matrix'
      },
      mark: 'bar',
      x: { field: 'xs', type: 'genomic', ... },
      xe: { field: 'xe', type: 'genomic', ... },
      y: { field: 'ys', type: 'genomic', ... },
      ye: { field: 'ye', type: 'genomic', ... },
      color: { field: 'value', type: 'quantitative' },
      width: ...,
      height: ...
}
```

To unify the HiGlass matrix with our grammar, I will need to test several things, such as overlaying gosling tracks on top of the higlass matrix (e.g., vertical and horizontal annotations) or using consistent color scales in the HiGlass matrix. Also, let me think more about `tabularData`. Perhaps, we can even find a way to use the `tileData` from the HiGlass server as-is, without converting to non-columnar data at all, for rendering.

I can make changes on this PR to use the conditional HiGlass rendering, and I can test other issues in the follow-up PRs. What do you think?


[^1]: image
[^2]: Screen Shot 2021-12-20 at 9 40 09 AM
[^3]: Screen Shot 2021-12-20 at 9 43 31 AM

manzt: > Having a conversation on this issue is really helpful!

Agreed!

> The vision of this PR is to fully support the grammar approach. 

I understand this now, and thank you for all the motivating examples. This will be great to have for future reference.

> I think, instead of using `tabularData` all the time, we can use the HiGlass rendering when a user uses the following spec. 

I think the idea of a conditional data transformation (depending on the spec) is a great compromise. This will be hidden from end users, offering flexibility in the grammar and efficiency when the data don't require transformation.

> I can make changes on this PR to use the conditional HiGlass rendering, and I can test other issues in the follow-up PRs. What do you think?

I'm happy to have this merged, but this sounds like a great plan in a follow up PR. Preferably this can be explored sooner than later, I think.

manzt (approved):

src/gosling-track/data-abstraction.ts
sehilyi: Removed the average calculation and added comments

sehilyi:
@manzt Thank you so much for all feedback! I have updated a bit to add comment statements and remove unneeded the mean calculation, reflecting your comments. I will open a follow-up PR to use the HiGlass matrix for certain cases.

manzt:
Great! Thanks.

PR #613, opened 2021-12-17
feat!: `quantitativeFields` is removed entirely from `csv` and `json` data specs

### LINKED ISSUE #579: More concise (text-based) columnar data definitions
Apologies for the lack on context behind the choice, but I don't know if I understand the motivation for including  `quantitativeFields`, `genomicFields`, and `chromosomeField` in the [CSV data definition](http://gosling-lang.org/docs/data#csv-no-higlass-server). These fields aren't marked as required in the docs, but every example I've seen includes their use.

### Motivation

To my knowledge, these columns inform how the CSV is parsed but this interpretation is also captured elsewhere in the track definition (type: genomic, quantitative, categorical, etc), so really it's an abstraction leak. I see how the `chromosomeField` is currently necessary, but I'm curious if that information could also be captured in the "genomic" type rather than the data definition.

As a motivating example, what is the expected behavior if I use a field type that differs from the data definition? I assume the track type takes precedent, and if so we don't need the data definition since a type is required on all tracks.

```python
import gosling as gos

data = gos.csv('./data.csv', genomicFields=['start', 'end'], chromosomeField='chr', quantitativeFields=['value'])

gos.Track(data).encode(x=gos.Channel('start:G'), y=gos.Channel('value:N')) # value doesn't match data definition
```

### Proposal

Remove `quantitativeFields`, `genomicFields`, and _maybe_ `chromosomeField` from CSV definition. This would make specifying CSV data more concise and avoid the case where data definition does not match track definition.

```python
import gosling as gos
import pandas as pd

data = gos.csv('./data.csv')
data = pd.read_csv('./data.csv').gos.csv() # no arguments needed
```

### Approach

Use build-in (d3?) auto-parsing of CSV into memory. We can coerce any data-types that are mis-interpreted based on the track definition. Perhaps as an extension to #575, we can think about if there is a way to include chromosome field in the `X` encoding definition.

```typescript
interface X {
  type: 'genomic';
  // tuples represent chromosome position OR chromosome region
  field: [chrom: string, start: string] | [chrom: string, start: string, end: string];
  // ...
}
```

sehilyi:
This removes `quantitativeFields` entirely from `csv` and `json` data specs. This property was unnecessary since we generate `scale`s per visual channel, and the value type (i.e., quantitative vs. nominal) is determined at the time we generate scales, not at the time we parse data. More detailed discussions can be found at #579.

# Breaking Updates
If you were using `quantitativeFields` in the data properties, you just need to remove them from your spec:
```diff
data: {
    url: 'https://raw.githubusercontent.com/sehilyi/gemini-datasets/master/data/cytogenetic_band.csv',
    type: 'csv',
    chromosomeField: 'Chr.',
    genomicFields: ['ISCN_start', 'ISCN_stop', 'Basepair_start', 'Basepair_stop'],
-   quantitativeFields: ['Band', 'Density']
},
```

Fix #579

### DIFF
--- a/editor/example/cancer-variant.ts
+++ b/editor/example/cancer-variant.ts
@@ -603,8 +603,7 @@ export function view(sample: string): GoslingSpec {
             //         url: `https://s3.amazonaws.com/gosling-lang.org/data/cancer/substitution.${sample}.csv`,
             //         type: 'csv',
             //         chromosomeField: 'Chrom',
-            //         genomicFields: ['Pos'],
-            //         quantitativeFields: ['GP', 'SP', 'DP', 'PosDiff']
+            //         genomicFields: ['Pos']
             //     },
             //     dataTransform: [
             //         { type: 'concat', fields: ['Ref', 'Alt'], separator: ' > ', newField: 'sub' },
--- a/editor/example/corces.ts
+++ b/editor/example/corces.ts
@@ -23,8 +23,7 @@ export const EX_SPEC_CORCES_ET_AL: GoslingSpec = {
                         url: 'https://raw.githubusercontent.com/sehilyi/gemini-datasets/master/data/cytogenetic_band.csv',
                         type: 'csv',
                         chromosomeField: 'Chr.',
-                        genomicFields: ['ISCN_start', 'ISCN_stop', 'Basepair_start', 'Basepair_stop'],
-                        quantitativeFields: ['Band', 'Density']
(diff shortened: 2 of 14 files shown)


manzt (approved):
Great! LGTM. 

~I know we chatted about this previously, but can you remind me why genome fields are necessary again? Is it something about calculating the "absolute" genomic position?~ found in previous discussion!

sehilyi:
I think I need to add a proper description about the genomics fields to our docs.

PR #615, opened 2022-01-06
feat: use HiGlass `heatmap` for conventional 2D matrix

sehilyi:
This PR allows to use HiGlass `heatmap` tracks conditionally (i.e., for given user specs) so that the conventional 2D matrix, which is one of the most common use cases w/ Hi-C, can be displayed in an efficient way. All the rendering and data managing is handled by the [`heatmap` track](https://github.com/higlass/higlass/blob/develop/app/scripts/HeatmapTiledPixiTrack.js).

(Find more detailed discussion: https://github.com/gosling-lang/gosling.js/pull/612#discussion_r771623844)

https://user-images.githubusercontent.com/9922882/148417261-2b16d0bf-2f5a-4224-8621-6379d43cf06b.mov

### DIFF
--- a/src/core/gosling-to-higlass.ts
+++ b/src/core/gosling-to-higlass.ts
@@ -130,7 +130,7 @@ export function goslingToHiGlass(
 
         // We use higlass 'heatmap' track instead of 'gosling-track' for rendering performance.
         // HiGlass is really well-optimized for matrix visualization, and rendering it in Gosling
-        // makes the zooming interaction slow.
+        // instead makes the zooming interaction slow.
         // See https://github.com/gosling-lang/gosling.js/pull/612#discussion_r771623844
         if (IsHiGlassMatrix(firstResolvedSpec)) {
             // By changing the track type, HiGlass uses its native heatmap track
--- a/src/core/gosling-to-higlass.ts
+++ b/src/core/gosling-to-higlass.ts
@@ -136,7 +136,7 @@ export function goslingToHiGlass(
             // By changing the track type, HiGlass uses its native heatmap track
             hgTrack.type = 'heatmap';
             const colorStr =
-                IsChannelDeep(firstResolvedSpec.color) && typeof firstResolvedSpec.color === 'string'
+                IsChannelDeep(firstResolvedSpec.color) && typeof firstResolvedSpec.color.range === 'string'
                     ? firstResolvedSpec.color.range
                     : 'viridis';
             hgTrack.options.colorRange = getHiGlassColorRange(colorStr);
--- a/src/core/gosling-to-higlass.ts
+++ b/src/core/gosling-to-higlass.ts
@@ -128,7 +128,7 @@ export function goslingToHiGlass(
             };
         }
 
-        // We use higlass 'matrix' track instead of 'gosling-track' for rendering performance.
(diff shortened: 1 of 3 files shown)


src/core/gosling-to-higlass.ts
```diff
@@ -120,6 +127,20 @@ export function goslingToHiGlass(
             };
         }
 
+        // We use higlass 'matrix' track instead of 'gosling-track' for rendering performance.
+        // HiGlass is really well-optimized for matrix visualization, and rendering it in Gosling
+        // makes the zooming interaction slow.
+        // See https://github.com/gosling-lang/gosling.js/pull/612#discussion_r771623844
+        if (IsHiGlassMatrix(firstResolvedSpec)) {
+            // By changing the track type, HiGlass uses its native heatmap track
+            hgTrack.type = 'heatmap';
```
sehilyi, line 137: By replacing `'gosling-track'` with `'heatmap'`, HiGlass will take care of this track for rendering visualization and managing data, i.e., Gosling does not tabularize matrix data unnecessarily.

src/core/gosling-to-higlass.ts
```diff
@@ -120,6 +127,20 @@ export function goslingToHiGlass(
...
+            // TODO: use consistent color scheme w/ Gosling
+            hgTrack.options.colorRange = ['white', 'rgba(245,166,35,1.0)', 'rgba(208,2,27,1.0)', 'black'];
```
sehilyi: Need to use a consistent color scheme between Gosling and HiGlass.

sehilyi:
I will need to allow using horizontal rules w/ matrix in a separate PR, but our vertical rules in the example above seem to work well with HiGlass matrix.

sehilyi:
@manzt This is a follow-up PR of our previous PR (https://github.com/gosling-lang/gosling.js/pull/612). Could you review this?

src/core/gosling-to-higlass.ts
manzt: Great!

manzt (approved):
Looks great, thanks @sehilyi ! Thrilled the changes are minimal to handle this special case for matrix rendering.

sehilyi:
@manzt Thank you for reviewing this! I am glad that we came up with this approach (handling the special cases w/ higlass matrix) since this seems to be a much reasonable direction than what I initially planned in https://github.com/gosling-lang/gosling.js/pull/612!

src/core/gosling-to-higlass.ts
```diff
@@ -120,6 +128,23 @@ export function goslingToHiGlass(
             };
         }
 
+        // We use higlass 'matrix' track instead of 'gosling-track' for rendering performance.
```
sehilyi:
```suggestion
        // We use higlass 'heatmap' track instead of 'gosling-track' for rendering performance.
```

src/core/gosling-to-higlass.ts
```diff
@@ -120,6 +128,23 @@ export function goslingToHiGlass(
...
+        // HiGlass is really well-optimized for matrix visualization, and rendering it in Gosling
+        // makes the zooming interaction slow.
```
sehilyi:
```suggestion
        // instead makes the zooming interaction slow.
```

PR #624, opened 2022-01-24
feat(experimental): allow defining alternative specs for responsive visualization

sehilyi:
```js
{
  "arrangement": "horizontal",
  // Select an alternative option if its `selectivity` condition is passed
  "responsiveSpec": [
    {
      // use vertical arrangement if there is no enough space
      "spec": { "arrangement": "vertical" },
      "selectivity": [ 
        {
          "operation": "GT",
          "measure": "width",
          "threshold": 100
        }
      ]
    }
  ],
  ...
}
```

Screen Shot 2022-01-31 at 15 16 56

Screen Shot 2022-01-31 at 15 17 05

# To-Do
- [ ] Add tests
- [ ] Handle `linkingId`


src/core/gosling.schema.ts
```diff
@@ -1,7 +1,9 @@
 import { Chromosome } from './utils/chrom-size';
 
 /* ----------------------------- ROOT SPEC ----------------------------- */
-export type GoslingSpec = RootSpecWithSingleView | RootSpecWithMultipleViews;
+export type GoslingSpec =
+    | (RootSpecWithSingleView & ResponsiveSpecOfSingleView)
```
sehilyi, line 5: TODO: Define specific types for this since this union is applied several times in this file.

src/core/compile.ts
```diff
@@ -1,29 +1,47 @@
...
+
+    // Generate arrangement data
+    const trackInfosAndSize = getRelativeTrackInfo(specCopy, theme);
+    let { trackInfos } = trackInfosAndSize;
+    const { size } = trackInfosAndSize;
+
+    // Handle responsive specs, either remove them or replace original specs w/ them
+    const wFactor = curSize ? curSize?.width / size.width : 1;
+    const hFactor = curSize ? curSize?.height / size.height : 1;
+    const replaced = manageResponsiveSpecs(specCopy, wFactor, hFactor);
```
sehilyi, line 37: TODO: Better to add tests for this function.

yarn.lock
```diff
@@ -3035,6 +3035,11 @@ cross-spawn@^7.0.0, cross-spawn@^7.0.2:
     shebang-command "^2.0.0"
     which "^2.0.1"
 
+css-element-queries@^1.2.3:
```
sehilyi, line 3038: This library is also being used in HiGlass

PR #627, opened 2022-01-26
feat: support `yDomain` for 2D tracks

sehilyi:
## Usage of `x/yDomain`

```js
xDomain: { chromosome: "1" },
yDomain: { chromosome: "2" },

...

xDomain: { chromosome: "1", interval: [1000, 2000] },
yDomain: { chromosome: "2", interval: [1000, 2000] },

...

xDomain: { interval: [1000, 2000] },
yDomain: { interval: [1000, 2000] },

```

gosling-visualization (4)

### DIFF
--- a/src/core/utils/spec-preprocess.ts
+++ b/src/core/utils/spec-preprocess.ts
@@ -150,6 +150,7 @@ export function traverseToFixSpecDownstream(spec: GoslingSpec | SingleView, pare
         if (spec.static === undefined) spec.static = parentDef.static !== undefined ? parentDef.static : false;
         if (spec.zoomLimits === undefined) spec.zoomLimits = parentDef.zoomLimits;
         if (spec.xDomain === undefined) spec.xDomain = parentDef.xDomain;
+        if (spec.yDomain === undefined) spec.yDomain = parentDef.yDomain;
         if (spec.linkingId === undefined) spec.linkingId = parentDef.linkingId;
         if (spec.centerRadius === undefined) spec.centerRadius = parentDef.centerRadius;
         if (spec.spacing === undefined && !('tracks' in spec)) spec.spacing = parentDef.spacing;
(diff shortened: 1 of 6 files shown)


src/core/utils/spec-preprocess.ts
```diff
@@ -261,6 +262,19 @@ export function traverseToFixSpecDownstream(spec: GoslingSpec | SingleView, pare
             if (Is2DTrack(track)) {
                 // TODO: Add a test for this.
                 track.layout = 'linear';
+
+                /**
+                 * Add y-axis domain
+                 */
+                if ((IsSingleTrack(track) || IsOverlaidTrack(track)) && IsChannelDeep(track.y) && !track.y.domain) {
```
sehilyi, line 269: The `yDomain` value is overridden by children tracks here (i.e., `y: { ..., domain: {...} }`).

manzt (approved):
Looks good to me! Just out of curiosity. What is the behavior for a 2D track if only an `xDomain` is configured? 

Also -- random -- but is it possible to use the old "warm" colormap in the latest gosling.js?

sehilyi:
@manzt, thanks a lot for reviewing!

> What is the behavior for a 2D track if only an xDomain is configured?

We currently use `xDomain` for `yDomain` if `yDomain` is undefined.

> Also -- random -- but is it possible to use the old "warm" colormap in the latest gosling.js?

I assume you are referring to this one:

Screen Shot 2022-01-27 at 13 48 28

It is not possible at the moment. Gosling.js currently only supports a subset of [d3-scale-chromatic](https://github.com/d3/d3-scale-chromatic) which is listed in our code [here](https://github.com/gosling-lang/gosling.js/blob/master/src/core/gosling.schema.guards.ts#L48). I am unsure if there is an exact same color scheme in the `d3-scale-chromatic` though https://github.com/d3/d3-scale-chromatic#interpolateOranges or https://github.com/d3/d3-scale-chromatic#interpolateOrRd seem to be almost identical. We can easily support any color schemes that are listed in d3-scale-chromatic. We can also support fully custom color schemes. Let me know if you need certain color schemes.

manzt:
> We currently use xDomain for yDomain if yDomain is undefined.

makes sense to me!

> We can easily support any color schemes that are listed in `d3-scale-chromatic`. We can also support fully custom color schemes. Let me know if you need certain color schemes.

Great. Thanks for letting me know. I think `schemeYlOrBr` or one you have listed would be nice to have as something similar is default for HiGlass, perhaps identifiable with the name `"hot"` or similar.

sehilyi:
I agree. Let me create another PR to add the "hot" color scheme.

PR #629, opened 2022-01-27
feat: add hot and pink color schemes

sehilyi:
Added two new colors, i.e., `"hot"` and `"pink"`.

`interpolateYlOrBr` indeed seems to be the most similar color scheme compared with the HiGlass' default color. "pink" is added as well which is just my personal taste.

## "hot"
hot

## "pink"
pink


manzt (approved):
Great!

PR #645, opened 2022-02-22
feat: use reactive rendering for responsive specs

sehilyi:
Current Issue
- [x] when responsive height (e.g., `responsiveSize.height`) was used and then turned off, the track height does not revert back to the original height that is specified in the spec
  - Potential source of this: https://github.com/higlass/higlass/blob/2a3786e13c2415a52abc1227f75512f128e784a0/app/scripts/HiGlassComponent.js#L2199

https://user-images.githubusercontent.com/9922882/155395921-fa97e458-8073-48cb-be93-5be44e356410.mov

<details>
```
Previous comment by Alex:

I am trying to insert a track into a viewconf using the setViewConfig  API function. It works, but the height of the view is not adjusted (I am using the default size mode). If I open the viewconf editor and close it again, the height seems to be recomputed and I can see all the tracks. Is there a way to force Higlass to recompute the view height?

FYI: I got it to work with

hgc.api.setViewConfig(resp.viewconfig, true).then((v) => {
     const settings = {
          viewId: viewId,
          trackId: resp.viewconfig.views[1].tracks.top[0].uid
      }
      hgc.trackDimensionsModifiedHandlerBound(settings)
  });
This calls adjustLayoutToTrackSizes  internally. Probably not the most elegant solution, but it works.

```
</details>

### DIFF
--- a/src/core/higlass-component-wrapper.tsx
+++ b/src/core/higlass-component-wrapper.tsx
@@ -44,6 +44,7 @@ export const HiGlassComponentWrapper = forwardRef<HiGlassApi | undefined, HiGlas
         }, [props.id]);
 
         const viewConfig = props.viewConfig || {};
+        const pixelPreciseMarginPadding = !props.options.responsiveHeight;
         const higlassComponent = useMemo(
             () => (
                 <HiGlassComponent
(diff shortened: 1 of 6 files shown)


sehilyi:
If this approach of using reactive rendering works well, we can make it official, instead of `experimental.reactive`

src/core/higlass-component-wrapper.tsx
```diff
@@ -53,7 +54,7 @@ export const HiGlassComponentWrapper = forwardRef<HiGlassApi | undefined, HiGlas
                         // Since using this disallows responsive resizing of track heights in HiGlass,
                         // we need to use this only when users do not want to use responsive height.
                         // (See https://github.com/higlass/higlass/blob/2a3786e13c2415a52abc1227f75512f128e784a0/app/scripts/HiGlassComponent.js#L2199)
-                        pixelPreciseMarginPadding: !props.options.responsiveHeight,
+                        pixelPreciseMarginPadding,
```
sehilyi, line 57: I don't see any downside of turning this off.

PR #666, opened 2022-04-26
feat: replace jest with vitest

manzt:
[Vitest](https://vitest.dev/) is a unit-testing framework that is intended to be a drop-in replacement for Jest. The main benefit is that is reuses Vite's config, transformers, resolvers, and plugins. This means that the testing environment  more closely mimics the final build, and we no longer need to worry about making things work
in Jest _and_ our build.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,11 +1,9 @@
 module.exports = {
     parser: '@typescript-eslint/parser',
-    plugins: ['@typescript-eslint', 'prettier', 'jest'],
+    plugins: ['@typescript-eslint', 'prettier'],
     extends: [
         'eslint:recommended',
         'plugin:react/recommended',
-        'plugin:jest/recommended',
-        'plugin:jest/style',
         'plugin:@typescript-eslint/recommended',
         'prettier',
         'prettier/@typescript-eslint',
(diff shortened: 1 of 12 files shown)


manzt:
Not sure where the required workflows are configured for this repo, but we need to bump the 12.x requirement to 16.x. (since this PR removes the 12.x node workflow)

sehilyi (commented):
Great! I like how things are becoming more unified with Vite ecosystems.

One thing I found is that the coverage report on my machine after running `yarn coverage` is empty. Is this the same on your machine?

Screen Shot 2022-04-26 at 16 56 53

manzt:
hmm that is very odd. It seems that it works in the CI, and locally I have this:

image

manzt:
@sehilyi perhaps try deleting the coverage directory you have locally (`rm -r ./coverage`) and running `yarn coverage` again

sehilyi:
@manzt Was no luck with the folder deletion. I think a node.js version installed on my machine (v14.17) was the source of the issue. After upgrading it to v16.15.0, the coverage works.

sehilyi:
I think I was one minor patch behind the minimum requirement:
Screen Shot 2022-04-27 at 13 15 51

manzt:
> I think I was one minor patch behind the minimum requirement:

Ah ok, that makes sense. I'm sure it had something to do with ESM node compatibility.

manzt:
Do you know where the "Test and build" requirement for v12.x is configured? I'm guessing it's somewhere in the repo Settings, but I don't have the ability to change this given my permissions.

sehilyi:
I was able to remove the v12.x tests!

> FYI for myself since the menu is located quite deeply, [Settings] => [Branches] => [edit] on master => remove "v12.x" from a list of "Status checks that are required." (https://stackoverflow.com/a/68988320)

PR #671, opened 2022-05-01
feat(editor): support javascript editor

wangqianwen0418:
image

Supporting a javascript editor enables:
- more structured and easy-to-read code, especially for multi-view visualizations
- comments & annotations in the code example


editor/editor-panel.tsx
```diff
@@ -132,7 +138,7 @@ function EditorPanel(props: {
             ></ReactResizeDetector>
             <MonacoEditor
                 // Refer to https://github.com/react-monaco-editor/react-monaco-editor
-                language="json"
+                language={language}
```
sehilyi, line 148: Even more great that two languages are now handled in a single tsx file.

editor/editor.tsx
```diff
@@ -416,20 +440,37 @@ function Editor(props: any) {
...
+                    valid = gosling.validateGoslingSpec(editedGos);
+                    setLog(valid);
+                } catch (e) {
+                    const message = '✘ Cannnot parse the code.';
+                    console.warn(message);
+                    setLog({ message, state: 'error' });
+                }
+            } else if (language === 'javascript') {
+                try {
+                    const transplieCode = ts.transpile(jsCode);
```
sehilyi, line 456: Seems like a typo?
```suggestion
                    const transpiledCode = ts.transpile(jsCode);
```

editor/editor.tsx
```diff
@@ -416,20 +440,37 @@ function Editor(props: any) {
...
+                    editedGos = window.Function(`${transplieCode}\n return spec`)();
```
sehilyi: Seems like a typo (`transp*li*eCode`)?

```suggestion
                    const transpiledCode = ts.transpile(jsCode);
                    editedGos = window.Function(`${transpiledCode}\n return spec`)();
```

editor/editor.tsx
```diff
@@ -21,9 +21,14 @@ import './editor.css';
 import { ICONS, ICON_INFO } from './icon';
 import type { HiGlassSpec } from '@higlass.schema';
 import type { Datum } from '@gosling.schema';
+import * as ts from 'typescript';
```
sehilyi: We could narrow down the scope (unsure if this will affect the size of our package though).

```suggestion
import { transpile } from 'typescript';
```

editor/editor.tsx
```diff
@@ -416,20 +440,37 @@ function Editor(props: any) {
...
+                }
+            } else if (language === 'javascript') {
+                try {
+                    const transplieCode = ts.transpile(jsCode);
+                    editedGos = window.Function(`${transplieCode}\n return spec`)();
+                    valid = gosling.validateGoslingSpec(editedGos);
+                    setLog(valid);
+                } catch (e) {
+                    const message = '✘ Cannnot parse the code.';
+                    console.warn(message);
```
sehilyi: It might be good to show a more detailed message for the case where a variable named `spec` is undefined at all. Perhaps, this can be explicitly explained somewhere (e.g., show a tooltip when hovering a mouse on a "JavaScript" tab).

editor/editor.tsx
```diff
@@ -854,15 +895,54 @@ function Editor(props: any) {
...
+                                            JSON
+                                        </button>
+                                        <button
+                                            className={`tablinks ${language == 'javascript' && 'active'}`}
+                                            onClick={() => {
+                                                changeLanguage('javascript');
+                                                setLog({ message: '', state: 'success' });
+                                            }}
+                                        >
+                                            Javascript
```
sehilyi:
```suggestion
                                            JavaScript
```

editor/editor.tsx
```diff
@@ -21,9 +21,14 @@ import './editor.css';
 import { ICONS, ICON_INFO } from './icon';
 import type { HiGlassSpec } from '@higlass.schema';
 import type { Datum } from '@gosling.schema';
+import * as ts from 'typescript';
 // @ts-ignore
 import { Themes } from 'gosling-theme';
 
+function json2js(jsonCode: string) {
+    return `var spec=${jsonCode}`;
```
sehilyi: Just to use the consistent formatting with your JS codes.
```suggestion
    return `var spec = ${jsonCode}`;
```

editor/example/spec/corces.ts
```diff
@@ -0,0 +1,275 @@
...
+        color: { value: 'none' },
+        stroke: { value: d.color },
+        opacity: { value: 0.1 },
+        overlayOnPreviousTrack: i != 0,
+        width,
+        height: 60
+    };
+});
+
+const spec = {
```
sehilyi: Could we explicitly define a type of individual variables used in JS-based examples? (e.g., GoslingSpec, DataDeep, OverlaidTrack)

Since our examples are becoming large, defining types will make it easier to maintain the examples (e.g., when we rename property names, IDE will automatically rename corresponding examples, etc).

```suggestion
const spec : GoslingSpec = {
```

editor/example/index.ts
```diff
@@ -105,25 +128,29 @@ export const examples: {
...
-        spec: EX_SPEC_BAND,
+        spec: JSON_SPEC_BAND as GoslingSpec,
+        specJs: jsStringParser(JS_SPEC_BAND),
         image: THUMBNAILS.BAND
     },
     RULE: {
         group: 'Visual Encoding',
         name: 'Rule Mark',
-        spec: EX_SPEC_RULE,
+        spec: JSON_SPEC_RULE as GoslingSpec,
```
sehilyi: We can remove all `as GoslingSpec` when we define types of a variable `spec` in individual example files.

```suggestion
        spec: JSON_SPEC_RULE,
```

sehilyi:
@wangqianwen0418 Thanks a lot for this update! This is putting together really nicely. 

I left some minor comments and thoughts, but mainly, I think it would be great if variables defined in JS-based examples have explicit types (e.g., `let spec: GoslingSpec`, `let data: DataDeep`) instead of being `any` JSON object. If specified, IDE will be able to understand the syntax, which is largely advantageous for us to maintain the large set of examples (e.g., if a property name in a `gosling.schema.ts` is changed in an IDE, all example files will be updated automatically).

wangqianwen0418:
@sehilyi, thanks a lot for the comments!

> I think it would be great if variables defined in JS-based examples have explicit types (e.g., let spec: GoslingSpec, let data: DataDeep) instead of being any JSON object. If specified, IDE will be able to understand the syntax, which is largely advantageous for us to maintain the large set of examples (e.g., if a property name in a gosling.schema.ts is changed in an IDE, all example files will be updated automatically).

Yes, this is what I really want to do but have not found a good solution yet. The `window.Function()` does not support `import`, which makes it hard to properly import type definitions. 
I will try some workarounds and push them later.

sehilyi:
All sounds good. (Just FYI, I resolved a conflict introduced by my latest merge.)

sehilyi:
@wangqianwen0418 If it turns out to be technically challenging to specify types, we could address the issue later in a separate PR.

wangqianwen0418:
> If it turns out to be technically challenging to specify types, we could address the issue later in a separate PR.

@sehilyi, no worry, I was just overwhelmed by the VIS review ddl last week and did not have enough bandwidth.

The type import has been supported in the lastest commits 


image

wangqianwen0418:
By the way, I am hesitating should we support both a typescript editor and a javascript editor.
On the one hand, a js editor can help those who are not familiar with the ts grammar (by showing the code after transpile). On the other hand, I feel it is a little redundant.

I am inclined to only include one typescript editor but would like to know your opinions

wangqianwen0418:
@sehilyi, also, do you think we can merge this PR as it is now? 
The example updating will take a longer time and is not very urgent.
I will gradually update them in another PR later.

sehilyi (commented):
Hi @wangqianwen0418,

> do you think we can merge this PR as it is now?
The example updating will take a longer time and is not very urgent.
I will gradually update them in another PR later

I thought the same. We should merge this sooner than later. Since all examples still work, we can update others later. Also merging this earlier would be good to avoid any potential conflict issues with the editor.

> By the way, I am hesitating should we support both a typescript editor and a javascript editor.
On the one hand, a js editor can help those who are not familiar with the ts grammar (by showing the code after transpile). On the other hand, I feel it is a little redundant.
I am inclined to only include one typescript editor but would like to know your opinions

I agree that people will be much more familiar with JS. (People will need to know what exact Types they need to use in TS). I think it would be best if we can enable people to use JS, but the spec templates in our code are defined in TS. However, unsure if this can be done fairly easily.

It looks like when I edit the JSON spec and then select the JS tab, the visualization is not updated based on the JS code. The visualization is updated after I either edit the JS code or click on the Run button. Do you think it would be more straightforward to update the visualization upon clicking on the tab?

wangqianwen0418:
> It looks like when I edit the JSON spec and then select the JS tab, the visualization is not updated based on the JS code. 

Good point, I have fixed it in the latest commit.

wangqianwen0418:
> People will need to know what exact Types they need to use in TS. 

Not really. The current js editor works totally fine for both JS and TS. The editor provides type checking when the code is TS but also works for code with no type definitions.

sehilyi:
> Not really. The current js editor works totally fine for both JS and TS. The editor provides type checking when the code is TS but also works for code with no type definitions.

I wonder if showing types by default confuse users. I was wondering if it is better to hide TS types in the editor (`const data = {}`, instead of `const data: MultivecData = {}`) but specify TS types in the examples in our code. In that way, users do not need to care about types at all but we manage specs more effectively in our repo.

sehilyi:
> Not really. The current js editor works totally fine for both JS and TS. The editor provides type checking when the code is TS but also works for code with no type definitions.

I see. My point was more like -- would showing types confuse users? I assume many people will not know the TS syntax and showing them gives an impression that they need to use it in the editor.

sehilyi:
@wangqianwen0418 I see many examples already include type definitions (which is great for code management!). Unless it will be fairly easy to show JS specs to users but maintain specs with type definitions, how about just merging this as-is? Since having TS codes also has advantages, I don't think we need to revert back and remove types.

wangqianwen0418:
@sehilyi, is this PR ready to merge, or do you still need to do some reviews?
Let me know if any other changes are needed

Getting JS-spec from TS spec is technically possible, but I need to do some experiments first. 
I will start two PRs laters to 1) update more examples and 2) support both TS and JS when the example is written in TS

sehilyi (approved):
Looks good to merge to me! Feel free to merge when ready.

sehilyi:
@wangqianwen0418 Oh, it looks like the merged code is not being deployed by an odd JS error... Can you take a look?

Screen Shot 2022-05-10 at 12 18 22

Screen Shot 2022-05-10 at 12 17 11

wangqianwen0418:
Sure, looks like an out-of-memory issue.
I will work on it soon

PR #674, opened 2022-05-03
fix: fix crash on matrix example

### LINKED ISSUE #673: Comparative matrices example crashes the editor
Ran into this during BMI 706 today. 

- Chrome: Version 101.0.4951.54 (Official Build) (x86_64)

Steps to reproduce: Navigate to https://gosling.js.org > Coordinated Multiple Views: Comparative Matrices (Micro-C vs. Hi-C)

image

sehilyi:
Fix #673


sehilyi:
Need to prevent such bugs in the future. Perhaps, image-diff tests.

PR #678, opened 2022-05-10
[draft] feat: show more ticks and baselines of circular axes

sehilyi:
Screen Shot 2022-05-10 at 12 20 49


sehilyi:
This rather would introduce large changes to the source codes. I can revisit this later once more important things are addressed.

PR #686, opened 2022-05-11
fix: remove legend layer in `remove()`; do not render twice on Editor when example ID is specified in URL

### LINKED ISSUE #685: color legend and controllers show up in the wrong place
In https://gosling.js.org/?example=ALIGNMENT (Gosling.js v0.9.18) the range slider for the color legend shows of in the wrong place and doesn't seem to work (see the two gray rectangles in the conservation and gap views):

image

sehilyi:
Fix #685

### DIFF
--- a/editor/editor.tsx
+++ b/editor/editor.tsx
@@ -201,9 +201,9 @@ function Editor(props: any) {
     const urlGist = urlParams?.gist ?? null;
     const urlExampleId = (urlParams?.example ?? '') as string;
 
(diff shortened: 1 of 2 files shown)


sehilyi:
The editor rendering gosling specs twice too quickly seems to be the source of the issue. This seems to break the rendering cycle of each HiGlass component, i.e., rendering the SVG color brush elements of the first spec after rendering the gosling canvas of the second spec.

sehilyi:
Removing the entire legend layer upon the removal of a track addresses the core issue:

```js
remove() {
   // ...
   this.gLegend.remove();
   this.gLegend = null;
}

PR #692, opened 2022-05-16
chore: clean up editor imports/deps

manzt:
- chore: add `vite/client` types to `tsconfig.json`
	- Adds all builtin vite-isms for TypeScript. Removes our manual `src/types/index.d.ts` definitions for Workers / static assest imports.
- chore: add router types
	- Replaces `any` for Editor props
- chore: remove `qs`, use URLSearchParams
	- Unecessary dependency. This behavior is built into browsers natively. (https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams)
- chore: fix use of Refs, export `GoslingRef` type
        - Removes use of `any` in Gosling editor. Uses appropriate typescript semantics around mutable and immutable refs.


manzt:
Was just poking around with the editor to expose some of this Twitter behavior, cleaned up a few things. A general observation is that there is liberal use of `any`, which makes TS utility extremely limited here.

tsconfig.json
```diff
@@ -28,7 +28,7 @@
       "zlib": ["./src/alias/zlib.ts"]
     },
     "baseUrl": ".",
-    "types": ["vitest/globals"]
+    "types": ["vitest/globals", "vite/client"]
```
sehilyi, line 31: Nice!

sehilyi (approved):
LGTM, thanks for the update!

Issue #582, opened 2021-11-15
Support custom chromsizes


zhangzhen:
Adding a custom-defined genome is easy for higlass, while doing this is hard for gosling. Could you please guide me in modifying the source code for gosling to achieve this? After finishing the modification, zooming, panning and zoom_to are expected to function correctly.
Besides, sequence abstraction can be easily implemented by add a custom-defined genome to gosling.

sehilyi:
Hi Zhang! I assume you want to use custom chromosome names and their sizes.

We have a collection of assemblies in [/src/core/utils/assembly.ts](https://github.com/gosling-lang/gosling.js/blob/master/src/core/utils/assembly.ts), so we will need to extend this.

This information is used in several places, for example, in the [axis track](https://github.com/gosling-lang/gosling.js/blob/3cd57a6a499de6a79f5f4bc71947fdb8a44fbdd9/src/gosling-genomic-axis/axis-track.ts#L110).

Grammar-wise, I think a user would need to specify a URL of a TSV file that contains chromosome names and sizes (just like `"chromInfoPath` in HiGlass), e.g., 

```js
// https://github.com/gosling-lang/gosling.js/blob/3cd57a6a499de6a79f5f4bc71947fdb8a44fbdd9/src/core/gosling.schema.ts#L84
assembly: { url: "//s3.amazonaws.com/pkerp/data/hg19/chromSizes.tsv" }
```

I wonder if we can allow users to specify the chr size array directly from the spec (e.g., `{ name: string, length: number}[]`), but this might not be feasible at the moment since HiGlass uses the URL of a TSV file and does not support the inline JSON object in its view config.

manzt:
Came across this recently. There seem to generally be inconsistencies between data-sources and genome assemblies. For example, vcf is hard-coded as `assembly = 'hg38'`, but also hard-coded with chromsizes url for `hg19`?

https://github.com/gosling-lang/gosling.js/blob/a12d454fa9272a03ce76a6cb8cc513ed9ed8d694/src/data-fetcher/vcf/gosling-vcf-data.ts#L30-L42

> I wonder if we can allow users to specify the chr size array directly from the spec (e.g., `{ name: string, length: number}[]`), but this might not be feasible at the moment since HiGlass uses the URL of a TSV file and does not support the inline JSON object in its view config.

You can always create data URLs on the client...

```javascript
let chromsizes = [
  ["foo", 1000],
  ["bar", 300],
  ["baz", 240]
];
[...]

PR #693, opened 2022-05-16
feat: genomic range selection in linear tracks

sehilyi:
# Updates

- Support genomic range selection using mouse + alt key
   - Individual marks (or grouped marks) that are entirely inside the given genomic range are selected and highlighted
   - Clicking on the outside of the brush removes the brush
   - Each track can have only one brush
   - Users can resize the brush by dragging the edge of the brush
   - Users can reposition the brush by dragging the body of the brush
   - Brush positions, as well as highlighted elements, are maintained upon zooming and panning
- Support `rangeselect` API to subscribe range brush change
```ts
api.subscribe('rangeselect', (type: string, eventData: RangeSelectEventData) => {
   console.log(
      type, // 'rangeselect'
      eventData.id. // 'track-3'
      eventData.genomicRange, // ['chr1:124,000', 'chr1:450,000'] or null if brush removed
      eventData.data // [{ chr: '1', pos: 124000, value: 12 }, ...]
   );
});
```
- Support customizing the selection and hovering effects, as well as the brush.
```ts
// ... track-level properties
// Mouse events
experimental?: {
    /* Group marks using keys in a field. This affects how a set of marks are highlighted/selected by interaction. */
    groupMarksByField?: string;

    hovering?: {
        enableMultiHovering?: boolean;
        showHoveringOnTheBack?: boolean;
        color?: string;
        stroke?: string;
        strokeWidth?: number;
        strokeOpacity?: number;
        opacity?: number;
    };
    selection?: {
        showOnTheBack?: boolean;
        color?: string;
        stroke?: string;
        strokeWidth?: number;
        strokeOpacity?: number;
        opacity?: number;
    };
    brush?: {
        color?: string;
        stroke?: string;
        strokeWidth?: number;
        strokeOpacity?: number;
        opacity?: number;
    };
};
```

## To-Do

- [x] Support APIs to access selected elements (e.g., `{ genomicRange, values }`)
- [x] Expose customization properties
- [x] Fix an issue that `rangeselect` API is called redundantly.
- [x] Support selecting a group of marks (e.g., a gene annotation)
- [x] Select marks that are entirely inside a range
- [x] Selection effects seem to be redundantly displayed when selecting grouped marks.
- [x] Better handle when selection is null
- [x] Remove brush when changed to circular layout

# Screenshots

Screen Shot 2022-05-18 at 15 56 57

https://user-images.githubusercontent.com/9922882/169143907-e1e814c9-eb93-4865-9bca-c2b05ac726fb.mov

Screen Shot 2022-05-24 at 12 13 00

Screen Shot 2022-05-24 at 12 21 25

# Open Issues
- Selection across separate tracks that are overlaid because they use different data (e.g., mark displaced lollipop plot example)
- Allow API callers to know the "type" of selection events (e.g., dragging, adjusting edge, creating, removing, zooming)
- Alternative triggering methods? (e.g., a selection-mode button like in Peax?)
- Selection along a single `row`?
- Proper visual cues when selection is triggered?
- Let the API subscriber know when the selected data changed although the range did not change (e.g., new tilesets by zooming in)
- Unify with the current `brush` mark.
- Merging sets of selections (e.g., union or intersection)
- Limit the size of the JSON array (i.e., `data`) that is sent to API subscribers?
- Should we add an option for users to only get the genomic range (w/o selected data)?

### DIFF
--- a/editor/editor.tsx
+++ b/editor/editor.tsx
@@ -26,7 +26,6 @@ import type { HiGlassSpec } from '@higlass.schema';
 import type { Datum } from '@gosling.schema';
 
 import './editor.css';
-import { RangeMouseEventData } from 'src/core/api';
 
 function json2js(jsonCode: string) {
     return `var spec = ${jsonCode} \nexport { spec }; \n`;
@@ -288,21 +287,21 @@ function Editor(props: RouteComponentProps) {
     // publish event listeners to Gosling.js
     useEffect(() => {
         if (gosRef.current) {
-            // gosRef.current.api.subscribe('rawdata', (type: string, data: RawDataEventData) => {
+            // gosRef.current.api.subscribe('rawdata', (type, data) => {
             // console.log('rawdata', data);
             // gosRef.current.api.zoomTo('bam-1', `chr${data.data.chr1}:${data.data.start1}-${data.data.end1}`, 2000);
             // gosRef.current.api.zoomTo('bam-2', `chr${data.data.chr2}:${data.data.start2}-${data.data.end2}`, 2000);
             // console.log('click', data.data);
             // TODO: show messages on the right-bottom of the editor
-            // gosRef.current.api.subscribe('mouseover', (type: string, eventData: CommonEventData) => {
+            // gosRef.current.api.subscribe('mouseover', (type, eventData) => {
             //     setMouseEventInfo({ type: 'mouseover', data: eventData.data, position: eventData.genomicPosition });
             // });
-            // gosRef.current.api.subscribe('click', (type: string, eventData: CommonEventData) => {
+            // gosRef.current.api.subscribe('click', (type, eventData) => {
             //     setMouseEventInfo({ type: 'click', data: eventData.data, position: eventData.genomicPosition });
             // });
 
             // Range Select API
-            gosRef.current.api.subscribe('rangeselect', (type: string, eventData: RangeMouseEventData) => {
+            gosRef.current.api.subscribe('rangeselect', (type, eventData) => {
                 console.warn(type, eventData.id, eventData.genomicRange, eventData.data);
             });
         }
--- a/editor/editor.tsx
+++ b/editor/editor.tsx
@@ -326,16 +326,15 @@ function Editor(props: RouteComponentProps) {
             // gosRef.current.api.subscribe('click', (type, eventData) => {
             //     setMouseEventInfo({ type: 'click', data: eventData.data, position: eventData.genomicPosition });
             // });
-
             // Range Select API
-            gosRef.current.api.subscribe('rangeselect', (type, eventData) => {
-                console.warn(type, eventData.id, eventData.genomicRange, eventData.data);
-            });
+            // gosRef.current.api.subscribe('rangeselect', (type, eventData) => {
+            //     console.warn(type, eventData.id, eventData.genomicRange, eventData.data);
+            // });
         }
         return () => {
             // gosRef.current.api.unsubscribe('mouseover');
             // gosRef.current.api.unsubscribe('click');
-            gosRef.current?.api.unsubscribe('rangeselect');
+            // gosRef.current?.api.unsubscribe('rangeselect');
         };
     }, [gosRef.current]);
 
--- a/editor/editor.tsx
+++ b/editor/editor.tsx
@@ -26,7 +26,7 @@ import type { HiGlassSpec } from '@higlass.schema';
 import type { Datum } from '@gosling.schema';
 
 import './editor.css';
-import { RangeSelectEventData } from 'src/core/api';
+import { RangeMouseEventData } from 'src/core/api';
 
 function json2js(jsonCode: string) {
     return `var spec = ${jsonCode} \nexport { spec }; \n`;
@@ -302,14 +302,14 @@ function Editor(props: RouteComponentProps) {
             // });
 
             // Range Select API
-            gosRef.current.api.subscribe('rangeselect', (type: string, eventData: RangeSelectEventData) => {
-                console.warn(type, eventData.genomicRange, eventData.data);
+            gosRef.current.api.subscribe('rangeselect', (type: string, eventData: RangeMouseEventData) => {
+                console.warn(type, eventData.id, eventData.genomicRange, eventData.data);
             });
         }
         return () => {
             // gosRef.current.api.unsubscribe('mouseover');
             // gosRef.current.api.unsubscribe('click');
-            // gosRef.current.api.unsubscribe('rangeselect');
+            gosRef.current?.api.unsubscribe('rangeselect');
         };
     }, [gosRef.current]);
 
--- a/editor/editor.tsx
+++ b/editor/editor.tsx
@@ -26,6 +26,7 @@ import type { HiGlassSpec } from '@higlass.schema';
 import type { Datum } from '@gosling.schema';
 
 import './editor.css';
+import { RangeSelectEventData } from 'src/core/api';
 
 function json2js(jsonCode: string) {
     return `var spec = ${jsonCode} \nexport { spec }; \n`;
@@ -299,10 +300,16 @@ function Editor(props: RouteComponentProps) {
             // gosRef.current.api.subscribe('click', (type: string, eventData: CommonEventData) => {
             //     setMouseEventInfo({ type: 'click', data: eventData.data, position: eventData.genomicPosition });
             // });
+
+            // Range Select API
+            gosRef.current.api.subscribe('rangeselect', (type: string, eventData: RangeSelectEventData) => {
+                console.warn(type, eventData.genomicRange, eventData.data);
+            });
         }
         return () => {
             // gosRef.current.api.unsubscribe('mouseover');
             // gosRef.current.api.unsubscribe('click');
+            // gosRef.current.api.unsubscribe('rangeselect');
         };
     }, [gosRef.current]);
 
--- a/src/core/api.ts
+++ b/src/core/api.ts
@@ -11,6 +11,11 @@ export type CommonEventData = {
     data: Datum[];
 };
 
+export type RangeSelectEventData = {
+    genomicRange: [string, string] | null; // null if range deselected
+    data: Datum[];
+};
+
 export type RawDataEventData = {
     id: string;
     data: Datum[];
@@ -28,8 +33,9 @@ type PubSubEvent<EventName extends string, Payload> = {
 };
 
 // New `PubSubEvent`s should be added to the `EventMap`...
-type EventMap = PubSubEvent<'mouseover' | 'click', CommonEventData> & PubSubEvent<'rawdata', RawDataEventData>;
-// & PubSubEvent<'my-event', { hello: 'world' }> & PubSubEvent<'foo', "bar">;
+type EventMap = PubSubEvent<'mouseover' | 'click', CommonEventData> &
+    PubSubEvent<'rangeselect', RangeSelectEventData> &
+    PubSubEvent<'rawdata', RawDataEventData>;
 
 /**
  * Information of suggested genes.
@@ -107,10 +113,9 @@ export function createApi(
         subscribe: (type, callback) => {
             switch (type) {
                 case 'mouseover':
-                    return PubSub.subscribe(type, callback);
                 case 'click':
-                    return PubSub.subscribe(type, callback);
                 case 'rawdata':
+                case 'rangeselect':
                     return PubSub.subscribe(type, callback);
                 default: {
                     console.error(`Event type not recognized, got ${JSON.stringify(type)}.`);
--- a/src/core/api.ts
+++ b/src/core/api.ts
@@ -6,20 +6,22 @@ import { GET_CHROM_SIZES } from './utils/assembly';
 import { CompleteThemeDeep } from './utils/theme';
 import { traverseViewsInViewConfig } from './utils/view-config';
 
-export type CommonEventData = {
-    genomicPosition: string;
-    data: Datum[];
-};
-
-export type RangeSelectEventData = {
-    genomicRange: [string, string] | null; // null if range deselected
-    data: Datum[];
-};
-
-export type RawDataEventData = {
+export interface CommonEventData {
+    /** Source visualization ID, i.e., `track.id` */
     id: string;
+    /** Values in a JSON array that represent data after data transformation */
     data: Datum[];
-};
+}
+
+export interface PointMouseEventData extends CommonEventData {
+    /* A genomic coordinate, e.g., `chr1:100,000`. */
+    genomicPosition: string;
+}
+
+export interface RangeMouseEventData extends CommonEventData {
+    /* Start and end genomic coordinates, e.g., `chr1:100,000`. NULL if a range is deselected. */
+    genomicRange: [string, string] | null;
+}
 
 // Utility type for building strongly typed PubSub API.
 //
@@ -33,9 +35,9 @@ type PubSubEvent<EventName extends string, Payload> = {
 };
 
 // New `PubSubEvent`s should be added to the `EventMap`...
-type EventMap = PubSubEvent<'mouseover' | 'click', CommonEventData> &
-    PubSubEvent<'rangeselect', RangeSelectEventData> &
-    PubSubEvent<'rawdata', RawDataEventData>;
+type EventMap = PubSubEvent<'mouseover' | 'click', PointMouseEventData> &
+    PubSubEvent<'rangeselect', RangeMouseEventData> &
+    PubSubEvent<'rawdata', CommonEventData>;
 
 /**
  * Information of suggested genes.
--- a/src/core/gosling.schema.ts
+++ b/src/core/gosling.schema.ts
@@ -257,6 +257,7 @@ interface SingleTrackBase extends CommonTrackDef {
             opacity?: number;
         };
         selection?: {
+            showOnTheBack?: boolean;
             color?: string;
             stroke?: string;
             strokeWidth?: number;
--- a/src/core/gosling.schema.ts
+++ b/src/core/gosling.schema.ts
@@ -232,6 +232,14 @@ export type Mark =
 /* ----------------------------- TRACK ----------------------------- */
 export type SingleTrack = SingleTrackBase & Encoding;
 
+export interface BrushAndMarkHighlightingStyle {
+    color: string;
+    stroke: string;
+    strokeWidth: number;
+    strokeOpacity: number;
+    opacity: number;
+}
+
 interface SingleTrackBase extends CommonTrackDef {
     // Data
     data: DataDeep;
@@ -250,27 +258,9 @@ interface SingleTrackBase extends CommonTrackDef {
         hovering?: {
             enableMultiHovering?: boolean;
             showHoveringOnTheBack?: boolean;
-            color?: string;
-            stroke?: string;
-            strokeWidth?: number;
-            strokeOpacity?: number;
-            opacity?: number;
-        };
-        selection?: {
-            showOnTheBack?: boolean;
-            color?: string;
-            stroke?: string;
-            strokeWidth?: number;
-            strokeOpacity?: number;
-            opacity?: number;
-        };
-        brush?: {
-            color?: string;
-            stroke?: string;
-            strokeWidth?: number;
-            strokeOpacity?: number;
-            opacity?: number;
-        };
+        } & Partial<BrushAndMarkHighlightingStyle>;
+        selection?: { showOnTheBack?: boolean } & Partial<BrushAndMarkHighlightingStyle>;
+        brush?: Partial<BrushAndMarkHighlightingStyle>;
     };
 
     // Mark
--- a/src/core/gosling.schema.ts
+++ b/src/core/gosling.schema.ts
@@ -255,6 +255,20 @@ interface SingleTrackBase extends CommonTrackDef {
             strokeOpacity?: number;
             opacity?: number;
         };
+        selection?: {
+            color?: string;
+            stroke?: string;
+            strokeWidth?: number;
+            strokeOpacity?: number;
+            opacity?: number;
+        };
+        brush?: {
+            color?: string;
+            stroke?: string;
+            strokeWidth?: number;
+            strokeOpacity?: number;
+            opacity?: number;
+        };
     };
 
     // Mark
--- a/src/core/gosling.schema.ts
+++ b/src/core/gosling.schema.ts
@@ -244,10 +244,11 @@ interface SingleTrackBase extends CommonTrackDef {
 
     // Mouse events
     experimental?: {
+        /* Group marks using keys in a field. This affects how a set of marks are highlighted/selected by interaction. */
+        groupMarksByField?: string;
+
         hovering?: {
             enableMultiHovering?: boolean;
-            enableGroupHovering?: boolean;
-            searchGroupByField?: string;
             showHoveringOnTheBack?: boolean;
             color?: string;
             stroke?: string;
--- a/src/core/mark/bar.ts
+++ b/src/core/mark/bar.ts
@@ -140,7 +140,7 @@ export function drawBar(trackInfo: any, tile: any, model: GoslingTrackModel) {
 
                 if (typeof ye !== 'undefined') {
                     // make sure `ye` is a larger range value
-                    [y, ye] = [y, ye].sort();
+                    [y, ye] = [y, ye].sort((a, b) => a - b);
                 }
 
                 const barWidth = model.encodedPIXIProperty('width', d, { tileUnitWidth });
--- a/src/core/mark/bar.ts
+++ b/src/core/mark/bar.ts
@@ -138,9 +138,9 @@ export function drawBar(trackInfo: any, tile: any, model: GoslingTrackModel) {
                 let y = model.encodedPIXIProperty('y', d);
                 let ye = model.encodedPIXIProperty('ye', d);
 
-                if (typeof ye !== 'undefined') {
-                    // make sure `ye` is a larger range value
-                    [y, ye] = [y, ye].sort((a, b) => a - b);
+                if (typeof ye !== 'undefined' && y > ye) {
+                    // ensure `ye` is larger
+                    [y, ye] = [ye, y];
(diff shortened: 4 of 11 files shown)


editor/editor.tsx
```diff
@@ -299,10 +300,16 @@ function Editor(props: RouteComponentProps) {
...
+
+            // Range Select API
+            gosRef.current.api.subscribe('rangeselect', (type: string, eventData: RangeMouseEventData) => {
+                console.warn(type, eventData.id, eventData.genomicRange, eventData.data);
+            });
         }
         return () => {
             // gosRef.current.api.unsubscribe('mouseover');
             // gosRef.current.api.unsubscribe('click');
+            gosRef.current?.api.unsubscribe('rangeselect');
```
sehilyi, line 312: I will comment this out before merge.

sehilyi: I will comment out these lines before merging

editor/editor.tsx
```diff
@@ -26,6 +26,7 @@ import type { HiGlassSpec } from '@higlass.schema';
 import type { Datum } from '@gosling.schema';
 
 import './editor.css';
+import { RangeMouseEventData } from 'src/core/api';
```
manzt: Generally we should try to avoid importing from `src/*` from the editor and instead only import what is exposed from `gosling.js` the package. Imports like this will not be available to end users and we can run into the situation where we are relying on behavior that others will not have.

editor/editor.tsx
```diff
@@ -299,10 +300,16 @@ function Editor(props: RouteComponentProps) {
             // gosRef.current.api.subscribe('click', (type: string, eventData: CommonEventData) => {
             //     setMouseEventInfo({ type: 'click', data: eventData.data, position: eventData.genomicPosition });
             // });
+
+            // Range Select API
+            gosRef.current.api.subscribe('rangeselect', (type: string, eventData: RangeMouseEventData) => {
```
manzt: Shouldn't need to add a type annotation for `type` or `eventData`. The types for this function should be inferred from the `"rangeselect"` event name.

src/core/api.ts
```diff
@@ -6,15 +6,22 @@ import { GET_CHROM_SIZES } from './utils/assembly';
...
+    /** Source visualization ID, i.e., `track.id` */
     id: string;
+    /** Values in a JSON array that represent data after data transformation */
     data: Datum[];
-};
+}
+
+export interface PointMouseEventData extends CommonEventData {
+    /* A genomic coordinate, e.g., `chr1:100,000`. */
+    genomicPosition: string;
```
manzt, line 18: Any reason this isn't `[name: string, position: number]`? It's somewhat trival to convert to string representation by end user but not the other way around. Also this is assuming a particular formatting for the number.

```javascript
let [chrom, len] = genomicPosition
let formatter = new Intl.NumberForamt('en')
console.log(`${chrom}:${formatter.format(len)}`)
```

vs

```javascript
let [chrom, lenStr] = genomicPosition.split(':');
let len = parseInt(lenStr.replace(/,/g, ''));
```

src/core/api.ts
```diff
@@ -6,15 +6,22 @@ import { GET_CHROM_SIZES } from './utils/assembly';
...
+}
+
+export interface RangeMouseEventData extends CommonEventData {
+    /* Start and end genomic coordinates, e.g., `chr1:100,000`. NULL if a range is deselected. */
+    genomicRange: [string, string] | null;
```
manzt, line 23: Similar thought here, `string`s are probably useful for a human reading but not for a consumer of the API. Consumers would have to know how strings are formatted in-order to parse into something more useable.

However, I also don't think a pair of locations is enough information either for range. E.g.,

```typescript
type GenomicLocation = [chrom: string, position: number]; 
type GenomicRange = [GenomicLocation, GenomicLocation];

interface RangeMouseEventData extends CommonEventData {
  genomicRange: GenomicRange | null;
}
```

single a `GenomicRange` could span multiple chromosomes.

For example, returning `["chr1", 100000]` and `["chr6", 10]` isn't very helpful since an end-users will need to look up individual chromosome lengths and determine start/ends of each chromosome within this interval _in Gosling_.

An genomic assembly only defines sequence composition _within_ chromosome, and Gosling decides that chromosomes be ordered in a particular way end-to-end. Let's say we have a different chromosome ordering in Gosling (e.g. (`'chr1', 'chr6', 'chr2', chr3', ...`)), then the  genomic range `["chr1", 100000]` to `["chr6", 10]` means something entirely different. Of course chromosomes with numbered names have a natural order, but that's totally arbitrary.

I think we could improve this API by returning data for genomic range within each _chromosome_ selected:

```typescript
type GenomicInterval = [chrom: name, left: number, right: number];

interface RangeMouseEventData {
    id: string;
    intervals: { interval: GenomicInterval, data: Datum[] }[],
}
```

so an interval spanning chr1, chr2, chr3 would return something like:

```typescript
// [
//   { interval: ['chr1', 100000, 248956422], data: [/* ... */] },
//   { interval: ['chr2', 0, 242193529], data: [/* ... */] },
//   { interval: ['chr3', 0, 20000], data: [/* ... */] },
// ]
```

src/core/gosling.schema.ts
```diff
@@ -244,17 +244,33 @@ interface SingleTrackBase extends CommonTrackDef {
 
     // Mouse events
     experimental?: {
```
manzt, line 254: Is there any way to just turn this range selection / hovering off? or only enable it iff used? Seems kind of wasteful by default if there are no event listeners or anything who is using that data.

src/core/mark/bar.ts
```diff
@@ -140,7 +140,7 @@ export function drawBar(trackInfo: any, tile: any, model: GoslingTrackModel) {
 
                 if (typeof ye !== 'undefined') {
                     // make sure `ye` is a larger range value
-                    [y, ye] = [y, ye].sort();
+                    [y, ye] = [y, ye].sort((a, b) => a - b);
```
manzt: why not 
```typescript
if (y > ye) {
  // ensure `ye` is larger
  [y, ye] = [ye, y] 
}
```

src/gosling-brush/linear-brush-model.ts
```diff
@@ -0,0 +1,238 @@
+import * as d3Selection from 'd3-selection';
+import { drag as d3Drag } from 'd3-drag';
```
manzt: To make it clear that these imports are only used for types annotations (and the implementation is provided by HiGlass), I'd recommend using `import type` syntax and capitalizing the imports:

```typescript
import type * as D3Selection from 'd3-selection';
import type * as D3Drag from 'd3-drag';
```

The main issue I see with using's HiGlass's internals (not mentioned in #702) is that we need to make sure our version of our d3 types is identical to HiGlass. This could be challenging to ensure, but we are fortunate with how stable the d3 apis are now.

src/gosling-brush/linear-brush-model.ts
```diff
@@ -0,0 +1,238 @@
...
+    type: 'start';
+    cursor: 'ew-resize';
+}
+
+export interface OneDimBrushEndEdgeData extends OneDimBrushDataCommon {
+    type: 'end';
+    cursor: 'ew-resize';
+}
+
+export type OnBrushCallbackFn = (start: number, end: number) => void;
```
manzt: Perhaps rename all `OneDim*` to `LinearBrush*` for consistency with the file naming (or `LinearBrush1D`)?

src/gosling-brush/linear-brush-model.ts
```diff
@@ -0,0 +1,238 @@
+import * as d3Selection from 'd3-selection';
+import { drag as d3Drag } from 'd3-drag';
+
+const HIDDEN_BRUSH_EDGE_SIZE = 3;
+
+export type oneDimBrushData = [OneDimBrushBodyData, OneDimBrushStartEdgeData, OneDimBrushEndEdgeData];
```
manzt: I generally avoid `camelCase` naming of types in favor or `PascalCase` naming for types.

src/gosling-brush/linear-brush-model.ts
```diff
@@ -0,0 +1,238 @@
...
+
+export type OnBrushCallbackFn = (start: number, end: number) => void;
+
+interface BrushStyle {
+    color: string;
+    stroke: string;
+    strokeWidth: number;
+    strokeOpacity: number;
+    opacity: number;
+}
```
manzt: import from Gosling schema?

src/gosling-brush/linear-brush-model.ts
```diff
@@ -0,0 +1,238 @@
...
+    };
+
+    // TODO (May-19-2022): A way to pass only the required function (e.g., onRangeBrush) and allow
+    // to use `this` (i.e., instance of GoslingTrack) in the function?
+    /* gosling track */
+    private track: any;
+
+    constructor(
+        selection: d3Selection.Selection<SVGGElement, unknown, HTMLElement, unknown>,
+        HGC: any,
```
manzt: hmm, not exactly `HGC` (`HiGlassComponent`)? perhaps `higlassLibraries`?

manzt: At a minimum, we could start to type out `HGC` in a followup PR.

https://github.com/higlass/higlass/blob/1fcadc24a9915960ee056e5fbd7d1097220e8d0e/app/scripts/configs/available-for-plugins.js#L99-L117

```typescript
// src/types/index.d.ts
declare module '@higlass/available-to-plugins' {
  export interface Libraries {
    d3Array: import('d3-array'),
    d3Axis: import('d3-axis'),
    d3Brush: import('d3-brush'),
    d3Color: import('d3-color'),
    d3Drag: import('d3-drag'),
    d3Dsv: import('d3-dsv'),
    // ...
  }
}
```

```
    constructor(
        selection: d3Selection.Selection<SVGGElement, unknown, HTMLElement, unknown>,
        higlassLibraries: import('@higlass/avaiable-for-plugins').Libraries,
    )
```

src/gosling-brush/linear-brush-model.ts
```diff
@@ -0,0 +1,238 @@
...
+        this.range = null;
+        this.prevExtent = [0, 0];
+        this.data = this.rangeToData(0, 0);
+
+        this.offset = [0, 0];
+        this.size = 0;
+
+        this.externals = {
+            d3Selection: HGC.d3Selection,
+            d3Drag: HGC.d3Drag.drag
```
manzt: perhaps,

```suggestion
            d3Selection: HGC.d3Selection,
            d3Drag: HGC.d3Drag
```

for consistency

src/gosling-brush/linear-brush-model.ts
```diff
@@ -0,0 +1,238 @@
...
+        };
+
+        this.style = {
+            color: style.color ?? '#777',
+            stroke: style.stroke ?? '#777',
+            strokeWidth: style.strokeWidth ?? 1,
+            strokeOpacity: style.strokeOpacity ?? 0.7,
+            opacity: style.opacity ?? 0.3
+        };
```
manzt: Spread operator can come in handy here

```suggestion
        this.style = {
            color: '#777',
            stroke: '#777',
            strokeWidth: 1,
            strokeOpacity: 0.7,
            opacity: 0.3,
            ...style
        };
```

or `this.style = Object.assign(DEFAULTS, style);`

src/gosling-brush/linear-brush-model.ts
```diff
@@ -0,0 +1,238 @@
...
+    /* External libraries that we re-use from HiGlass */
+    private externals: {
+        d3Selection: typeof d3Selection;
+        d3Drag: typeof d3Drag;
+    };
+
+    // TODO (May-19-2022): A way to pass only the required function (e.g., onRangeBrush) and allow
+    // to use `this` (i.e., instance of GoslingTrack) in the function?
+    /* gosling track */
+    private track: any;
```
manzt: Can you elaborate on this? Is there a reason why `track` needs to be stored on the class itself? Why not have a track _use_ the brush?

src/gosling-brush/linear-brush-model.ts
```diff
@@ -0,0 +1,238 @@
...
+        this.offset = [offsetX, offsetY];
+        return this;
+    }
+
+    /**
+     * Update brush data based on the positions of two edges.
+     */
+    public updateRange(range: [number, number] | null) {
+        if (range) {
+            this.range = range.sort((a, b) => a - b) as [number, number];
```
manzt: Think min/max-ing the range could be a little clearer

```javascript
this.rangeToData(Math.min(range), Math.max(range))
```

src/gosling-mouse-event/polygon.ts
```diff
@@ -12,6 +12,44 @@ export const isPointNearPoint: (point: [number, number], point2: number[], radiu
...
+ */
+export const isAnyPointsWithinRange: (range: [number, number], path: number[]) => boolean = ([x1, x2], path) => {
+    let lx;
+    let isWithin = true;
+    for (let i = 0; i < path.length; i += 2) {
+        lx = path[i];
+        isWithin = isWithin && isPointInsideRange([x1, x2], lx);
+    }
+    return isWithin;
+};
```
manzt: I think this can just early exit, unless I'm missing something. Also why `i += 2` in for-loop?

```suggestion
export function isAnyPointsWithinRange([x1, x2]: [number, number], path: number[]): boolean {
    let lx;
    for (let i = 0; i < path.length; i += 2) {
        lx = path[i];
        if (!isPointInsideRange([x1, x2], lx)) {
          return false;
        }
    }
    return true;
};
```

if it were every element, this would be easier with `path.every(p => isPointInsideRange(range, p))`

src/gosling-track/gosling-track.ts
```diff
@@ -2,14 +2,14 @@ import * as PIXI from 'pixi.js';
 import PubSub from 'pubsub-js';
 import * as uuid from 'uuid';
 import { isEqual, sampleSize, uniqBy } from 'lodash-es';
-import { scaleLinear } from 'd3-scale';
+import { ScaleLinear } from 'd3-scale';
```
manzt: I think use `import type` again, for consistency (for maybe all the new imports in this file?) I think all are types.

src/gosling-track/gosling-track.ts
```diff
@@ -120,19 +136,34 @@ function GoslingTrack(HGC: any, ...args: any[]): any {
...
             // Enable click event
+            this.isBrushActivated = false;
             this.pMask.interactive = true;
+            this.gBrush = HGC.libraries.d3Selection.select(this.context.svgElement).append('g');
+            this.mRangeBrush = new OneDimBrushModel(
+                this.gBrush,
+                HGC.libraries,
+                this,
+                this.options.spec.experimental?.brush
+            );
```
manzt: Hmm I think I see the issue. Track implements a method that the brush needs to call at somepoint in its lifecycle?

You can bind the class instance (this) and pass in the method as a regular function

```typescript
this.mRangeBrush = new OneDimBrushModel(
    this.gBrush,
    HGC.libraries,
    this.onRangeBrush.bind(this),
    this.options.spec.experimental?.brush,
);
```

src/gosling-track/gosling-track.ts
```diff
@@ -327,6 +364,7 @@ function GoslingTrack(HGC: any, ...args: any[]): any {
                 this.gLegend.remove();
                 this.gLegend = null;
             }
+            this.mRangeBrush.remove();
```
manzt, line 380: Can you leave a comment describing why removed twice if layout if circular?

src/gosling-track/gosling-track.ts
```diff
@@ -1014,52 +1061,182 @@ function GoslingTrack(HGC: any, ...args: any[]): any {
...
+         */
+        highlightMarks(
+            g: Graphics,
+            marks: MouseEventData[],
+            stroke: string,
+            strokeWidth: number,
+            strokeOpacity: number,
+            color: string,
+            fillOpacity: number
+        ) {
```
manzt: In case there end up being more options, I'd change this signature to:

```typescript

        highlightMarks(
            g: Graphics,
            marks: MouseEventData[],
            options: {
                stroke: string,
                strokeWidth: number,
                strokeOpacity: number,
                color: string,
                fillOpacity: number
            }
        ) {
```

src/gosling-track/gosling-track.ts
```diff
@@ -1014,52 +1061,182 @@ function GoslingTrack(HGC: any, ...args: any[]): any {
...
+            }
+
+            if (capturedElements.length !== 0) {
+                // selection effect graphics
+                const g = this.pMouseSelection;
+                const stroke = this.options.spec.experimental?.selection?.stroke ?? 'black';
+                const strokeWidth = this.options.spec.experimental?.selection?.strokeWidth ?? 1;
+                const strokeOpacity = this.options.spec.experimental?.selection?.strokeOpacity ?? 1;
+                const color = this.options.spec.experimental?.selection?.color ?? 'none';
+                const fillOpacity = this.options.spec.experimental?.selection?.opacity ?? 1;
```
manzt: and then this can be simplified

```typescript
let options = {
    stroke: 'black', 
    strokeWidth: 1,
    strokeOpacity: 1,
    color: 'none',
    opacity: 1,
    ...this.options.spec.experimental?.selection
};

**Review by manzt — COMMENTED** [review]
Nice work, Sehi! 

It's great to see this feature coming along. I don't think I have anything critically blocking this PR since it is under and experimental flag, but I offered some suggestions about API for end users. Additionally, we should think more deeply about how the track and brush interact with each other, but this can all be refactored if necessary.

File: src/gosling-brush/linear-brush-model.ts (hunk 15 above)
Comment (manzt) [reply src/gosling-brush/linear-brush-model.ts]: EDIT: see comment below

File: editor/editor.tsx (hunk 2 above)
Comment (sehilyi) [reply editor/editor.tsx]: Ah, right. I will keep that in mind and avoid importing hidden components.

File: src/gosling-brush/linear-brush-model.ts (hunk 10 above)
Comment (sehilyi) [reply src/gosling-brush/linear-brush-model.ts]: I like the convention, and we should use this in the project in a consistent manner.

File: src/gosling-brush/linear-brush-model.ts (hunk 14 above)
Comment (sehilyi) [reply src/gosling-brush/linear-brush-model.ts]: I like the `Object.assign()`!

File: src/gosling-brush/linear-brush-model.ts (hunk 16 above)
Comment (sehilyi) [reply src/gosling-brush/linear-brush-model.ts]: Nice!

File: src/gosling-mouse-event/polygon.ts (hunk 17 above)
Comment (sehilyi) [reply src/gosling-mouse-event/polygon.ts]: Cool, I like `every()`! The `path` is a flat array that contains both x and y values and this array only uses x. I've changed this to

```
path.filter((_, i) => i % 2 === 0).every(x => isPointInsideRange(range, x));
```

File: src/gosling-brush/linear-brush-model.ts (hunk 12 above)
Comment (sehilyi) [reply src/gosling-brush/linear-brush-model.ts]: Renamed the variable, and I agree – type definitions on the libraries will be really helpful.

File: src/gosling-track/gosling-track.ts (hunk 20 above)
Comment (sehilyi) [reply src/gosling-track/gosling-track.ts:380]: Left comment statements on the `remove()` and `rerender()` functions.

Usually, you only need to remove any graphics on the `remove()` function. This function is called when a track is removed from the GoslingComponent (i.e., end of the rendering cycle).

I am additionally removing the range brush on `rerender()`. This `rerender()` function is called when a spec for the current track is changed by a user. Since we do not support a circular brush at the moment, I am removing the linear brush when the spec is changed from linear to circular:

```ts
rerender(newOption) {
  if (newOption.spec.layout === 'circular') {
    // this means the spec has been changed to use circular layouts
    this.mRangeBrush.remove();
  }
```

For example, you can see that any range brushes that you made on a default example of the editor (i.e., "Visual Encoding") are being removed when you change the `layout` to `circular`.

src/gosling-track/gosling-track.ts
sehilyi: Ah, `.bind(this)` is exactly what I was looking for 😍

src/core/gosling.schema.ts
sehilyi: Yes, I think it makes sense to turn these mouse events off by default. 

Since we currently have three mouse event listeners (i.e., `click`, `mouseover`, `rangeselect`), how about adding the following properties to enable using the events?

```js
experimental?: {
  // this will eventually be a track-level property, i.e., sibling of `mark`
  mouseEvent: boolean | {
    click: boolean,
    hover: boolean,
    rangeSelect: boolean,
    // or `enableClick`, `enableHover`, `enableRangeSelect`
    groupMarksByField?: string,
    ...
  }
}
// `mouseEvent: true` will turn all three events on
```

This means that if users want to listen to mouse events through API, they will need to turn these on, e.g., `mouseEvent: { hover: true }` and `api.subscript('mouseover', ...)`.

Also, we can move the current styling objects to the `style` object:

```diff
experimental?: {
-  hovering: { color?: string, ...}
-  selection: { color?: string, ...}
-  brush: { color?: string, ...}
},
...,
style?: {
+  hover: { color?: string, ...}
+  select: { color?: string, ...}
+  brush: { color?: string, ...}
}
```

src/core/api.ts
sehilyi: I like the idea! Given that we are planning to support custom chromosome sizes (#582), this approach is making more sense to me. Consumers can still easily make a flat array of `data` if they want.

Regarding the naming and structure, what do you think of the following, which is more consistent with `DomainChrInterval` in our grammar?

```js
// [
//   { chromosome: 'chr1', interval: [100000, 248956422], data: [/* ... */] },
//   { chromosome: 'chr2', interval: [0, 242193529], data: [/* ... */] },
//   { chromosome: 'chr3', interval: [0, 20000], data: [/* ... */] },
// ]
```

```ts
// gosling.schema.ts
export interface DomainChrInterval {
    chromosome: Chromosome;
    interval: [number, number];
}
```

src/core/api.ts
sehilyi: Will address this together with the next comment.

src/gosling-brush/linear-brush-model.ts
```diff
@@ -0,0 +1,235 @@
+import type * as D3Selection from 'd3-selection';
+import type * as D3Drag from 'd3-drag';
+import { BrushAndMarkHighlightingStyle } from '@gosling.schema';
+
+const HIDDEN_BRUSH_EDGE_SIZE = 3;
+
+export type LinearBrushData = [LinearBrushBodyData, LinearBrushStartEdgeData, LinearBrushEndEdgeData];
+
+export type LinearBrushDataUnion = LinearBrushBodyData | LinearBrushStartEdgeData | LinearBrushEndEdgeData;
```
manzt: Small nit. Rather than duplicating each of the types in `LinearBrushData` and `LinearBrushDataUnion`. It might be easier to just refer to the union _from_ the tuple using `[number]` indexing. E.g.,

```typescript
import type { LinearBrushData } from './linear-brush-model';
const dragged = (d: LinearBrushData[number]) => { /* ... */ };

File: src/core/gosling.schema.ts (hunk 6 above)
Comment (sehilyi) [reply src/core/gosling.schema.ts:254]: (Btw, I gotta use a consistent term between hover and mouseover..)

File: src/gosling-brush/linear-brush-model.ts (hunk 23 above)
Comment (sehilyi) [reply src/gosling-brush/linear-brush-model.ts]: This is neat!

File: src/core/api.ts (hunk 5 above)
Comment (manzt) [reply src/core/api.ts:23]: > Regarding the naming and structure, what do you think of the following, which is more consistent with `DomainChrInterval` in our grammar?

I think this is a great idea!

File: src/core/gosling.schema.ts (hunk 6 above)
Comment (manzt) [reply src/core/gosling.schema.ts:254]: Since it's under the `experimental` key in the schema, I think we have some flexibility in deciding how to finally expose this behavior. Minor suggestion, but perhaps `mouseEvents` instead of `mouseEvent`. 

```typescript
type MouseEvent = 'click' | 'hover' | 'rangeSelect';

{
  experimental?: {
    mouseEvents: boolean | {
      [Event in MouseEvent]: boolean;
      groupMarksByField?: string;
    }
  }
}
```

manzt: And I agree, the `MouseEvent` type should be identical to the pubsub event bus keys. E.g. We should avoid having a user configure `'rangeSelect'` event and then need subscribe to `api.subscribe('rangeselect')`, casing should be the same.

src/gosling-track/gosling-track.ts
manzt: Ok! Thanks for the clarification.

src/core/gosling.schema.ts
sehilyi: Sound good. I will make this change in a follow-up PR.

src/core/api.ts
sehilyi: I realized that we need to consider cases where marks do not correspond to a single chromosome. For example, a user can select a link that connects between `chr1` and `chr7`.

We can perhaps include the same value in both `chromosome: 'chr1', data: [...]` and `chromosome: 'chr7', data: [...]`, but API consumers will need to remove the duplication by themselves if they want a flat array.

Unsure at the moment what method would be better between (1) a flat array vs. (2) per-chromosome arrays. The flat array is simpler in terms of the structure, and consumers can still get per-chromosome arrays by looking up raw data fields (e.g., a `chr` column).

sehilyi:
I will merge this PR and will open follow-up PRs to address unresolved issues.

src/core/api.ts
manzt: That's a good point, and something I had not thought of. It would be more simple to return the flat array as you suggested, since these correspond to the "selected" entries in the data. I wonder if we could do something similar to what you had before then,

```typescript
type GenomicRegion = [chr: string, start: number, end: number];

export interface RangeMouseEventData extends CommonEventData {
  interval: [GenomicRegion, GenomicRegion],
}
```

I guess users have enough information to understand which chromosomes are within this `interval` from the assembly (chromsizes).

sehilyi: Sounds good. 

Yes, users will have enough information in all cases. When users use a custom `chromsizes`, they can refer to the order in the `chromsizes` array.

PR #667, opened 2022-04-26
feat: capture mouse events on marks for tooltips, hovering effects, and APIs

sehilyi:
# Updates

- Created a class that handles mouse events in individual gosling tracks (`gosling-mouse-event`).
- `click` and `mouseover` events can be subscribed via API which returns raw JSON data with genomic position

```js
{ genomicPosition: 'chr2:13,000', data: { val: 29, cat: 'sample 3', ...  } }
```

- Hovering behavior can be customized using experimental track-level properties:

```js
..., // track definition
experimental: {
    hovering: {
        enableMultiHovering: true,
        enableGroupHovering: true,
        searchGroupByField: "chr.",
        showHoveringOnTheBack: true,
        stroke: 'red',
        strokeWidth: 2,
        color: 'blue'
    }
}
```

- Example screenshot:

Screen Shot 2022-05-09 at 17 03 08

- Events are either point-based (point, line, area), line-based (line-based links), or polygon-based (triangle, text, band-based links). We could support others as well.

# Open Issues

- ~~How to enable customizing hovering effects? (e.g., outline color and width, fill color, opacity)~~
   - refer to https://vega.github.io/vega-lite/examples/interactive_bar_select_highlight.html
- ~~Single mark hovering vs. multi-mark hovering when multiple marks are overlapped (e.g., track-level property something like `enableMultiSelection` can handle this).~~
- ~~Hovering on glyphs (e.g., gene models). We can somehow use unique IDs (e.g., searching for data records with the same ID and highlighting all records at once). Users can specify ID columns from the data (e.g., `data.idField`). Otherwise, we can assign unique IDs to individual data records.~~
- Range or Lasso selection (e.g., genomic range or value range or 2D)
- Simpler mouse events that captures mouse in and out on an entire track (related to #669)
- Linking between tracks (e.g., hovering on a track highlights related marks on another)

### Additional Screenshots

Screen Shot 2022-05-09 at 18 17 55
Screen Shot 2022-05-09 at 18 17 50
Screen Shot 2022-05-09 at 18 17 42
Screen Shot 2022-05-09 at 18 17 30
Screen Shot 2022-05-09 at 18 17 22

Screen Shot 2022-05-09 at 17 03 08

Fix #31 
Fix #30

PR #665, opened 2022-04-26
feat: treeshake pkg info in module export

manzt:
The code below imports `package.json` as a javascript object and 
then exports the `name` and `version` fields as a module.

```javascript
// input.js
import pkg from "../package.json";
export const { name, version } = pkg;

// bundle.js
const name$1 = "gosling.js";
const author = "Sehi L'Yi";
const version$1 = "0.9.17";
const license = "MIT";
const repository = {
  type: "git",
  url: "https://github.com/gosling-lang/gosling.js"
};
const homepage = "https://gosling-lang.github.io/gosling.js/";
const main = "dist/gosling.js";
const module = "dist/gosling.es.js";
const types = "dist/src/index.d.ts";
const files = [
  "dist"
];
const exports = {
  ".": "./dist/gosling.es.js",
  "./embed": "./dist/embed/index.js"
};
const scripts = {
  start: "vite --mode editor",
  "start-embed": "vite",
  build: "run-p build-types build-lib",
  "build-lib": "vite build --mode lib && node scripts/build-umd && node scripts/build-embed",
  "build-types": "tsc --emitDeclarationOnly",
  "build-editor": "vite build",
  preview: "vite preview",
  test: "jest editor/ src/",
  changelog: "conventional-changelog -p angular -i CHANGELOG.md -s -r 0",
  format: "eslint src/ editor/ --fix && prettier 'editor/**/*.css' --write",
  schema: "ts-json-schema-generator -p src/index.ts -f tsconfig.json -t GoslingSpec --no-type-check --no-ref-encode > schema/gosling.schema.json",
  "schema-higlass": "ts-json-schema-generator -p src/index.ts -f tsconfig.json -t HiGlassSpec --no-type-check --no-ref-encode > schema/higlass.schema.json",
  "schema-theme": "ts-json-schema-generator -p src/index.ts -f tsconfig.json -t Theme --no-type-check --no-ref-encode > schema/theme.schema.json",
  "schema-template": "ts-json-schema-generator -p src/index.ts -f tsconfig.json -t TemplateTrackDef --no-type-check --no-ref-encode > schema/template.schema.json",
  "schema-versioning": "mkdir -p schema/history/$npm_package_version && cp schema/gosling.schema.json schema/history/$npm_package_version/gosling$npm_package_version.schema.json && cp src/core/gosling.schema.ts schema/history/$npm_package_version/gosling$npm_package_version.schema.ts",
[...]

PR #694, opened 2022-05-17
chore: (re)treeshake pkg info in module export

manzt:
#665 was reversed with #667?


sehilyi (approved):
Sorry, I probably did force push without pulling this change. I will be more cautious next time.

Fortunately, this seems to be the only change that is reversed...

PR #697, opened 2022-05-17
[draft] maint: monorepo

manzt:
The number of features and tools in Gosling continues to grow. So much so, that the `gosling.js` repo contains not only the core renderer, but the Editor as well.
This "package" is then quite large and it's challenging to know which dependencies are related to validation, templates, rendering, the editor, etc. Additionally, `gosling-theme` lives in a separate repository and there is no reason why is shouldn't be included here as well.

This PR lays the foundation for building Gosling in a monorepo using [`pnpm` workspaces](https://pnpm.io/workspaces). This PR breaks up the source code for gosling (the package)
and the editor into separate packages (under the `@gosling/*`) namespace under packages. We had somewhat of a hacked together monorepo with `vite` configs recently and many build scripts.

### Key ideas

Each folder in `packages/` is a JavaScript package, and can be included as a dependency of any other package in the folder with the `workspace:` specifier:

```json
{
    "name": "gosling.js",
    "version": "0.0.0",
    "type": "module",
    "main": "index.ts",
    "types": "index.ts",
    "dependencies": {
        "@gosling/core": "workspace:*",
        "@gosling/schema": "workspace:*"
    }
}
```

Therefore, in the source code rather than nested relative imports like,

```typescript
import type { Datum } from '../../src/core/gosling.schema';
```

use the package import,

```typescript
import type { Datum } from '@gosling/schema';
```

In this process I found _many_ circular dependencies (e.g. `@gosling/core` imports from `@gosling/editor`!).
This should be corrected if possible, and perhaps we can think of better module structure. For now, these packages will
remain private to the repository and will will only continue to publish `gosling.js` (`packages/main`).

I still have some things to work on with CI and build scripts, but there is enough here to experiment with if anyone is curious.
Currently because we are not publishing each package `@gosling/*` the source code is TypeScript for those libraries.

# TODO:

- [x] linting
- [x] formatting
- [x] tests
- [ ] setup recursive `pnpm dev`  to watch files and start editor
- [ ] restructure build for monorepo


manzt:
This is going to be very painful to apply all the recent changes in Gosling, so I'm somewhat inclined to close this PR and open another. Will probably have time to look into more closely after ISMB.

etowahadams:
I've started thinking about the restructuring do making things a monorepo and have been looking into some of the circular dependencies you mentioned here, specifically `@gosling/core` imports from `@gosling/editor`. 

The only dependencies from `@gosling/core` in the editor that I see are schema imports (`import type { GoslingSpec } from 'gosling.js';`) and a specific core util function: 
https://github.com/gosling-lang/gosling.js/blob/526882fb82463ad4262568d9569840431c709fc9/editor/Editor.tsx#L24

Were these the dependencies you were thinking of? Or am I missing something else?

manzt:
I don't recall, and since closing this PR I'm fairly sure that there have been many improvements to the improved organization of gosling code.

PR #698, opened 2022-05-19
fix: use assembly in vcf data fetcher

sehilyi:
Reference: https://github.com/gosling-lang/gosling.js/issues/582#issuecomment-1131808700

Similar should be fixed in other data fetchers, too

### DIFF
--- a/src/data-fetcher/vcf/gosling-vcf-data.ts
+++ b/src/data-fetcher/vcf/gosling-vcf-data.ts
@@ -38,12 +38,15 @@ class GoslingVcfData {
         this.tbiVCFParser = null;
 
         this.initPromise = this.worker.then((tileFunctions: any) => {
-            const vcfUrl = dataConfig.url;
-            const tbiUrl = dataConfig.indexUrl ?? `${vcfUrl}.tbi`;
-            const chromSizesUrl = `https://s3.amazonaws.com/gosling-lang.org/data/${dataConfig.assembly}.chrom.sizes`;
-            const sampleLength = dataConfig.sampleLength ?? 1000;
-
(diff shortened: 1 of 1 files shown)


src/data-fetcher/vcf/gosling-vcf-data.ts
```diff
@@ -36,13 +38,14 @@ class GoslingVcfData {
...
-            dataConfig['chromSizesUrl'] = 'https://s3.amazonaws.com/gosling-lang.org/data/hg19.chrom.sizes';
-            dataConfig['sampleLength'] = dataConfig.sampleLength ?? 1000;
-
             return tileFunctions
-                .init(this.uid, dataConfig.vcfUrl, dataConfig.tbiUrl, dataConfig.chromSizesUrl, dataConfig.sampleLength)
+                .init(
+                    this.uid,
+                    dataConfig.url,
+                    dataConfig.indexUrl,
+                    `https://s3.amazonaws.com/gosling-lang.org/data/${dataConfig.assembly}.chrom.sizes`,
```
manzt: Isn't there a utility for this?

```javascript
GET_CHROM_SIZES(this.assembly).path
```

manzt (changes requested):
Looks good. Just a comment to avoid hard-coding URLs. This will be important if for custom chromosomes sizes.

manzt:
It's important to note that our dependency on these self-hosted TSVs on s3 means that Gosling can't be used offline.

src/data-fetcher/vcf/gosling-vcf-data.ts
sehilyi: Yes, it is. Thanks for this suggestion.

sehilyi:
Yes, I think when we support the custom assemblies, we should remove the dependencies entirely and use the inline URLs

manzt (approved):

PR #700, opened 2022-05-20
chore(editor): open new window using current domain

sehilyi:


editor/editor.tsx
```diff
@@ -662,7 +662,7 @@ function Editor(props: RouteComponentProps) {
             >
                 <span
                     style={{ cursor: 'pointer', lineHeight: '40px' }}
-                    onClick={() => window.open('https://gosling.js.org', '_blank')}
+                    onClick={() => window.open(`${window.location.origin}${window.location.pathname}`, '_blank')}
```
manzt: I believe this should work with `window.open(window.location.pathname, '_blank')`

manzt: URLs paths are resolved to the current origin

manzt (approved):

editor/editor.tsx
sehilyi: Thanks! Good to know

PR #701, opened 2022-05-20
fix(editor): minor bugs when importing & exporting examples

wangqianwen0418:
fixes several minor bugs, including:
- only copy URL spec from JSON editor even when the JS editor is activated
- no feedback after users copy the URL spec & execCommand("copy") is obsolete
- cannot open demos in the gallery if the URL has spec params


editor/editor.tsx
```diff
@@ -866,26 +851,34 @@ function Editor(props: RouteComponentProps) {
...
+                                    );
                                     const url = `${window.location.origin}${window.location.pathname}?full=${isHideCode}&spec=${crushedSpec}`;
-                                    const element = document.getElementById('spec-url-exporter');
-                                    (element as any).type = 'text';
-                                    (element as any).value = url;
-                                    (element as any).select();
-                                    document.execCommand('copy');
-                                    (element as any).type = 'hidden';
+
+                                    // fix execCommand("copy") is obsolete
```
sehilyi:
```suggestion
```

editor/editor.tsx
```diff
@@ -312,20 +308,9 @@ function Editor(props: RouteComponentProps) {
...
-        } else if (urlGist) {
-            setCode(emptySpec());
-        } else {
-            const jsonCode = stringifySpec(demo.spec as gosling.GoslingSpec);
-            setCode(jsonCode);
-            setJsCode(demo.specJs ?? json2js(jsonCode));
-        }
+        const jsonCode = stringifySpec(demo.spec as gosling.GoslingSpec);
+        setCode(jsonCode);
+        setJsCode(demo.specJs ?? json2js(jsonCode));
```
sehilyi: I assume this change fixes a bug?

editor/editor.tsx
```diff
@@ -866,26 +851,34 @@ function Editor(props: RouteComponentProps) {
                         </span>
                         <span
                             title={
-                                code.length <= LIMIT_CLIPBOARD_LEN
+                                stringifySpec(goslingSpec || '').length <= LIMIT_CLIPBOARD_LEN
```
sehilyi: I see `stringifySpec(goslingSpec || '')` is used multiple times. Would it make sense to let `stringifySpec` deal with `undefined`, something like?:

```js
const stringifySpec = (spec: string | gosling.GoslingSpec | undefined): string => {
    if (!spec) return '';
    else if (typeof spec === 'string') return spec;
    else return stringify(spec);
};
```

sehilyi (dismissed):
Thanks for the updates. I like the message dialog upon saving the URL.

I noticed that opening a URL with a `spec` parameter is broken (i.e., the default spec is shown regardless of the embedded spec). This seems to work on current gosling.js.org. Steps to reproduce:

1. Open `localhost:3000`
2. Select an example with a short spec (e.g., Rule Mark)
3. Click on a SAVE URL button
4. Open the saved page

Can you check this issue?

editor/editor.tsx
wangqianwen0418: yes, it fixes a bug. Specifically, if you open the current gosling editor with the urlparams (e.g., ?example=xxxx), the editor will not response when you click demo in the example gallery.

editor/editor.tsx
wangqianwen0418: Good point!

editor/editor.tsx
```diff
@@ -866,26 +851,34 @@ function Editor(props: RouteComponentProps) {
...
                             }
                             onClick={() => {
-                                if (code.length <= LIMIT_CLIPBOARD_LEN) {
+                                if (stringifySpec(goslingSpec || '').length <= LIMIT_CLIPBOARD_LEN) {
                                     // copy the unique url to clipboard using `<input/>`
-                                    const crushedSpec = encodeURIComponent(JSONCrush.crush(code));
+                                    const crushedSpec = encodeURIComponent(
+                                        JSONCrush.crush(stringifySpec(goslingSpec || ''))
+                                    );
                                     const url = `${window.location.origin}${window.location.pathname}?full=${isHideCode}&spec=${crushedSpec}`;
```
wangqianwen0418, line 869: seems to be some cache-related issue. Change the url prefix will lead to workable urlspec

editor/editor.tsx
wangqianwen0418: sorry for overlooking this issue while merging the master branch.
It seems to be related to the URL cache.
Changing the URL prefix to "https://gosling.js.org/" (as in the commit https://github.com/gosling-lang/gosling.js/pull/701/commits/bdf12cfce11fa9fcd731f91b4437a27c59bb552f) will fix this issue.

sehilyi:
@wangqianwen0418 I think it is still not working... Your last [commit](https://github.com/gosling-lang/gosling.js/commit/bdf12cfce11fa9fcd731f91b4437a27c59bb552f) actually opens the Editor that is already deployed.

```diff
- const url = `${window.location.origin}${window.location.pathname}?full=${isHideCode}&spec=${crushedSpec}`;
+ const url = `https://gosling.js.org/?full=${isHideCode}&spec=${crushedSpec}`;
```

Since the current Editor works fine with the `spec=` parameter, you can see the embedded example correctly rendered on the page. However, if you change the URL to the `localhost:3000`, I still see the default example.


I assume that somehow the embedded spec is not correctly used to render visualization in the current code.

wangqianwen0418:
> @wangqianwen0418 I think it is still not working... Your last [commit](https://github.com/gosling-lang/gosling.js/commit/bdf12cfce11fa9fcd731f91b4437a27c59bb552f) actually opens the Editor that is already deployed.
> 
> ```diff
> - const url = `${window.location.origin}${window.location.pathname}?full=${isHideCode}&spec=${crushedSpec}`;
> + const url = `https://gosling.js.org/?full=${isHideCode}&spec=${crushedSpec}`;
> ```
> 
> Since the current Editor works fine with the `spec=` parameter, you can see the embedded example correctly rendered on the page. However, if you change the URL to the `localhost:3000`, I still see the default example.
> 
> I assume that somehow the embedded spec is not correctly used to render visualization in the current code.

Thanks for explaining the issue!
It should be fixed in the latest commit

sehilyi (approved):
Looks good!

PR #703, opened 2022-05-22
chore: use github warning markdown

manzt:
GitHub introduced support for `**Note**` and `**Warning**` in markdown. I've applied the change here, which you can see in the markdown preview.

As a general question, should we change the wording in the warning? It seems to me that Gosling has officially been "released" but the grammar is subject to changes?


sehilyi (commented):
We can say that the grammar is subject to change before the release of version 1.0.0? After several missing features are added, I expect that the grammar will be stable and we can remove this description at that point.

sehilyi (approved):

PR #704, opened 2022-05-24
feat(editor): visualize JS spec using Gist URL

wangqianwen0418:
enable users to open gist examples that are written in js/ts

e.g., http://localhost:3000/?gist=wangqianwen0418/d92364a49e434ee22acdf2b16de11d76

### DIFF
--- a/editor/editor.tsx
+++ b/editor/editor.tsx
@@ -351,7 +351,7 @@ function Editor(props: RouteComponentProps) {
             setCode(urlSpec);
             setJsCode(json2js(urlSpec));
         } else if (urlGist) {
-            setCode(emptySpec());
+            setCode(emptySpec('loading....'));
         } else {
             const jsonCode = stringifySpec(demo.spec as gosling.GoslingSpec);
             setCode(jsonCode);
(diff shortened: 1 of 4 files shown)


editor/editor.tsx
```diff
@@ -169,19 +180,31 @@ const fetchSpecFromGist = async (gist: string) => {
...
+        let jsonCode: string | null, jsCode: string, language: string;
+        if (isJSON(code)) {
+            language = 'json';
+            jsonCode = resolveRelativeCsvUrls(code, specUrl);
+            jsCode = json2js(jsonCode);
+        } else {
+            jsCode = code || '';
+            jsonCode = '{}';
+            language = 'javascript';
+        }
```
sehilyi: Given that `code` can be either `null` or JS in this code block, it would be better to separate the two cases. 

If `code` is not found, we can set the `language` as 'json' since it is the default one of the Editor.

If `code` is JS, I think we need to set the JSON spec that is generated by the JS code, instead of the empty spec ('{}').

sehilyi (dismissed):
Looks good 👍

Just one comment--The editor shows an empty JSON spec (`"{}"`) when a JS spec is loaded via Gist. We can instead show a JSON spec that is generated by the JS spec.

editor/editor.tsx
wangqianwen0418: Good point! 
Will update them accordingly soon

editor/editor.tsx
```diff
@@ -277,7 +302,7 @@ function Editor(props: RouteComponentProps) {
     const gosRef = useRef<gosling.GoslingRef>(null);
 
     const debounceCodeEdit = useRef(
-        debounce((code: string, language) => {
+        debounce((code: string, language: string) => {
```
sehilyi: Would be good to define a type for `langauge`.

```ts
type EditorLangauge = 'json' | 'javascript';
...
debounce((code: string, language?: EditorLangauge) => {
```

sehilyi (approved):
Looks good. Left a minor comment regarding the type definition of `language`. There is no need to review again for merging.

PR #706, opened 2022-05-26
refactor: use dynamic import of ts, update target to ESM

manzt:
I noticed the Editor imports the TypeScript compiler even if it is never used the by user. This PR makes `transpile` an asynchronous function which dynamically imports `typescript`. This means that TypeScript will only need to be loaded if it is used in the application by the end user. Additionally, I saw there were some hacks to convert the `commonjs` output from the compiler to ESM. This can be avoided entirely by setting a more modern target (`ES2018`) for the compiler.


wangqianwen0418 (approved):

wangqianwen0418:
@manzt, thanks a lot for the PR! It looks great 👍

PR #708, opened 2022-05-27
chore: set `importsNotUsedAsValues` compiler flag

manzt:
Adds the `importsNotUsedAsValues: "error"` TS compiler option to enforce [type-only imports](https://devblogs.microsoft.com/typescript/announcing-typescript-3-8-beta/#type-only-imports-exports) for types in the codebase.

Imports which are included but not used as values (e.g., only as a type) will raise an error. This is now enforced by the compiler for consistency across the codebase.

```diff
- import { GoslingSpec } from '@gosling.schema';
+ import type { GoslingSpec } from '@gosling.schema';

function example(spec: GoslingSpec) { 
  /* ... */
}
```

Additional changes:

- uses of `assign` from `lodash-es` are replaced with `Object.assign`
- the type `PREDEFINED_COLORS` is renamed to `PredefinedColors` since it refers to a type, not a constant value.

### DIFF
--- a/src/core/gosling.schema.ts
+++ b/src/core/gosling.schema.ts
@@ -1,4 +1,4 @@
-import { Chromosome } from './utils/chrom-size';
+import type { Chromosome } from './utils/chrom-size';
 
 /* ----------------------------- ROOT SPEC ----------------------------- */
 export type GoslingSpec =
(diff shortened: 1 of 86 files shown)


src/core/gosling.schema.ts
```diff
@@ -651,8 +651,8 @@ export type FieldType = 'genomic' | 'nominal' | 'quantitative';
 export type ValueExtent = string[] | number[];
 export type GenomicDomain = DomainInterval | DomainChrInterval | DomainChr | DomainGene;
 export type Domain = ValueExtent | GenomicDomain;
-export type Range = ValueExtent | PREDEFINED_COLORS;
-export type PREDEFINED_COLORS = 'viridis' | 'grey' | 'spectral' | 'warm' | 'cividis' | 'bupu' | 'rdbu' | 'hot' | 'pink';
+export type Range = ValueExtent | PredefinedColors;
+export type PredefinedColors = 'viridis' | 'grey' | 'spectral' | 'warm' | 'cividis' | 'bupu' | 'rdbu' | 'hot' | 'pink';
```
sehilyi, line 655: Thanks for updating the letter case!

sehilyi (approved):
Great 👍!

Issue #710, opened 2022-06-01
Finalize property names and defaults for mouse events


sehilyi:
Since we currently have three mouse event listeners (i.e., `click`, `mouseover`, `rangeselect`), how about adding the following properties to enable using the events?

```js
experimental?: {
  // this will eventually be a track-level property, i.e., sibling of `mark`
  mouseEvent: boolean | {
    click: boolean,
    hover: boolean,
    rangeSelect: boolean,
    // or `enableClick`, `enableHover`, `enableRangeSelect`
    groupMarksByField?: string,
    ...
  }
}
// `mouseEvent: true` will turn all three events on
```

This means that if users want to listen to mouse events through API, they will need to turn these on, e.g., `mouseEvent: { hover: true }` and `api.subscript('mouseover', ...)`.

Also, we can move the current styling objects to the `style` object:

```diff
experimental?: {
-  hovering: { color?: string, ...}
-  selection: { color?: string, ...}
-  brush: { color?: string, ...}
},
...,
style?: {
+  hover: { color?: string, ...}
+  select: { color?: string, ...}
+  brush: { color?: string, ...}
}
```

_Originally posted by @sehilyi in https://github.com/gosling-lang/gosling.js/pull/693#discussion_r883937292_

Refer to the discussion thread above.

sehilyi:
Since it's under the `experimental` key in the schema, I think we have some flexibility in deciding how to finally expose this behavior. Minor suggestion, but perhaps `mouseEvents` instead of `mouseEvent`. 

```typescript
type MouseEvent = 'click' | 'hover' | 'rangeSelect';

{
  experimental?: {
    mouseEvents: boolean | {
      [Event in MouseEvent]: boolean;
      groupMarksByField?: string;
    }
  }
}
```

_Originally posted by @manzt in https://github.com/gosling-lang/gosling.js/pull/693#discussion_r885696225_

sehilyi:
And I agree, the `MouseEvent` type should be identical to the pubsub event bus keys. E.g. We should avoid having a user configure `'rangeSelect'` event and then need subscribe to `api.subscribe('rangeselect')`, casing should be the same.

_Originally posted by @manzt in https://github.com/gosling-lang/gosling.js/pull/693#discussion_r885699372_

sehilyi:
[...]

PR #712, opened 2022-06-02
feat: turn off mouse events by default; rename API keys; rename and restructure mouse event properties

sehilyi:
- Use camelCases for API keys (e.g., `rangeSelect`, `mouseOver`)
- Change structure and rename of mouse events
```ts
type MouseEvent = 'click' | 'mouseOver' | 'rangeSelect'; // 'mouseOver' was 'hover'

{
  experimental?: {
    mouseEvents: boolean | {
      [Event in MouseEvent]: boolean;
      groupMarksByField?: string;
      enableMouseOverOnMultipleMarks?: boolean; // was 'enableMultiHovering'
    }
  }
}
```
- Do not call APIs and do not display relevant visual effects for mouse events when they are not turned on by a user, i.e., a brush for the range select does not appear unless either `experimental.mouseEvents === true` or `experimental.mouseEvents.rangeSelect === true`.

Toward #710

### DIFF
--- a/src/core/api.ts
+++ b/src/core/api.ts
@@ -27,17 +27,17 @@ export interface RangeMouseEventData extends CommonEventData {
 //
 // Add named events using a string union for `EventName`
 //
-// - Two different events ('mouseover' & 'my-event') with the same payload
+// - Two different events ('mouseOver' & 'my-event') with the same payload
 //
-// PubSubEvent<'mouseover' | 'my-event', { same: 'payload' }>
+// PubSubEvent<'mouseOver' | 'my-event', { same: 'payload' }>
 type PubSubEvent<EventName extends string, Payload> = {
     [Key in EventName]: Payload;
 };
 
 // New `PubSubEvent`s should be added to the `EventMap`...
-type EventMap = PubSubEvent<'mouseover' | 'click', PointMouseEventData> &
-    PubSubEvent<'rangeselect', RangeMouseEventData> &
-    PubSubEvent<'rawdata', CommonEventData>;
+type EventMap = PubSubEvent<'mouseOver' | 'click', PointMouseEventData> &
+    PubSubEvent<'rangeSelect', RangeMouseEventData> &
+    PubSubEvent<'rawData', CommonEventData>;
 
 /**
  * Information of suggested genes.
@@ -114,10 +114,10 @@ export function createApi(
     return {
         subscribe: (type, callback) => {
             switch (type) {
-                case 'mouseover':
+                case 'mouseOver':
                 case 'click':
-                case 'rawdata':
-                case 'rangeselect':
+                case 'rawData':
+                case 'rangeSelect':
                     return PubSub.subscribe(type, callback);
                 default: {
                     console.error(`Event type not recognized, got ${JSON.stringify(type)}.`);
--- a/src/core/api.ts
+++ b/src/core/api.ts
@@ -1,43 +1,10 @@
 import * as PIXI from 'pixi.js';
-import type { Datum } from './gosling.schema';
 import type { HiGlassApi } from './higlass-component-wrapper';
 import type { HiGlassSpec } from './higlass.schema';
+import { subscribe, unsubscribe } from './pubsub';
 import { GET_CHROM_SIZES } from './utils/assembly';
 import type { CompleteThemeDeep } from './utils/theme';
 import { traverseViewsInViewConfig } from './utils/view-config';
-
-export interface CommonEventData {
-    /** Source visualization ID, i.e., `track.id` */
-    id: string;
-    /** Values in a JSON array that represent data after data transformation */
-    data: Datum[];
-}
-
-export interface PointMouseEventData extends CommonEventData {
-    /* A genomic coordinate, e.g., `chr1:100,000`. */
-    genomicPosition: string;
-}
-
-export interface RangeMouseEventData extends CommonEventData {
-    /* Start and end genomic coordinates, e.g., `chr1:100,000`. NULL if a range is deselected. */
-    genomicRange: [string, string] | null;
-}
-
-// Utility type for building strongly typed PubSub API.
-//
-// Add named events using a string union for `EventName`
-//
-// - Two different events ('mouseOver' & 'my-event') with the same payload
-//
-// PubSubEvent<'mouseOver' | 'my-event', { same: 'payload' }>
-type PubSubEvent<EventName extends string, Payload> = {
-    [Key in EventName]: Payload;
-};
-
-// New `PubSubEvent`s should be added to the `EventMap`...
-type EventMap = PubSubEvent<'mouseOver' | 'click', PointMouseEventData> &
-    PubSubEvent<'rangeSelect', RangeMouseEventData> &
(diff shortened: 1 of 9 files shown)


src/gosling-brush/linear-brush-model.ts
```diff
@@ -84,7 +84,7 @@ export class LinearBrushModel {
             d3Drag: hgLibraries.d3Drag
         };
 
-        this.style = Object.assign(BRUSH_STYLE_DEFAULT, style);
+        this.style = Object.assign({}, BRUSH_STYLE_DEFAULT, style);
```
sehilyi, line 87: To prevent from `BRUSH_STYLE_DEFAULT` being edited

src/core/gosling.schema.ts
```diff
@@ -250,17 +263,17 @@ interface SingleTrackBase extends CommonTrackDef {
...
-        hovering?: {
-            enableMultiHovering?: boolean;
+        // TODO: move all following style-related properties to `style` (June-02-2022)
+        mouseOveredMarks?: {
             showHoveringOnTheBack?: boolean;
         } & Partial<BrushAndMarkHighlightingStyle>;
-        selection?: { showOnTheBack?: boolean } & Partial<BrushAndMarkHighlightingStyle>;
-        brush?: Partial<BrushAndMarkHighlightingStyle>;
+        selectedMarks?: { showOnTheBack?: boolean } & Partial<BrushAndMarkHighlightingStyle>;
+        rangeSelectBrush?: Partial<BrushAndMarkHighlightingStyle>;
```
sehilyi, line 299: I will move these styling properties under `style` in a separate PR.

src/gosling-brush/linear-brush-model.ts
manzt: Ah good catch!

src/core/api.ts
```diff
@@ -27,17 +27,17 @@ export interface RangeMouseEventData extends CommonEventData {
...
     [Key in EventName]: Payload;
 };
 
 // New `PubSubEvent`s should be added to the `EventMap`...
-type EventMap = PubSubEvent<'mouseover' | 'click', PointMouseEventData> &
-    PubSubEvent<'rangeselect', RangeMouseEventData> &
-    PubSubEvent<'rawdata', CommonEventData>;
+type EventMap = PubSubEvent<'mouseOver' | 'click', PointMouseEventData> &
+    PubSubEvent<'rangeSelect', RangeMouseEventData> &
+    PubSubEvent<'rawData', CommonEventData>;
```
manzt: I'm starting to think this was more complex than it needed to be (my fault), and a simple interface would suffice:

```suggestion
type EventMap = {
  mouseOver: PointMouseEventData;
  click: PointMouseEventData;
  rangeSelect: RangeMouseEventData;
  rawData: CommonEventData;
}
```

src/core/gosling.schema.ts
```diff
@@ -240,6 +240,19 @@ export interface BrushAndMarkHighlightingStyle {
...
+    /* Turn on and off individual mouse events. */
+    [Event in MouseEvent]?: boolean;
+} & {
+    /* Group marks using keys in a data field. This affects how a set of marks are highlighted/selected by interaction. __Default__: `undefined` */
+    groupMarksByField?: string;
+
+    /* Determine whether all marks underneath the mouse point should be affected by mouse over. __Default__: `false` */
+    enableMouseOverOnMultipleMarks?: boolean;
+};
+
```
manzt: This looks good. My only concern is that there is an implicit link between the events/payloads in `api.ts` and`gosling.schema.ts`. The types are not shared or linked in the code at all. 

This means that anywhere we actually publish event data is not covered by our type-system,

```typescript
  PubSub.publish('rangeSelect', "I can put anything here and it's ok!");
```

The type checker does **not** ensure that either the event name _or_ payload are as defined in `api.ts`. Instead, we (manually) have to be extremely careful that whatever is in `api.ts` and our `PubSub.publish` use stay in sync.

One way around this would be to export a utility module that provides type-safe use of `PubSub.{publish,subscribe}`:

```typescript
// pubsub.ts
export function publish<
  Name extends keyof EventMap
>(name: Name, data: EventMap[Name]): void {
  PubSub.publish(name, data);
}

export function subscribe<
  Name extends keyof EventMap
>(name: Name, callback: (msg: string, data: EventMap[Name]) => void): void {
  PubSub.subscribe(name, callback);
}

```

manzt: Personally, I think it would make sense to define events in `gosling.schema.ts` and then import the `EventMap` in `pubsub.ts/api.ts`:


#### core/gosling.schema.ts

```typescript
export type _EventMap = { /* ... */ } // underscore for internal type

type MouseEvents = {
    [Event in keyof Omit<_EventMap, 'rawData'>]: boolean,

    /* Group marks using keys in a data field. This affects how a set of marks are highlighted/selected by interaction. __Default__: `undefined` */
    groupMarksByField?: string;

    /* Determine whether all marks underneath the mouse point should be affected by mouse over. __Default__: `false` */
    enableMouseOverOnMultipleMarks?: boolean;
}
```

#### core/pubsub.ts

```typescript
import type { _EventMap } from './gosling.schema.ts';

export function publish<
  Name extends keyof _EventMap
>(name: Name, data: EventMap[Name]): void {
  PubSub.publish(name, data);
}

export function subscribe<
  Name extends keyof _EventMap
>(name: Name, callback: (msg: string, data: EventMap[Name]) => void): void {
  PubSub.subscribe(name, callback);
}
```

#### core/api.ts

```typescript
import { subscribe } from './pubsub.ts';

export interface GoslingApi {
    subscribe: typeof subscribe;
    /* ... */
}

export function createApi(
    hg: HiGlassApi,
    hgSpec: HiGlassSpec | undefined,
    theme: Required<CompleteThemeDeep>
): GoslingApi {
    /* .. */
    return {
        subscribe,
        /* ... */
    }
}
```

manzt (approved):
Looks good. Just some comments regarding future-proofing the types for these events.

src/core/gosling.schema.ts
sehilyi: This is a really great idea! I updated all `publish`, `subscribe`, and `unsubscribe` in the [last commit](https://github.com/gosling-lang/gosling.js/pull/712/commits/f23297d3d076e64c705f4533da665e8aa666d860).

src/core/pubsub.ts
```diff
@@ -0,0 +1,16 @@
...
+}
+
+export function subscribe<Name extends keyof _EventMap>(
+    name: Name,
+    callback: (msg: string, data: _EventMap[Name]) => void
+): void {
+    PubSub.subscribe(name, callback);
+}
+
+export function unsubscribe<Name extends keyof _EventMap>(name: Name): void {
```
manzt: This can just be

```suggestion
export function unsubscribe(name: keyof _EventMap): void {
Write
```

since the generic type parameter does need to be used in the function signature.

manzt: Better yet, a local type alias

```typescript
type EventName = keyof _EventMap;
```

might be nice in this file.

sehilyi: I agree!

PR #716, opened 2022-06-04
fix: fix json track errors due to using undefined mRangeBrush

sehilyi:
I found that JSON tracks are making errors by using a class property `mRangeBrush` in `draw()` before defining it in the constructor, i.e., the parent class were calling `draw()` ([code](https://github.com/higlass/higlass/blob/387a03e877dcfa4c2cfeabc0869375b58c0b362d/app/scripts/TiledPixiTrack.js#L216)). 

All other non-JSON tracks seem to work okay since loading non-JSON data takes some times and the `draw()` function returns early.

Screen Shot 2022-06-03 at 17 05 33

```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'drawBrush')
    at GoslingTrackClass.draw (gosling-track.ts:265:30)
    at GoslingTrackClass.receivedTiles (hglib.js:12016:14)
    at higlass-raw-datafetcher.ts:137:17
```

### DIFF
--- a/src/gosling-track/gosling-track.ts
+++ b/src/gosling-track/gosling-track.ts
@@ -218,6 +218,9 @@ function GoslingTrack(HGC: any, ...args: any[]): any {
 
         /* ----------------------------------- RENDERING CYCLE ----------------------------------- */
 
+        // !! Be aware that this function is called in the middle of `constructor()` by a parent class (i.e., `super(...)`).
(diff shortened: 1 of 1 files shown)


manzt (approved):
Oh good catch!  Is it possible to assign `this.mRangeBrush` prior to calling `super()` in the constructor?

sehilyi:
It's not possible. If I move the `this.mRangeBrush` before calling `super()`, a pop-up says, `'super' must be called before accessing 'this' in the constructor of a derived class.ts (17009)`

Issue #711, opened 2022-06-01
Refine the structure of rangeSelect API data


sehilyi:
I wonder if we could do something similar to what you had before then,

```typescript
type GenomicRegion = [chr: string, start: number, end: number];

export interface RangeMouseEventData extends CommonEventData {
  interval: [GenomicRegion, GenomicRegion],
}
```

I guess users have enough information to understand which chromosomes are within this `interval` from the assembly (chromsizes).

_Originally posted by @manzt in https://github.com/gosling-lang/gosling.js/pull/693#discussion_r887231280_

PR #717, opened 2022-06-04
refactor: move mouse event styles under the style property

### LINKED ISSUE #710: Finalize property names and defaults for mouse events

sehilyi:
This PR moves all styling properties for mouse events (i.e., mouse over and selection effects and the range brush) under the `style` property. Users can specify this at the track level

```ts
{
  mark: 'point', ..., // track definition
  style: {
    mouseOver: {
      color: 'blue',
      opacity: 0.5,
      strokeWidth: 0,
      arrange: 'behind'
    },
    select: {
      color: 'red',
      opacity: 0.5,
      arrange: 'behind'
    },
    brush: {
      color: 'purple',
      stroke: 'purple'
    }
  }
}
```

or at the any upper level

```ts
{
  title: 'gos visualization', ..., // root-level gosling spec
  style: {
    mouseOver: {
      color: 'blue',
      opacity: 0.5,
      strokeWidth: 0,
      arrange: 'behind'
    },
    select: {
      color: 'red',
      opacity: 0.5,
      arrange: 'behind'
    },
    brush: {
      color: 'purple',
      stroke: 'purple'
    },
    arrange: 'behind'
  },
  views: [...] // all children use the same style unless they are defined at lower levels.
}
```

Fix #710

### DIFF
--- a/src/core/gosling.schema.ts
+++ b/src/core/gosling.schema.ts
@@ -138,6 +138,14 @@ export interface CommonViewDef {
      */
     centerRadius?: number;
 
+    // Experimental
+    experimental?: {
+        /*
+         * Determine whether to use mouse events, such as click and mouse over on marks. __Default__: `false`
+         */
+        mouseEvents?: boolean | MouseEventsDeep;
+    };
+
     /**
      * Define the [style](http://gosling-lang.org/docs/visual-channel#style-related-properties) of multive views.
      * Will be overriden by the style of children elements (e.g., view, track).
@@ -268,14 +276,6 @@ export type MouseEventsDeep = {
 /* ----------------------------- TRACK ----------------------------- */
 export type SingleTrack = SingleTrackBase & Encoding;
 
-export interface BrushAndMarkHighlightingStyle {
-    color: string;
-    stroke: string;
-    strokeWidth: number;
-    strokeOpacity: number;
-    opacity: number;
-}
-
 interface SingleTrackBase extends CommonTrackDef {
     // Data
     data: DataDeep;
@@ -285,19 +285,6 @@ interface SingleTrackBase extends CommonTrackDef {
 
     // Tooltip
     tooltip?: Tooltip[];
-
-    // Experimental
-    experimental?: {
-        // Mouse events
-        mouseEvents?: boolean | MouseEventsDeep;
-
-        // TODO: move all following style-related properties to `style` (June-02-2022)
-        mouseOveredMarks?: {
-            showHoveringOnTheBack?: boolean;
-        } & Partial<BrushAndMarkHighlightingStyle>;
-        selectedMarks?: { showOnTheBack?: boolean } & Partial<BrushAndMarkHighlightingStyle>;
-        rangeSelectBrush?: Partial<BrushAndMarkHighlightingStyle>;
-    };
 
     // Mark
     mark: Mark;
@@ -365,6 +352,14 @@ export type OverlaidTrack = Partial<SingleTrack> &
         // This is a property internally used when compiling
         overlay: Partial<Omit<SingleTrack, 'height' | 'width' | 'layout' | 'title' | 'subtitle'>>[];
     };
+
+export interface BrushAndMarkHighlightingStyle {
+    color?: string;
+    stroke?: string;
+    strokeWidth?: number;
+    strokeOpacity?: number;
+    opacity?: number;
+}
 
 export interface Style {
     // Top-level Styles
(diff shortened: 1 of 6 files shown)


src/core/gosling.schema.ts
```diff
@@ -460,6 +455,21 @@ export interface Style {
...
+
+    /**
+     * Customize visual effects selection events on marks with range brushes.
+     */
+    selectedMarks?: { showOnTheBack?: boolean } & BrushAndMarkHighlightingStyle;
+
+    /**
+     * Customize the style of range brushes.
+     */
+    rangeSelectBrush?: BrushAndMarkHighlightingStyle;
```
manzt: The naming is a little confusing to me. What about `markMouseOver`, `markSelected`, `brush`? What does `showOnTheBack` mean?

src/core/gosling.schema.ts
```diff
@@ -366,6 +353,14 @@ export type OverlaidTrack = Partial<SingleTrack> &
         overlay: Partial<Omit<SingleTrack, 'height' | 'width' | 'layout' | 'title' | 'subtitle'>>[];
     };
 
+export interface BrushAndMarkHighlightingStyle {
```
manzt: `SelectionStyle` or `EventStyle`? Just trying to think of something more general for naming. If we ever had a select that was different then we'd have to remain this `BrushAndMarkHighlightingAndXXXXStyle`. 

It's generally preferred to define a composible type and reuse accordingly. See example  https://mobile.twitter.com/Swizec/status/1532837585944334336

src/core/gosling.schema.ts
manzt: I think of each key here as `style` for `key-name`. The styles are applied to each individual mark, so we can have a singular name.

manzt (approved):
Looks good! Just some minor comments on naming and type semantics.

src/core/gosling.schema.ts
sehilyi: Determining variable names is always tricky for me, so these comments are really helpful!

I like the proposed ones. Do you think we can further remove `mark` (e.g., `mouseOver`, `select`, `brush`)? Given that we only support mouse over and selection effects on marks, removing the term would not be confusing to users.

`showOnTheBack` determines whether to show the mouse hovering/selection effects on the back of marks (e.g., as a background that does not occlude other marks).

src/core/gosling.schema.ts
sehilyi: I agree. I will use `EventStyle` for now but will think about defining a common type and reusing them across the schema. I think similar properties are defined and used redundantly for `text` styling.

src/core/gosling.schema.ts
manzt: > I like the proposed ones. Do you think we can further remove mark (e.g., mouseOver, select, brush)? Given that we only support mouse over and selection effects on marks, removing the term would not be confusing to users.

Sounds good to me!

> `showOnTheBack` determines whether to show the mouse hovering/selection effects on the back of marks (e.g., as a background that does not occlude other marks).

Maybe something like `arrangement: 'behind'` or `order: 'behind'` or `display: 'behind'` or `applyEffect: 'behind'`?

sehilyi: Good idea! I will go with `arrangement: 'behind' | 'front'`. Realized that Keynote and other applications use similar terms `"Arrange: Move To Back/Front"`.

src/core/gosling.schema.ts
```diff
@@ -366,6 +353,25 @@ export type OverlaidTrack = Partial<SingleTrack> &
...
+}
+
+/*
+ * Show event effects behind or in front of marks.
+ * __Default__: `'front'`
+ */
+export interface EventArrange {
+    arrange?: 'behind' | 'front';
+}
+
```
manzt, line 374: Could just unify in a single type,

```typescript
export interface EventStyle {
    color?: string;
    stroke?: string;
    strokeWidth?: number;
    strokeOpacity?: number;
    opacity?: number;
    arrange?: 'behind' | 'front',
}
```

And then later,

```typescript
  brush: Omit<EventStyle, 'arrange'>,
```

for interface which don't have `arrange`

manzt: oh oops, just seeing you merged. no worries!

src/core/gosling.schema.ts
manzt: Arrange is great!

src/core/gosling.schema.ts
sehilyi: Oh, actually this is better.

sehilyi: Let me make this change when open an PR for #711

PR #719, opened 2022-06-07
refactor: give chromosome and position info. separately to API consumers

### LINKED ISSUE #711: Refine the structure of rangeSelect API data

sehilyi:
This PR changes the format of genomic position or range that are exposed to the consumers of mouse event APIs (i.e., 'rangeSelect', 'mouseOver', 'click'). Previously, we used string values (i.e., `chr1:10,000`), and this PR separates the chromosome from the position:

```ts
// Event data for `click` or 'mouseOver' that is exposed to API consumers
genomicPosition: { chromosome: 'chr1', position: 10000 },
...
```

For the `rangeSelect` event:

```ts
// Event data for `rangeSelect` that is exposed to API consumers
genomicRange: [
  { chromosome: 'chr1', position: 10000 },
  { chromosome: 'chr11', position: 2000 }
],
...
```

- refactor: use object instead of string for genomic range and position
- test: add a test for abs to rel genomic position

Fix #711
Fix https://github.com/gosling-lang/gosling.js/pull/717#discussion_r891411407

### DIFF
--- a/src/core/utils/assembly.ts
+++ b/src/core/utils/assembly.ts
@@ -17,22 +17,20 @@ export interface ChromSize {
 }
 
 /**
- * Get relative chromosome position (e.g., `100` => `chr:100`)
+ * Get relative chromosome position (e.g., `100` => `{ chromosome: 'chr1', position: 100 }`)
  */
 export function getRelativeGenomicPosition(absPos: number, assembly?: string): GenomicPosition {
-    const chrAndRange = Object.entries(GET_CHROM_SIZES(assembly).interval).find(d => {
-        const [, [start, end]] = d;
+    const [chromosome, absInterval] = Object.entries(GET_CHROM_SIZES(assembly).interval).find(d => {
+        const [start, end] = d[1];
         return start <= absPos && absPos < end;
-    });
+    }) ?? [null, null];
 
-    if (!chrAndRange) {
+    if (!chromosome || !absInterval) {
         // The number is out of range
         return { chromosome: 'unknown', position: absPos };
     }
 
-    const chromosome = chrAndRange[0];
-    const position = absPos - chrAndRange[1][0];
-    return { chromosome, position };
+    return { chromosome, position: absPos - absInterval[0] };
 }
 
 /**
--- a/src/core/utils/assembly.ts
+++ b/src/core/utils/assembly.ts
@@ -1,4 +1,4 @@
-import { format } from 'd3-format';
+import type { GenomicPosition } from '@gosling.schema';
 import {
     CHROM_SIZE_HG16,
     CHROM_SIZE_HG17,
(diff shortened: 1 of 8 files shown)


src/core/utils/assembly.ts
```diff
@@ -19,19 +19,20 @@ export interface ChromSize {
 /**
  * Get relative chromosome position (e.g., `100` => `chr:100`)
  */
-export function getRelativeGenomicPosition(absPos: number, assembly?: string): string {
+export function getRelativeGenomicPosition(absPos: number, assembly?: string): GenomicPosition {
     const chrAndRange = Object.entries(GET_CHROM_SIZES(assembly).interval).find(d => {
         const [, [start, end]] = d;
```
manzt:
```suggestion
        const [start, end] = d[1];
```

src/core/utils/assembly.ts
```diff
@@ -19,19 +19,20 @@ export interface ChromSize {
 /**
  * Get relative chromosome position (e.g., `100` => `chr:100`)
  */
-export function getRelativeGenomicPosition(absPos: number, assembly?: string): string {
+export function getRelativeGenomicPosition(absPos: number, assembly?: string): GenomicPosition {
     const chrAndRange = Object.entries(GET_CHROM_SIZES(assembly).interval).find(d => {
```
manzt: Maybe destructure here, rather than below:

```javascript
    const [chromosome, absInterval] = Object.entries(GET_CHROM_SIZES(assembly).interval).find(d => {
```

src/core/utils/assembly.ts
```diff
@@ -19,19 +19,20 @@ export interface ChromSize {
...
         const [, [start, end]] = d;
         return start <= absPos && absPos < end;
     });
 
     if (!chrAndRange) {
         // The number is out of range
-        return `${absPos}`;
+        return { chromosome: 'unknown', position: absPos };
```
manzt, line 30: Maybe add a unit test for this branch?

manzt (approved):
Looks good to me!

src/core/utils/assembly.ts
sehilyi: Good idea!

src/core/utils/assembly.ts
sehilyi: I like this way

src/core/utils/assembly.ts
sehilyi: 👍

PR #721, opened 2022-06-07
feat(editor): compact side panel buttons

### LINKED ISSUE #683: Make editor buttons compact
Screen Shot 2022-05-10 at 15 44 58

The menu now became very long...

wangqianwen0418:
fix #683
image image

### DIFF
--- a/editor/editor.css
+++ b/editor/editor.css
@@ -133,7 +133,7 @@
 
 .side-subpanel {
     padding-top: 5px;
-    background-color: #F1F1F1;
+    background-color: #f1f1f1;
     border-bottom: lightgray solid 1px;
     border-top: lightgray solid 1px;
 }
@@ -160,12 +160,11 @@
 .side-subpanel .side-subpanel-button {
     margin-top: 10px;
     margin-bottom: 10px;
-    font-size: 6px;
 }
 
 .side-panel .side-panel-button-not-active,
 .side-subpanel .side-subpanel-button-not-active {
-    color: lightgray;
+    color: #aaa;
     cursor: not-allowed;
 }
 
--- a/editor/editor.css
+++ b/editor/editor.css
@@ -133,9 +133,9 @@
 
 .side-subpanel {
     padding-top: 5px;
-    background-color: #fff;
-    border-bottom: gray solid 1px;
-    border-top: gray solid 1px;
+    background-color: #F1F1F1;
+    border-bottom: lightgray solid 1px;
+    border-top: lightgray solid 1px;
 }
 
 .side-panel .side-panel-button,
--- a/editor/editor.css
+++ b/editor/editor.css
@@ -131,7 +131,15 @@
     height: 100%;
 }
 
-.side-panel .side-panel-button {
+.side-subpanel {
+    padding-top: 5px;
+    background-color: #fff;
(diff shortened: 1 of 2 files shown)


editor/editor.css
```diff
@@ -131,7 +131,15 @@
     height: 100%;
 }
 
-.side-panel .side-panel-button {
+.side-subpanel {
+    padding-top: 5px;
+    background-color: #fff;
```
sehilyi: Maybe, use a darker color instead? I feel using a darker color instead of white is more natural for deeper hidden menus. But, this might be just a personal preference.

```suggestion
    background-color: #F1F1F1;
```

**Screenshot:**
Screen Shot 2022-06-07 at 17 49 38

editor/editor.css
```diff
@@ -131,7 +131,15 @@
...
+    border-bottom: gray solid 1px;
+    border-top: gray solid 1px;
```
sehilyi: We can use the `lightgrey` color, consistent with most borders in the editor.

```suggestion
    border-bottom: lightgray solid 1px;
    border-top: lightgray solid 1px;
```

editor/editor.css
```diff
@@ -149,11 +157,26 @@
     /* justify-content: center; */
 }
 
-.side-panel .side-panel-button-not-active {
+.side-subpanel .side-subpanel-button {
+    margin-top: 10px;
+    margin-bottom: 10px;
+    font-size: 6px;
```
sehilyi: For the readability, it might be better to use bigger size icons and labels for subpanel buttons. We can even use the same size as the upper-level icons/labels. I think `9px` for regular buttons is already small enough, so we may not want to use smaller fonts.

editor/editor.tsx
```diff
@@ -800,30 +800,38 @@ function Editor(props: RouteComponentProps) {
...
-                            className="side-panel-button"
-                            onClick={() => {
-                                setIsfontZoomOut(!isFontZoomOut);
-                            }}
-                        >
-                            {getIconSVG(ICONS.TEXT, 15, 15)}
-                            -
-                            <br />
-                            SMALLER
+                            FONT
```
sehilyi: Users might think they can change font styles. How about `FONT SIZE` if this fit into the narrow side panel.

sehilyi (approved):
Looks good! I left several comments which should be easy to address.

In terms of a way to open and close sub-panels, I wonder if it makes sense to use click instead of mouse over given that the side panel is fairly narrow (i.e., clicking once opens the sub-panel and clicking again closes it). As we are considering adding more sub-panel buttons, this approach might be easier to use. With the current approach, users will have to be careful about moving the cursor not to close the menu.

wangqianwen0418:
Thanks for the feedback! will update accordingly soon

wangqianwen0418:
I will keep "hovering" as the current way to open a sub panel, considering that "hovering" is the common way to explore a submenu. People usually will not click a button unless they are very certain, therefore they are very likely to miss the functions in the sub panel.
@sehilyi , does this make sense to you?

sehilyi:
Yes, that works for me, too. We can always change the approach as we add more sub-panel menus. For now, I think both approaches work.

Issue #340, opened 2021-04-07
Preserve view specification in session storage 


flekschas:
I noticed when reloading the editor, the previous view specification is gone. It would be great to preserve the current view specification in session storage, so when I (as a user) reload the page, I will load the view specification I've previously worked on.

The nice thing about `sessionStorage` is that it's **not** preserved between tabs. So if I wanted to work on two view specifications at the same time I can do this without having to worry about weird caching issues.

sehilyi:
Partly supported in #722. We can additionally support `language` and store two specs separately, i.e., JSON and JS codes.  See below:

---

I am wondering will it make sense to store `language` `jscode` and `code` rather than just the `spec`.
Currently, the gosling spec is saved in session storage, which might lead to the following issues:
- the comments in the JSON editor will be removed
- the structured jscode will be changed to `var spec={code}` even though the users only edit in the JSON editor
- even though users edit in the js editor, they will be directed to the JSON editor when refreshing

_Originally posted by @wangqianwen0418 in https://github.com/gosling-lang/gosling.js/pull/722#pullrequestreview-1000237500_

PR #722, opened 2022-06-07
feat(editor): preserve compiled JSON spec in session storage

sehilyi:
This preserves a compiled JSON spec in `sessionStorage` (i.e., keeping spec in each browser tab). 
- The spec will be preserved after you refresh the page (or go to other sites and come back to the editor)
- When a browser tab is closed, the latest spec can be reused by opening the closed tab (e.g., in Chrome, Right Click → "Reopen Closed Tab").

This stores a valid Gosling spec that is compiled lastly, i.e., when you edit on the JS editor, the editor stores the parsed JSON spec.

Toward #340

### DIFF
--- a/editor/editor.tsx
+++ b/editor/editor.tsx
@@ -62,6 +62,9 @@ const LIMIT_CLIPBOARD_LEN = 4096;
 const EDITOR_HEADER_HEIGHT = 40;
 const BOTTOM_PANEL_HEADER_HEIGHT = 30;
 
+// A key to store and get a Gosling spec via sessionStorage
+const SESSION_KEY_SPEC = 'session-gosling-spec';
+
 export const GoslingLogoSVG = (width: number, height: number) => (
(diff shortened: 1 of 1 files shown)


wangqianwen0418 (commented):
Hi @sehilyi, thanks for this cool update! It looks good to me.

I am wondering will it make sense to store `language` `jscode` and `code` rather than just the `spec`.
Currently, the gosling spec is saved in session storage, which might lead to the following issues:
- the comments in the JSON editor will be removed
- the structured jscode will be changed to `var spec={code}` even though the users only edit in the JSON editor
- even though users edit in the js editor, they will be directed to the JSON editor when refreshing

sehilyi:
Thanks for the review! Yes, I agree with all issues, and I was actually thinking the same about storing `language` as well as the two raw specs (instead of the compiled Gosling spec).

I can open a separate PR for the improvements when I get time. I think having this without those improvements can be still helpful since users can at least save their previous compiled spec that could be otherwise entirely gone.

(Edited the title and the original description a bit to clarify that this PR saves a compiled JSON spec only).

PR #728, opened 2022-06-21
feat: add partial types for `HGC`

### LINKED ISSUE #707: Add partial types for `HGC`
Towards #702. Rather than typing `any` for `HGC` we could start to type out the parts of the interface which rely on.

-------------- 

At a minimum, we could start to type out `HGC` in a followup PR.

https://github.com/higlass/higlass/blob/1fcadc24a9915960ee056e5fbd7d1097220e8d0e/app/scripts/configs/available-for-plugins.js#L99-L117

```typescript
// src/types/index.d.ts
declare module '@higlass/available-to-plugins' {
  export interface Libraries {
    d3Array: import('d3-array'),
    d3Axis: import('d3-axis'),
    d3Brush: import('d3-brush'),
    d3Color: import('d3-color'),
    d3Drag: import('d3-drag'),
    d3Dsv: import('d3-dsv'),
    // ...
  }
}
```

```
    constructor(
        selection: d3Selection.Selection<SVGGElement, unknown, HTMLElement, unknown>,
        higlassLibraries: import('@higlass/avaiable-for-plugins').Libraries,
    )
```

_Originally posted by @manzt in https://github.com/gosling-lang/gosling.js/pull/693#discussion_r883067202_

manzt:
Fixes #707 

Replaces all instances of `HGC: any` with a more strictly typed interface. There are many apparent instances of bugs that are introduced by `any`, and this PR does not address all of them. Typing out all of `HGC` would be extremely time consuming, so this PR was an effort to provide the minimal types necessary for what we use from `HGC.utils`, `HGC.services`, and `HGC.libraries`. 

 follow up PR will be necessary to type `HGC.tracks.BarTrack` and `HGC.tracks.PixiTrack` since these interfaces are much more complicated and use some level of inheritance (read: it would be easier to spend some time on higlass to export these types rather than copying the interfaces ourselves).


src/core/mark/legend.ts
```diff
@@ -46,7 +49,7 @@ export function drawColorLegend(
 }
 
 export function drawColorLegendQuantitative(
-    HGC: any,
+    HGC: { libraries: Libraries },
     trackInfo: any,
     tile: any,
```
manzt, line 54: we should have these types somewhere in the library.

src/core/utils/text-style.ts
```diff
@@ -16,13 +18,13 @@ export const DEFAULT_TEXT_STYLE: TextStyle = {
...
 };
 
-export function getTextStyle(style: TextStyle = DEFAULT_TEXT_STYLE) {
+export function getTextStyle(style: TextStyle = DEFAULT_TEXT_STYLE): Partial<PIXI.ITextStyle> {
+    const fontWeight = style.fontWeight ?? DEFAULT_TEXT_STYLE.fontWeight;
     return {
         fontSize: `${style.size ?? DEFAULT_TEXT_STYLE.size}px`,
         fontFamily: style.fontFamily ?? DEFAULT_TEXT_STYLE.fontFamily,
-        fontWeight: style.fontWeight ?? DEFAULT_TEXT_STYLE.fontWeight,
+        fontWeight: fontWeight === 'light' ? 'lighter' : fontWeight,
```
manzt, line 26: This was a bug apparently, or it's a bug with the type definitions for `PIXI.ITextStyle` which doesn't include `light` and only `lighter`.

src/gosling-genomic-axis/axis-track.ts
```diff
@@ -17,15 +17,22 @@ const TICK_COLOR = 0x777777;
...
     }
 
-    const { absToChr, colorToHex, pixiTextToSvg, showMousePosition, svgLine } = HGC.utils;
+    const {
+        absToChr,
+        colorToHex,
+        pixiTextToSvg,
+        svgLine,
+        // @ts-expect-error did you mean `setupShowMousePosition`?
+        showMousePosition
```
manzt: This is probably the biggest bug I found. We use this function twice but with the `setupShowMousePosition` arguments instead. Not sure what the desired behavior is.

src/core/mark/title.ts
```diff
@@ -85,6 +85,7 @@ export function drawCircularTitle(
     g.closePath();
 
     // Render a label
+    // @ts-expect-error, missing argument to updateText
```
manzt, line 89: type error now that `textGraphic` isn't `any`

src/core/mark/title.ts
```diff
@@ -67,7 +67,7 @@ export function drawCircularTitle(
     const scaledEndX = padding + txtWidth;
 
     // Determine the points of a rope element for a lebel
-    const ropePoints: number[] = [];
+    const ropePoints: import('pixi.js').Point[] = [];
```
manzt: incorrect type

src/gosling-genomic-axis/axis-track.ts
```diff
@@ -172,6 +179,7 @@ function AxisTrack(HGC: any, ...args: any[]): any {
                 const text = new HGC.libraries.PIXI.Text(chromName, this.pixiTextConfig);
 
                 // give each string a random hash so that some get hidden when there's overlaps
+                // @ts-expect-error invalid property?
                 text.hashValue = Math.random();
```
manzt: property not found on `PIXI.Text`?

src/gosling-genomic-axis/axis-track.ts
```diff
@@ -566,6 +574,7 @@ function AxisTrack(HGC: any, ...args: any[]): any {
                 this.allTexts.push({
                     importance: chrText.hashValue,
                     text: chrText,
+                    // @ts-expect-error possibly null or undefined and should just be SimpleRope
```
manzt: Confusing logic here. `rope` can be `SimpleRope | null | undefined` and was assumed to be `SimpleRope` since `rope` was `any` previously.

src/missing-types.d.ts
```diff
@@ -0,0 +1,136 @@
...
+        // fetchTilesDebounced();
+        // json();
+        // text();
+        // tileDataToPixData();
+    };
+
+    type Tracks = {
+        BarTrack: Track; // TODO(06-21-22)
+        PixiTrack: Track;
+    };
```
manzt: For a follow-up PR. We should really try to have the types for these tracks minimally.

src/core/mark/index.ts
```diff
@@ -63,7 +68,7 @@ export function drawMark(HGC: any, trackInfo: any, tile: any, model: GoslingTrac
     // console.log(trackInfo._xScale.domain(), trackInfo._yScale.domain(), (model.getChannelScale('x') as any)?.domain(), (model.getChannelScale('y') as any)?.domain());
 
     // Replace the scale of a genomic axis with the one that is generated by the HiGlass data fetcher.
-    ['x', 'x1', 'x1e', 'xe'].forEach((d: any) => {
+    (['x', 'x1', 'x1e', 'xe'] as const).forEach(d => {
```
sehilyi, line 71: Ah, good to know this works.

src/core/mark/legend.ts
```diff
@@ -6,10 +6,13 @@ import type { Dimension } from '../utils/position';
 import { scaleLinear, ScaleLinear } from 'd3-scale';
 import { getTextStyle } from '../utils/text-style';
 
+// Just the libraries necesssary fro this module
+type Libraries = Pick<import('@higlass/available-for-plugins').Libraries, 'PIXI' | 'd3Selection' | 'd3Drag'>;
```
sehilyi: Just to learn – I assume this is necessary to reduce the size of the module?

src/core/mark/title.ts
sehilyi: Just to learn – is there a difference between this and importing the type at the start of the file? (e.g., `import type { Point } from 'pixi.js';`)

src/core/utils/text-style.ts
sehilyi: Ah, thanks for the correction. I can confirm the 'light' is not listed in the PIXI documentation (at least in the current version).

src/gosling-genomic-axis/axis-track.ts
sehilyi: I think so, too. Maybe we can make the type of `text` the union of PIXI.Text and `hashValue: number`.

src/gosling-genomic-axis/axis-track.ts
```diff
@@ -20,12 +20,19 @@ type TickLabelInfo = {
     rope: PIXI.SimpleRope;
```
sehilyi: I think I had to make this optional:

```suggestion
    rope?: PIXI.SimpleRope;
```

src/gosling-genomic-axis/axis-track.ts
sehilyi: The `rope` is used only when the layout is circular, and it can be `undefined` by the if statement on line 549:

```
if (circular) {
    rope = this.addCurvedText(chrText, viewportMidX);
```

src/gosling-genomic-axis/axis-track.ts
sehilyi: Related to my comment below.

src/missing-types.d.ts
```diff
@@ -0,0 +1,136 @@
...
+
+    type Track = any;
+
+    type Utils = {
+        setupShowMousePosition(context: Track, is2d?: boolean, isGlobal?: boolean): void;
```
sehilyi: I see it is `showMousePosition` in the higlass repo. Maybe this is a typo or I am referring to a wrong file?

https://github.com/higlass/higlass/blob/54f5aae61d3474f9e868621228270f0c90ef9343/app/scripts/utils/index.js#L66

sehilyi (commented):
Thanks a lot for putting in this effort. So helpful update for maintenance (I can already see many bugs we had!).

It looks all good to me except `showMousePosition` (see comment).

I can address several bugs you identified later.

src/core/mark/title.ts
manzt: No there is no difference! It would be more stylistic I guess. Would you have a preference? Another Option which I like is:

```typescript
import type * as PIXI from "pixi.js";

const ropePoints: PIXI.Point[] = [];
```

manzt: One thing I couldn't seem to figure out was how to type `@higlass/available-for-plugins` so that we could import it like this....


```typescript
import type * as HGC from '@higlass/available-for-plugins';
```

src/core/mark/legend.ts
manzt: This is done for testing, and to avoid uses of `any` or `@ts-ignore` in our tests. 

Now that `HGC` is fully typed, our tests have type errors:

```typescript
drawMark({ libraries: { PIXI: /* ... */ } }, ...args)
                   // ^ type error, not a complete HGC type.
```

So I narrowed the type to just the features which are used for the module so that it is easier to test.

src/core/utils/text-style.ts
manzt: Are you ok with this approach? This way we keep our API consistent but adapt it sufficiently for PIXI.

src/missing-types.d.ts
manzt: Ah I see... the default function exported from `show-mouse-position.js` is actually `setupShowMousePosition`: https://github.com/higlass/higlass/blob/54f5aae61d3474f9e868621228270f0c90ef9343/app/scripts/utils/show-mouse-position.js#L145

but it is renamed in index.js

manzt: side note: I'm really happy i found `@ts-expect-error` because it leads to a compile error when they are no longer necessary and can be removed. Changed `setupShowMousePosition` to `showMousePosition` and...

```
src/gosling-genomic-axis/axis-track.ts:33:9 - error TS2578: Unused '@ts-expect-error' directive.

33         // @ts-expect-error did you mean `setupShowMousePosition`?
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

src/gosling-track/gosling-track.ts:82:5 - error TS2578: Unused '@ts-expect-error' directive.

82     // @ts-expect-error, should be setupShowMousePosition?
       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


```

src/gosling-genomic-axis/axis-track.ts
manzt: It seems like this can also be `null` though, since `this.addCurvedText(...)` can return `null`:


```typescript
        addCurvedText(textObj: any, cx: number) {
// ...
             if (ropePoints.length === 0) {
                return null;
            }

```

src/core/mark/legend.ts
sehilyi: Ah, gotcha!

src/missing-types.d.ts
sehilyi: Oh, that's a surprise! I thought it will just not complain about the correct type.

src/core/utils/text-style.ts
sehilyi: Yes, I am okay with it.

I am thinking that we can later better unify the `TextStyle` with the `PIXI.ITextStyle`.

src/gosling-genomic-axis/axis-track.ts
manzt: So the type would need to be:

```typescript
    rope?: PIXI.SimpleRope | null;
```

 to allow this code to type-check completely. Allowing a field to be `null` | `undefined` is a pattern I really try to avoid since it leads to complicated interfaces like this, and semantically `null` and `undefined` usually mean the same thing. My personal preference is to avoid `null` in JavaScript, since we have also have `undefined` (nice post here on topic https://github.com/sindresorhus/meta/discussions/7), but I'm fine with choosing one or the other.

src/core/mark/title.ts
sehilyi: No strong preference. I think only when certain types are used multiple times in the same file, we can import them at the start of the file and reuse them.

src/gosling-genomic-axis/axis-track.ts
manzt: I added

```typescript
const text = Object.assign(
    new HGC.libraries.PIXI.Text(chromName, this.pixiTextConfig),
    { hashValue: Math.random() } ,
);
```

which auto infers the type `PIXI.Text & { hashValue: Math.random() }`

src/gosling-genomic-axis/axis-track.ts
sehilyi: I agree. I am also trying to avoid using `null`. We can return `undefined` instead of adding `| null` to the type:

```ts
if (ropePoints.length === 0) {
                return undefined;
            }
```

src/missing-types.d.ts
manzt: Yes! I'd used `// @ts-ignore` or `as any` to turn off the type checker previously, but those strategies are permanently hidden from the compiler. I much prefer `// @ts-expect-error` for the (hopefully) rare cases when this is necessary, but I think it will very much come in handy as we continue to remove `any` from Gosling codebase.

src/gosling-genomic-axis/axis-track.ts
manzt: Ok great!

src/core/mark/title.ts
manzt: I think in TS code, we should prefer:

```typescript
import type { ... } from "...";
```

compared to inline type imports with `import(...)`. I changed this in the source code.

However, I cannot figure out how to get:

```typescript
import type * as HGC from '@higlass/available-for-plugins';
```

to work correctly, so I have left the inline imports for these expressions.

manzt:
I reorganized the types in `missing-types.d.ts` into seperate "virtual" modules. I think it should be easier to maintian this way.

```typescript
declare module '@higlass/available-for-plugins' {
    export * as libraries from '@higlass/libraries';
    export * as services from '@higlass/services';
    export * as tracks from '@higlass/tracks';
    export * as utils from '@higlass/utils';
}
declare module '@higlass/libraries' { /* ... */ }
declare module '@higlass/services' { /* ... */ } 
declare module '@higlass/tracks' { /* ... */ }
declare module '@higlass/utils' { /* ... */ }
```

manzt:
I think this PR is ready to merge if you do

src/core/mark/title.ts
manzt: Ah actually this works correctly. We need to access the _namespace_ as a type with `typeof`:

```typescript
import type * as _HGC from '@higlass/available-for-plugins';

function test(HGC: typeof _HGC) { / * ... * / }
```

PR #732, opened 2022-06-24
add types for `@higlass/tracks`

manzt:
This PR:

- Replaces `typeof import('@higlass/avaiable-for-plugins')` with `import('@higlass/types').HGC`
- Add types for `HGC.tracks.PixiTrack` and `HGC.tracks.Track`
- Adds `definePluginTrack` utility

```typescript
export default definePluginTrack({
  type: "my-track",
  defaultOptions: {
     foo: "bar",
  }
}, (HGC, context, options) => {
  
  HGC // { libraries: ..., tracks: ..., ... }

  context // { id: string, ... } from https://github.com/higlass/higlass/blob/0b2cac5a770db6d55370a61d5dbbe09c4f577c68/app/scripts/TrackRenderer.js#L1507-L1536

  options // inferred from `defaultOptions` above
  options.foo // string
  
  class MyTrack extends HGC.tracks.PixiTrack { /* ... */ };

  return new MyTrack();
})
```

### DIFF
--- a/src/core/api.ts
+++ b/src/core/api.ts
@@ -126,7 +126,7 @@ export function createApi(
                 const a = document.createElement('a');
                 document.body.append(a);
                 a.download = 'gosling-visualization';
-                a.href = URL.createObjectURL(blob);
+                a.href = URL.createObjectURL(blob!);
                 a.click();
                 a.remove();
             }, 'image/png');
--- a/src/gosling-genomic-axis/axis-track.ts
+++ b/src/gosling-genomic-axis/axis-track.ts
@@ -8,35 +8,134 @@ import { format, precisionPrefix, formatPrefix } from 'd3-format';
 import { GET_CHROM_SIZES } from '../core/utils/assembly';
 import { cartesianToPolar } from '../core/utils/polar';
 import { getTextStyle } from '../core/utils/text-style';
+import { definePluginTrack } from '../core/utils/define-plugin-track';
+
+import type { TextStyle } from '../core/utils/text-style';
+import type { PluginTrackFactory, TrackConfig } from '../core/utils/define-plugin-track';
 
 const TICK_WIDTH = 200;
 const TICK_HEIGHT = 6;
 const TICK_TEXT_SEPARATION = 2;
 const TICK_COLOR = 0x777777;
 
+type AxisTrackOptions = {
+    innerRadius: number;
+    outerRadius: number;
+    startAngle: number;
+    endAngle: number;
+    width: number;
+    height: number;
+    layout: 'linear' | 'circular';
+    labelPosition: string;
+    labelColor: string;
+    labelTextOpacity: number;
+    trackBorderWidth: number;
+    trackBorderColor: string;
+    tickPositions: 'even' | 'ends';
+    fontSize: number;
+    fontFamily: string; // 'Arial',
+    fontWeight: NonNullable<TextStyle['fontWeight']>;
+    color: string;
+    stroke: string;
+    backgroundColor: string;
+    showMousePosition: boolean;
+    tickColor: number;
+    tickFormat?: string;
+    assembly?: string;
+    reverseOrientation?: boolean;
+};
 type TickLabelInfo = {
     importance: number;
     text: PIXI.Text;
     rope?: PIXI.SimpleRope;
 };
+type TickLine = [number, number, number, number];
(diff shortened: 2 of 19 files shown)


manzt:
Should be ready to review!

src/core/api.ts
```diff
@@ -126,7 +126,7 @@ export function createApi(
                 const a = document.createElement('a');
                 document.body.append(a);
                 a.download = 'gosling-visualization';
-                a.href = URL.createObjectURL(blob);
+                a.href = URL.createObjectURL(blob!);
```
sehilyi, line 129: Thanks for adding this.

src/missing-types.d.ts
```diff
@@ -91,40 +98,233 @@ declare module '@higlass/services' {
...
+        subscribe(msg: string, handler: Handler): Subscription;
+        unsubscribe(msg: string): void;
+    };
+
+    type TrackOptions = Record<string, unknown>;
+
     type Track = any;
-    export declare const BarTrack: Track;
-    export declare const PixiTrack: Track;
+    export const BarTrack: Track;
```
sehilyi: These two lines seem to be duplicated with the lines above.

src/missing-types.d.ts
```diff
@@ -91,40 +98,233 @@ declare module '@higlass/services' {
...
+        dataFetcher: DataFetcher;
+        getLockGroupExtrema(): [min: number, max: number] | null;
+        handleTilesetInfoReceived(tilesetInfo: TilesetInfo): void;
+        animate(): void;
+        svgElement: HTMLElement;
+        isValueScaleLocked(): boolean;
+        onValueScaleChanged(): void;
+        onTrackOptionsChanged(newOptions: Options): void;
+        // TODO: not sure what the props are here
+        onMouseMoveZoom(props: unknown): void;
```
sehilyi: Looks like this function is used in both [HorizontalTiled1DPixiTrack.js](https://github.com/higlass/higlass/blob/54f5aae61d3474f9e868621228270f0c90ef9343/app/scripts/HorizontalTiled1DPixiTrack.js#L201) and [HeatmapTiledPixiTrack.js](https://github.com/higlass/higlass/blob/54f5aae61d3474f9e868621228270f0c90ef9343/app/scripts/HeatmapTiledPixiTrack.js#L224), and the type seem to be the following (ignore the random type names):

```ts
interface commonProps {
  trackId: string,
  data: number,
  absX: number,
  absY: number,
  relX: number,
  relY: number,
  dataX: number,
  dataY: number,
}
interface commonProps1D extends commonProp {
  orientation: '1d-horizontal' | '1d-vertical';
}
interface commonProps2D extends commonProp {
  orientation: '2d',
  dataLens: Float32Array,
  dim: number,
  toRgb: [number, number, number, number],
  center: [number, number],
  xRange: [number, number],
  yRange: [number, number],
  isGenomicCoords: boolean,
}
type props = commonProps1D | commonProps2D;
```

sehilyi (commented):
Thank you for putting this effort to define missing types. This PR looks good to me!

src/gosling-genomic-axis/axis-track.ts
```diff
@@ -8,35 +8,134 @@ import { format, precisionPrefix, formatPrefix } from 'd3-format';
...
+    function createTickText(text: string, style: Partial<PIXI.ITextStyle>): TickText {
+        return Object.assign(new HGC.libraries.PIXI.Text(text, style), { hashValue: Math.random() });
+    }
+
+    class AxisTrackClass extends HGC.tracks.PixiTrack<typeof options> {
         allTexts: TickLabelInfo[];
+        searchField: null;
+        chromInfo: ChromInfo;
+        dataConfig: Record<string, unknown>;
+        pTicksCircular: PIXI.Graphics;
```
sehilyi, line 108: Given that we have explicit types, I can gradually revisit each file in the future and try to directly use HGC libraries, e.g., `HGC.libraries.PIXI.Graphics`.

src/missing-types.d.ts
manzt: thanks so much for looking into this! The only type I changed was `dataLens` to `ArrayLike<number>` since it is assigned differently in the component.

src/gosling-genomic-axis/axis-track.ts
manzt: Since types are erased, it technically shouldn't matter where they are imported but I agree for consistency this would be nice!

PR #734, opened 2022-06-25
feat: use `definePluginTrack` utility for `AxisTrack`

manzt:
Demonstrates use of `definePluginTrack` utility from #732. If these changes are acceptable, we can merge this PR with #732 and then merge in `master`. We can gradually re-define other custom HiGlass Tracks with this utility later. 

> Note: I squashed the changes on #732 so that I could apply these changes in a single diff on top.

### DIFF
--- a/src/core/utils/text-style.ts
+++ b/src/core/utils/text-style.ts
@@ -1,5 +1,3 @@
-import type * as PIXI from 'pixi.js';
-
 export type TextStyle = {
     color?: string;
     size?: number;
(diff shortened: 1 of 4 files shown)


src/core/utils/text-style.ts
```diff
@@ -18,15 +16,20 @@ export const DEFAULT_TEXT_STYLE: Required<TextStyle> = {
...
+        fill: merged.color,
         lineJoin: 'round',
-        stroke: style.stroke ?? DEFAULT_TEXT_STYLE.stroke,
-        strokeThickness: style.strokeThickness ?? DEFAULT_TEXT_STYLE.strokeThickness
+        stroke: merged.stroke,
+        strokeThickness: merged.strokeThickness
+    } as const;
+    // Allow returned object to be mutable (strip `readonly` modifier from `const`)
+    return pixiTextStyle as {
+        -readonly [K in keyof typeof pixiTextStyle]: typeof pixiTextStyle[K];
```
sehilyi, line 33: Nice!

sehilyi (approved):
Thanks for demonstrating this example. The new way of implementing a track looks reasonable to me!

Since I just generally prefer having a less number of indentations, is it possible to define the two parameters of `definePluginTrack()` separately?

```ts
const config: /* ? */ = {
// ...
};
const factory: /* ? */ = (HGC, context, options) ⇒ {
// ...
};
const AxisTrack = definePluginTrack(config, factory);
```

manzt:
@sehilyi Have a look at the changes I just made. We can define an `Options` type for each Plugin Track and then use  `TrackConfig<Options>` and `PluginTrackFactory<Options>` to infer the types of `config` and `factory`.

sehilyi:
Thanks! It looks great to me.

PR #739, opened 2022-06-29
feat(data-fetchers): migrate BAM worker to TS

manzt:
The worker code is some of the more complicated in this repo and does not have any type safety.

I tried to type out the information to the best of my ability and removed and remove vestigial `any` properties which are unused. I will continue with the other workers.

The most notable change is that I replaced `lru-cache` with `quick-lru`. Ilan and I have run into an issue with `lru-cache` because it changed its underlying implementation recently, and `quick-lru` is a much more simple (and typed) implementation of a cache.


manzt:
@sehilyi I bumped the version of TS to get access to the `Awaited` type utility. Things seem to be compiling fine, but we should also bump the eslint-ts things.

package.json
```diff
@@ -87,6 +88,7 @@
         "@testing-library/react": "^10.4.8",
         "@testing-library/user-event": "^12.1.1",
         "@types/d3-drag": "1",
+        "@types/d3-request": "^1.0.6",
```
sehilyi, line 91: Maybe I missed it – is this used somewhere in the PR?

src/missing-types.d.ts
```diff
@@ -312,10 +315,11 @@ declare module '@higlass/utils' {
     import type { ScaleContinuousNumeric } from 'd3-scale';
     import type { TilesetInfo } from '@higlass/services';
 
-    type ChromInfo<Name extends string> = {
+    export type ChromInfo<Name extends string = string> = {
```
sehilyi: Just learned the `= string` syntax :)

Could you teach me why `= string` is necessary when `Name` is extending `string`? I assume this is because the `chrom` of `chrToAbs` allows non-string types (e.g., `chr1` | `chr1` | ...)?

src/data-fetcher/bam/bam-worker.ts
```diff
@@ -4,30 +4,34 @@ import { bisector } from 'd3-array';
...
-    return type;
-};
+import QuickLRU from 'quick-lru';
+
+import type { ChromInfo, TilesetInfo } from '@higlass/types';
+import type { BamRecord } from '@gmod/bam';
+
+// not used anywhere?
+/* eslint-disable-next-line @typescript-eslint/no-unused-vars */
+function cigarTypeToText<T extends string>(type: T) {
```
sehilyi: I think this is not indeed used and we can remove this...

src/data-fetcher/bam/bam-worker.ts
```diff
@@ -335,50 +361,75 @@ function ChromosomeInfo(filepath, success) {
...
 /////////////////////////////////////////////////////
 
-const bamRecordToJson = (bamRecord, chrName, chrOffset) => {
+const bamRecordToJson = (bamRecord: BamRecord, chrName: string, chrOffset: number) => {
     const seq = bamRecord.get('seq');
-    const segment = {
+    const segment: Segment = {
         // if two segments have the same name but different id, they are paired reads.
         // https://github.com/GMOD/bam-js/blob/7a57d24b6aef08a1366cca86ba5092254c7a7f56/src/bamFile.ts#L386
+        // @ts-expect-error private field!!
```
sehilyi, line 356: Ah, I think we can later add this function directly to the class to avoid the error?

sehilyi (approved):
@manzt Apologize that it took some time to review this PR, and thanks a lot for making this change. Using TS is a huge addition to maintaining workers. To be honest I did not know I just needed to simply convert the syntax (from JS to TS) w/o additionally editing the bundling part.

src/missing-types.d.ts
manzt: `= string` provides a default for the generic so you can use the type without `<>` if desired. I think it's sensible to have a default because we likely don't have enough type information often (so falling back to `string` is most flexible). But in the case where we do have more information (the chrom names), we can capture that.

https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgMIAsoHsC2BJEGLAdWDHSwFcwARCeSgGzAB4A5OHFCAD0hAAmAZ2RCwUUAHMAfMgDeAKGTIElHAAUsQssCwghALnnIADlqMg1AI2gBuFZiNiJIScgC+AbQC6tpQ6hNbTBdfSMAJQgELCgBdk4IABpjM0NkSxwbKA9pP2UETFwAGQhXcjTI6Nj4rmSMrNz-MCwwOEYSsvQLazsFdwUFASjGOCgUaP0wZEgxAAYjDGx8QhIyCmo6BmY-QeHR8b0xaYgxAEYFwuWiUnIqWno4JlZnKUbFfMOpuVMtD2QAXmOZwAdAVAlodIdgVYRuhkAB6eHpLDIaDYKAAQj6CneKk+KV+7gBQLAp1BmCCkP05Kgp2xuwQIzGeMmJIATBclgRrms7ptHswWAByMGnIVvfwTI7fVJ-QEzMBsmmUkJQmFwdDY3FSr4-ERE+UnRXKiGq6mi+kKUCQWCIFCLXDcrA1bh8UrCUTiKTEl6uWTatQq0JpGXmdI9KD2MFOL2uDw+PIBIOHCJRGJxDi1AlperQHKJsHFUqScqpqoZhJ1CONZTNVrtYvkbqZXr9BlMg6s0BETmOlb2RGiBLIOAiXicEyMFBskdWLAANwgA21+ND+uJ3awJuCwehsK1ktXerlyE326pQhpdLbQ0Z+xZR035zQlydwtF4r8K9Za5PT-PZqXuqmr9N+0rHgap4rGSYLJuamDXgMt4dsgMCUCACBmgEAAqWAAIJWEILqom6ggiL6MgABQgAkRiZkkUE9i+XIrC60gAJTNlkOxgrhBFCJRQroBAjCMFgQrJJu7EKLx+GEYJwmieJknQexQA

manzt:
```typescript

declare const info: ChromInfo;

{
  const { pos } = info.chrPositions.blah // ok because any string key is valid
}

{
  const { pos } = info.chrPositions.chr1
}
```


```typescript

declare const info: ChromInfo<'chr1' | 'chr2'>;

{
  const { pos } = info.chrPositions.blah // error!!! not 'chr1' or 'chr2' 
}
```

package.json
manzt: the `text` function is imported at the tope of `src/data-fetcher/bam/bam-worker.ts`. Since it was a JS file previously, there was no type checking and the compiler didn't care that we didn't have `d3-request` available.

src/data-fetcher/bam/bam-worker.ts
manzt: The `BamRecord` type coming from `@gmod/bam`, so it would need to be modified there. Perhaps there is a different method where we can get the `_id` without relying on a private field.

manzt: TypeScript's `private` modifier is technically not actually private in runtime code, but trying to access that field means that we are relying on behavior that could change from the library without us knowing. I doubt it will, but something to keep in mind. Additionally, javascript recently added actual [private class fields](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields), which if `gmod` ever used instead we couldn't rely on accessing the "private" fields.

manzt: Perhaps `bamFile.seq_id` would be sufficent:

https://github.com/GMOD/bam-js/blob/9533ae737235b78876232cd923b4342b4de11987/src/record.ts#L49

PR #740, opened 2022-06-30
feat(data-fetchers): migrate VCF data-fetcher to TS

manzt:
Follow up to #739


src/data-fetcher/utils.ts
```diff
@@ -0,0 +1,81 @@
...
+
+    return info;
+}
+
+export type ExtendedChromInfo = HiGlass.ChromInfo & {
+    absToChr(absPos: number): ReturnType<typeof absToChr> | null;
+    chrToAbs(chr: [name: string, pos: number]): number | null;
+};
+
+export async function fetchChromInfo(url: string): Promise<ExtendedChromInfo> {
```
manzt, line 71: We should be able to reuse this module for the BAM worker and others.

src/data-fetcher/vcf/gosling-vcf-data.ts
```diff
@@ -3,41 +3,34 @@
...
     private uid: string;
-    private fetchTimeout: any;
-    private toFetch: Set<any>;
+    private assembly: Assembly;
+    private toFetch: Set<string>;
+    private fetchTimeout?: ReturnType<typeof setTimeout>;
     private track: any;
-    private prevRequestTime: number;
-    private tbiIndexed: any;
-    private tbiVCFParser: any;
```
manzt, line 26: I only kept properties which are assigned in the constructor.

src/missing-types.d.ts
```diff
@@ -312,10 +315,11 @@ declare module '@higlass/utils' {
     import type { ScaleContinuousNumeric } from 'd3-scale';
     import type { TilesetInfo } from '@higlass/services';
 
-    type ChromInfo<Name extends string> = {
+    export type ChromInfo<Name extends string = string> = {
```
sehilyi: I think we can reuse this in our assembly definitions as well:

https://github.com/gosling-lang/gosling.js/blob/db1b657f3918272759236c1317c0ab00fc4acfb2/src/core/utils/assembly.ts#L12

sehilyi (approved):
Looks good to me!

src/missing-types.d.ts
manzt: Ok great!

manzt:
Hold on... I think I know how to make these diffs nicer..

PR #743, opened 2022-07-01
fix: override RemoteFile.read to be more flexible

manzt:
Client-side loading of `bigwig`/`bam`/`vcf` is currently block when using the local data-server in Gos _within_ **Google Colab**. However, CSV and other data sources are supported. What gives? Spoiler: Partial content & 200 vs 206 HTTP status codes.

The issue is related to how HTTP requests are proxied by Google Colab. Colab aims to be _very_ flexible by allowing the client-side JavaScript to make requests to `localhost` on the remote machine. Normally a request to `http://localhost:{port}` from the browser would ping the users local machine, but Colab proxies these requests to forward to the _remote machine_. 

After some digging, I discovered that this proxing is accomplished via a custom [service worker](https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API) client-side, that redirects all `fetch` requests with appropriate auth (pretty cool). Unfortunately, it seems that this proxing overrides the HTTP response header from a 206 to a 200 even if the content is actually partial (via `content-size` header). This subtly causes `generic-filehandle` to throw an error in `RemoteFile.read`

https://github.com/GMOD/generic-filehandle/blob/0e8209be25e3097307bd15e964edd8c017e808d7/src/remoteFile.ts#L134-L158

This PR overrides the `read` method to be more flexible and just read the content of any 200 or 206 response, which should solve this weird bug with Colab. I guess we could add an additional check which asserts that the `content-size` is what is expected if we asked for partial content but received a 200, but this was the most simple fix.

cc: @cmdcolin

### DIFF
--- a/src/data-fetcher/bam/bam-worker.ts
+++ b/src/data-fetcher/bam/bam-worker.ts
@@ -6,7 +6,7 @@ import QuickLRU from 'quick-lru';
 import type { ChromInfo, TilesetInfo } from '@higlass/types';
 import type { BamRecord } from '@gmod/bam';
 
-import { fetchChromInfo } from '../utils';
+import { fetchChromInfo, RemoteFile } from '../utils';
 
 function parseMD(mdString: string, useCounts: true): { type: string; length: number }[];
 function parseMD(mdString: string, useCounts: false): { pos: number; base: string; length: 1; bamSeqShift: number }[];
@@ -349,10 +349,10 @@ const init = (
     if (!bamFiles[bamUrl]) {
         // we do not yet have this file cached
         bamFiles[bamUrl] = new BamFile({
-            bamUrl,
-            baiUrl
-            // fetchSizeLimit: 500000000,
-            // chunkSizeLimit: 100000000 ,
+            bamFilehandle: new RemoteFile(bamUrl),
+            baiFilehandle: new RemoteFile(baiUrl),
+            // fetchSizeLimit: 500000000, 
+            // chunkSizeLimit: 100000000 , 
             // yieldThreadTime: 1000
         });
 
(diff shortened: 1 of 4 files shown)


manzt:
Colab localhost stuff: https://stackoverflow.com/questions/63478954/on-google-colab-localhost-access-from-javascript-fails

cmdcolin:
nice catch. this has been seen in a recent issue on our end too, it is a hard error to debug since it doesn't really contain much other context beside the status codes. it seems like it can be an indication of a non-compliant server but if it is actually giving proper data, may be worth allowing!

manzt:
> it seems like it can be an indication of a non-compliant server

This was my main concern any why I didn't directly open a PR directly. 

>  but if it is actually giving proper data, may be worth allowing!

Right, I'm thinking that if we run into the case where we know data are correct (e.g., `content-size` header is present and what we expect) then we could just be a bit more flexible. Another approach would be to set the "flexible" behavior with some kind of flag, e.g.,

```typescript
let fh = new RemoteFile('https://example.com/data.bw', { strict: true }); // or { mode: 'strict' }
```

But my feeling would be to try the former (sensible default) without exposing more API and defining semantics over strictness.

Happy to open a PR if that is of interest.

manzt:
@sehilyi I was able to reproduce the Colab error locally by changing the response header on the Gos server to `200` for partial content (instead of `206`) and can confirm that this patch fixes the problem from our side.

src/data-fetcher/bam/bam-worker.js
```diff
@@ -374,9 +375,9 @@ const dataConfs = {};
 const init = (uid, { bamUrl, baiUrl, chromSizesUrl, loadMates = false, maxInsertSize = 5000, extractJunction = false, junctionMinCoverage = 1}) => {
     if (!bamFiles[bamUrl]) {
         // we do not yet have this file cached
-        bamFiles[bamUrl] = new BamFile({ 
-            bamUrl, 
-            baiUrl, 
+        bamFiles[bamUrl] = new BamFile({
+            bamFilehandle: new RemoteFile(bamUrl),
+            baiFilehandle: new RemoteFile(baiUrl),
```
manzt: Ensures that we use our custom implementation.

sehilyi (approved):
Nice!!

PR #741, opened 2022-06-30
feat(data-fetchers): migrate bigwig to TS

manzt:
Extends #739 & #740

PR #744, opened 2022-07-02
refactor(data-fetchers): scope worker imports to data-fetchers

manzt:
The instantiation & use of the BAM or VCF data-fetcher `Worker` directly in `gosling-track.ts` is somewhat of a leaky abstraction. Ideally the data-fetcher alone provides a unified API for fetching data, rather than a mix of `worker` and `dataFetcher`. 

This PR refactors the BAM and VCF data-fetchers to:

1.) Instantiate their own workers (within the constructor).
2.) Hide the use of a web `Worker` behind the public data fetcher API. Instead an async `getTabularData` method is added (which can be called by `GoslingTrack`) and internally calls out to the worker.

This refactor provides a more consistent data fetching interface, and allows for `getTabularData` to be implemented in other data-fetchers without requiring the use of `threads` module and a web worker.

Perhaps this makes the most sense to review/merge after #739, #740, #741.

### DIFF
--- a/src/data-fetcher/bam/bam-datafetcher.ts
+++ b/src/data-fetcher/bam/bam-datafetcher.ts
@@ -2,6 +2,9 @@
  * This code is based on the following repo:
  * https://github.com/higlass/higlass-pileup
  */
+import { spawn } from 'threads';
+import Worker from './bam-worker.ts?worker&inline';
+
 import type { Assembly } from '../../core/gosling.schema';
 import type { ModuleThread } from 'threads';
 import type { WorkerApi, TilesetInfo, Tiles } from './bam-worker';
@@ -11,6 +14,7 @@ const DEBOUNCE_TIME = 200;
 class BAMDataFetcher {
     dataConfig: Record<string, any>;
     assembly: Assembly;
+    worker: Promise<ModuleThread<WorkerApi>>;
 
     uid: string;
     fetchTimeout?: ReturnType<typeof setTimeout>;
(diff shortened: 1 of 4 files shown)


manzt:
This PR is ready for review. Something for a follow-up however:

I noticed that `vcf-worker.ts` and `bam-worker.ts` both cache information by dataconfig UID/urls globally in the worker scope. However, the initialization code (within `gosling-vcf-data` and `bam-datafetcher`) create a **new worker per data fetcher (data-config)**. For example,

```typescript
const HGC = { /* ... */ };
const dataConfigs = [ /* ... */ ];
const dataFetchers = [];
for (const config of dataConfigs) {
  const dataFetcher = new BamDataFetcher(HGC, config);
  dataFetchers.push(dataFetcher);
}
```

spawns 4 new workers with a cache size of 1, instead of 1 worker with a cache size of 4.

To change this, we could initialize a `worker` per data-fetcher module:

```typescript
// bam-datafetcher.ts
let worker: Promise<ModuleThread<WorkerApi>>;

class BAMDataFetcher {
  private ready: Promise<void>;
  constructor(HGC: HiGlass.HGC, dataConfig: DataConfig) {
    /* ... */
    if (!worker) worker = spawn<WorkerApi>(new Worker());
    this.ready = worker.then(api => api.init(/* ... */));
  }
  async tilesetInfo(...args) {
    await this.ready;
    worker!.tilesetInfo(...args);  
  }
}
```

sehilyi (approved):
I agree with the motivation, and changes look good to me.

Also, the idea about the follow-up PR sounds good!

PR #748, opened 2022-07-03
refactor(utils): remove use of `any` in array utils

manzt:
Preserves array item type information in array utils and removes use of `any`.


sehilyi (approved):
Nice!

PR #749, opened 2022-07-03
fix: include type definitions in the bundle; do not bundle types for editor codes

### LINKED ISSUE #735: Fail to compile after upgrading gosling.js
The react app fail to compile after upgrading gosling.js from `0.9.0` to `0.9.19`. 
The error info is as below.

```
Failed to compile.

./node_modules/gosling.js/dist/gosling.es.js
Attempted import error: 'generic-filehandle' does not contain a default export (imported as 'require$$7').
```

@sehilyi , any thoughts about what might need to this issue?

sehilyi:
Fix #735 

- Correctly bundle type definitions
- Do not bundle types for `\editor` (which was bundled previously)

I basically turned off the default behavior of Vite that made a `dist` folder empty before bundling (https://vitejs.dev/config/#build-emptyoutdir), which led to removing already created type definitions.

You can now find the `src` folder that contains types after running `yarn build`:

Screen Shot 2022-07-03 at 17 28 21

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -14,7 +14,7 @@ module.exports = {
         node: true
     },
     parserOptions: {
-        project: 'tsconfig.editor.json',
+        project: 'tsconfig.json',
         ecmaVersion: 6,
         sourceType: 'module'
     },
--- a/package.json
+++ b/package.json
@@ -21,10 +21,11 @@
     "scripts": {
         "start": "vite --mode editor",
         "start-embed": "vite",
-        "build": "run-p build-types build-lib",
+        "build": "run-s build-clear build-types build-lib",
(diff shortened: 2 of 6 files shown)


manzt (approved):
Ah I see the issue. Thanks!

sehilyi:
@manzt I am thinking of bumping a version. Let me know if you would like to include any other remaining PRs in the next version `0.9.20`

manzt:
I'm just going to merge #743, and then it should be good to bump the version! I will then bump `gos` to `v0.0.9` with the latest, and run the notebooks to ensure everything is working.

manzt:
Ok merged, good for a version bump. Thanks for all the reviews today!

PR #759, opened 2022-07-06
chore: specify data config types for csv and json

sehilyi:
This PR adds types of `dataConfig` for CSV and JSON data fetchers:

```diff
+ type CsvDataConfig = JSONData & { assembly: Assembly };
...
- // @ts-ignore
- private dataConfig: GeminiDataConfig;
+ private dataConfig: CsvDataConfig;
```

I also made the file/folder/class names of data fetchers consistent. Thoese were quite inconsistent, e.g., `bam-datafetcher.ts` vs. `gosling-vcf-data.ts` vs. `higlass-csv-datafetcher.ts`.


src/data-fetcher/json/json-datafetcher.ts
```diff
@@ -55,28 +57,32 @@ function RawDataFetcher(HGC: any, ...args: any): any {
...
-                    } catch (e) {
-                        // genomic position did not parse properly
-                        successfullyGotChrInfo = false;
-                        // console.warn(
-                        //     '[Gosling Data Fetcher] Genomic position cannot be parsed correctly.',
-                        //     this.dataConfig.chromosomeField
-                        // );
-                    }
-                });
+                if (genomicFields && chromosomeField) {
```
sehilyi: These properties can be undefined, but since we did not use the explicit types, we were ignoring this case.

manzt (approved):
This looks good to me! Thanks for making the naming more consistent as well. A couple more thoughts on naming:

It would make sense to me to `mv src/data-fetcher src/data-fetchers` since the directory includes many data-fetcher definitions. 

I would have a slight preference to keep use `data-fetcher.ts` over `datafetcher.ts` in the individual file names e.g (`bigwig-data-fetcher.ts`). This would follow the directory name `data-fetchers` as well as the fact that we name the classes `BigwigDataFetcher` (not `BigwigDatafetcher`).

In addition the `data-fetcher/*/index` file just re-exports the default `class` from the corresponding file. I would have a slight preference to create a `src/data-fetchers/index.ts` which just exports these names instead (like a package entry point).

```typescript
// src/data-fetcher/index.ts
export { default as RawDataFetcher } from './raw/raw-data-fetcher';
export { default as CsvDataFetcher } from './csv/csv-data-fetcher';
```

This way in other parts of the code base we can import all data-fetchers together:

```typescript
// src/core/init.ts
import { CsvDataFetcher, RawDataFetcher } from '../data-fetchers/index';
```

and if we ever make a monorepo, just replace the import path:

```typescript
// src/core/init.ts
import { CsvDataFetcher, RawDataFetcher } from '@gosling-lang/data-fetchers'
```

sehilyi:
Thank you for the suggestions! They all sound good, and I will update them accordingly

PR #767, opened 2022-07-15
chore: Vite v3.0

manzt:
Migrates Vite to the latest major version (3.0). The biggest change (from our perspective) is that vite now unifies dev/build bundling with esbuild, so the `esbuildOptions` accomplish the behavior of the custom plugin we had previously.

Vite v3 also changes how workers are loaded, and the `bundleWebWorker` plugin no longer works. This means that we will need to develop in Chrome (which is the only browser that supports `Worker(..., { type: 'module' })`). The final build for the editor can be tested in safari and firefox.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,13 +9,13 @@
     },
     "homepage": "https://gosling-lang.github.io/gosling.js/",
     "main": "dist/gosling.js",
-    "module": "dist/gosling.es.js",
+    "module": "dist/gosling.mjs",
     "types": "dist/src/index.d.ts",
     "files": [
         "dist"
     ],
     "exports": {
-        ".": "./dist/gosling.es.js",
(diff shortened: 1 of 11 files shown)


manzt:
hmm now the final build doesn't seem to be working...

sehilyi:
Thanks for this update! Could you elaborate on the issue with the final build?

When I test this locally, I can see that it builds fine.

Screen Shot 2022-07-29 at 14 29 51
Screen Shot 2022-07-29 at 14 32 50

manzt:
Sorry... should have elaborated so it's not just confusing. The problem is that the final editor build `yarn run build-editor` is broken. The site throws an error for the BAM example, since the worker code uses `Buffer`...

manzt:
closing bc we should now upgrade to vite 4

PR #769, opened 2022-07-17
feat: make `column`, `position`, `row` properties optional for `bigwig`, `vector`, and `multivec`

### LINKED ISSUE #768: default `column` and `position` for bigwig data
Gosling aims in part to provide a consistent tabular data abstraction for genomics data formats. However, for some binary formats this means mapping the binary representation to something columnar. 

For bigwig, the user must explicitly provide explicit names for the `column` and `value`. In Gos, I am considering setting these fields by default to `position` and `value` so that users can avoid repetition.

```python
import gosling as gos

data = gos.bigwig("https://example.com/data.bw")

gos.Track(data).mark_bar().encode(
	x=gos.X("position:G"),
    y=gos.Y("value:Q"),
)
```

Of course this default behavior is something that we can decide in a library like Gos, but it got me thinking about whether it is necessary to allow these fields to be configured in the first place. What if for bigwig, the user always has access to two labeled fields: `position` and `value`. This would mimic the behavior of a csv with named columns. The user doesn't have the ability to name columns of a csv with headers provided, so any bigwig would just be like a csv with the header `position` and `value` always.

sehilyi:
Fix #768 

```diff
// multivec
"data": {
  "url": "https://server.gosling-lang.org/api/v1/tileset_info/?d=cistrome-multivec",
  "type": "multivec",
-  "row": "category",
-  "column": "position",
-  "value": "value",
  "binSize": 4
},

// bigwig or vector
"data": {
  "url": "https://s3.amazonaws.com/gosling-lang.org/data/DopaNeurons_Cluster10_AllFrags_projSUNI2_insertions_bin100_RIPnorm.bw",
  "type": "bigwig",
-  "column": "position",
-  "value": "peak"
},
```

### DIFF
--- a/src/core/gosling.schema.ts
+++ b/src/core/gosling.schema.ts
@@ -873,17 +873,17 @@ export interface MultivecData {
     /**
      * Assign a field name of the middle position of genomic intervals.
      */
-    column: string;
+    column?: string;
 
     /**
      * Assign a field name of samples.
      */
-    row: string;
+    row?: string;
 
     /**
      * Assign a field name of quantitative values.
      */
-    value: string;
+    value?: string;
 
     /**
      *  assign names of individual samples.
@@ -920,12 +920,13 @@ export interface BIGWIGData {
     /**
      * Assign a field name of the middle position of genomic intervals.
      */
-    column: string;
+    column?: string;
 
     /**
      * Assign a field name of quantitative values.
      */
-    value: string;
+    value?: string;
+
     /**
      * Assign a field name of the start position of genomic intervals.
      */
@@ -958,10 +959,10 @@ export interface VectorData {
     url: string;
 
     /** Assign a field name of the middle position of genomic intervals. */
-    column: string;
+    column?: string;
 
     /** Assign a field name of quantitative values. */
-    value: string;
+    value?: string;
 
     /** Assign a field name of the start position of genomic intervals. */
     start?: string;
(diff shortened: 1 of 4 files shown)


src/core/utils/spec-preprocess.ts
```diff
@@ -524,11 +524,16 @@ export function overrideDataTemplates(spec: GoslingSpec) {
         switch (t.data.type) {
             case 'vector':
             case 'bigwig':
-                ts[i] = Object.assign(getVectorTemplate(t.data.column, t.data.value), t);
+                ts[i] = Object.assign(getVectorTemplate(t.data.column ?? 'position', t.data.value ?? 'value'), t);
```
sehilyi, line 527: Ignore the changes in this file. I will remove this `overrideDataTemplates()` function in a separate PR as we will make data templates deprecated. So, default values for `column`, `value`, and `row` are assigned in a single place, i.e., `data-abstraction.ts`.

src/core/gosling.schema.ts
```diff
@@ -982,6 +983,7 @@ export interface VectorData {
  */
 export interface BEDDBData {
```
sehilyi, line 984: Should we change letter cases from

`JSONData, CSVData, BIGWIGData, VCFData, BAMData, BEDDBData`

to

`JsonData, CsvData, BigWigData, VcfData, BamData, BeddbData`

consistent with data fetcher class names? (e.g., `BigWigDataFetcher`)

manzt: I'd be in support of this choice!

manzt (approved):
Looks great!

src/core/gosling.schema.ts
sehilyi: Okay! I will update this in a follow-up PR.

PR #773, opened 2022-07-19
fix: use consistent letter case for data types (e.g., `JSONData` → `JsonData`)

sehilyi:
Changes from

`JSONData, CSVData, BIGWIGData, VCFData, BAMData, BEDDBData`

to

`JsonData, CsvData, BigWigData, VcfData, BamData, BeddbData`,

consistent with data fetcher class names (e.g., `BigWigDataFetcher`)

### DIFF
--- a/schema/gosling.schema.json
+++ b/schema/gosling.schema.json
@@ -372,14 +372,9 @@
         },
         "scaleOffset": {
           "description": "Whether to use offset of the domain proportionally. This is bound to brushes on the color legend. __Default__: `[0, 1]`",
-          "items": [
-            {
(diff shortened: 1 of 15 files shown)


schema/gosling.schema.json
```diff
@@ -3370,7 +3370,41 @@
       ],
       "type": "object"
     },
-    "JSONData": {
+    "JSONParseTransform": {
```
manzt: Should this also change to `JsonParseTransform`?

manzt (approved):
Looks good to me!

PR #774, opened 2022-07-19
feat(api): capture mouseover event on a track; click event on a linear track

sehilyi:
```ts
goslingRef.current.api.subscribe('trackMouseOver', (type, e) ⇒ {
  console.log(
    type, // 'trackMouseOver'
    e.id, // identical to `id` (random ID when unspecified), e.g., "track-2"
    e.spec, // spec of a source track, e.g., { assembly: 'hg38', layout: 'linear', ...}
    e.shape, // shape of the track, e.g., {x: 0, y: 250, width: 600, height: 130}
  )
});
```

```ts
goslingRef.current.api.subscribe('trackClick', (type, e) ⇒ {
  console.log(
    type, // 'trackClick'
    e.id, // identical to `id` (random ID when unspecified), e.g., "track-2"
    e.spec, // spec of a source track, e.g., { assembly: 'hg38', layout: 'linear', ...}
    e.shape, // shape of the track, e.g., {x: 0, y: 250, width: 600, height: 130}
  )
});
```

(We only support Click event on linear tracks.)

### DIFF
--- a/package.json
+++ b/package.json
@@ -131,7 +131,7 @@
         "react-split-pane": "^0.1.91",
         "remark-gfm": "^1.0.0",
         "strip-json-comments": "^3.1.1",
-        "ts-json-schema-generator": "^0.75.0",
+        "ts-json-schema-generator": "^1.0.0",
         "typescript": "^4.7.4",
         "vite": "^2.9.5",
         "vitest": "^0.10.0"
--- a/schema/gosling.schema.json
+++ b/schema/gosling.schema.json
@@ -13,17 +13,43 @@
       "type": "string"
     },
     "Assembly": {
-      "enum": [
-        "hg38",
-        "hg19",
-        "hg18",
-        "hg17",
-        "hg16",
-        "mm10",
-        "mm9",
-        "unknown"
-      ],
-      "type": "string"
+      "anyOf": [
+        {
+          "const": "hg38",
+          "type": "string"
+        },
+        {
+          "const": "hg19",
+          "type": "string"
+        },
+        {
+          "const": "hg18",
+          "type": "string"
+        },
+        {
+          "const": "hg17",
+          "type": "string"
+        },
+        {
+          "const": "hg16",
+          "type": "string"
+        },
+        {
+          "const": "mm10",
+          "type": "string"
+        },
+        {
+          "const": "mm9",
+          "type": "string"
+        },
+        {
+          "const": "unknown",
+          "type": "string"
+        },
+        {
+          "$ref": "#/definitions/ChromSizes"
+        }
+      ]
     },
     "AxisPosition": {
       "enum": [
@@ -35,7 +61,7 @@
       ],
       "type": "string"
     },
-    "BAMData": {
+    "BamData": {
       "additionalProperties": false,
       "description": "Binary Alignment Map (BAM) is the comprehensive raw data of genome sequencing; it consists of the lossless, compressed binary representation of the Sequence Alignment Map-files.",
       "properties": {
@@ -75,7 +101,7 @@
       ],
       "type": "object"
     },
-    "BEDDBData": {
+    "BeddbData": {
       "additionalProperties": false,
       "description": "Regular BED or similar files can be pre-aggregated for the scalable data exploration. Find our more about this format at [HiGlass Docs](https://docs.higlass.io/data_preparation.html#bed-files).",
       "properties": {
@@ -183,7 +209,7 @@
       ],
       "type": "object"
     },
-    "BIGWIGData": {
+    "BigWigData": {
       "additionalProperties": false,
       "properties": {
         "aggregation": {
@@ -231,82 +257,6 @@
         "sum"
       ],
       "type": "string"
-    },
-    "CSVData": {
-      "additionalProperties": false,
-      "description": "Any small enough tabular data files, such as tsv, csv, BED, BEDPE, and GFF, can be loaded using \"csv\" data specification.",
-      "properties": {
-        "chromosomeField": {
-          "description": "Specify the name of chromosome data fields.",
-          "type": "string"
-        },
-        "chromosomePrefix": {
-          "description": "experimental",
-          "type": "string"
-        },
-        "genomicFields": {
-          "description": "Specify the name of genomic data fields.",
-          "items": {
-            "type": "string"
-          },
-          "type": "array"
-        },
-        "genomicFieldsToConvert": {
-          "description": "experimental",
-          "items": {
-            "additionalProperties": false,
-            "properties": {
-              "chromosomeField": {
-                "type": "string"
-              },
-              "genomicFields": {
-                "items": {
-                  "type": "string"
-                },
-                "type": "array"
-              }
-            },
-            "required": [
-              "chromosomeField",
-              "genomicFields"
-            ],
-            "type": "object"
-          },
-          "type": "array"
-        },
-        "headerNames": {
-          "description": "Specify the names of data fields if a CSV file does not contain a header.",
-          "items": {
-            "type": "string"
-          },
-          "type": "array"
-        },
-        "longToWideId": {
-          "description": "experimental",
-          "type": "string"
-        },
-        "sampleLength": {
-          "description": "Specify the number of rows loaded from the URL.\n\n__Default:__ `1000`",
-          "type": "number"
-        },
-        "separator": {
-          "description": "Specify file separator, __Default:__ ','",
-          "type": "string"
-        },
-        "type": {
-          "const": "csv",
-          "type": "string"
-        },
-        "url": {
-          "description": "Specify the URL address of the data file.",
-          "type": "string"
-        }
-      },
-      "required": [
-        "type",
-        "url"
-      ],
-      "type": "object"
     },
     "Channel": {
       "anyOf": [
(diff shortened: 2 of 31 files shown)


schema/gosling.schema.json
```diff
@@ -448,14 +448,9 @@
...
           "description": "Whether to use offset of the domain proportionally. This is bound to brushes on the color legend. __Default__: `[0, 1]`",
-          "items": [
-            {
-              "type": "number"
-            },
-            {
-              "type": "number"
-            }
-          ],
+          "items": {
```
sehilyi: Changes in `schema.json` files seem to be due to the upgrade of `ts-json-schema-generator`

sehilyi:
Not sure if the name `trackMouseOver` is fine.

We can later probably unify the track-level events with mark-level events (`mouseOver`, `rangeSelect`, `click`).

package.json
```diff
@@ -131,7 +131,7 @@
         "react-split-pane": "^0.1.91",
         "remark-gfm": "^1.0.0",
         "strip-json-comments": "^3.1.1",
-        "ts-json-schema-generator": "^0.75.0",
```
manzt, line 134: Was this just routine maintenance  or was there something in this PR that required updating the schema generator?

src/core/gosling.schema.ts
```diff
@@ -254,16 +254,51 @@ interface RangeMouseEventData extends CommonEventData {
     genomicRange: [GenomicPosition, GenomicPosition] | null;
 }
 
+/** The visual parameters that determine the shape of a linear track */
+interface LinearTrackShape {
+    x: number; // The origin is the left top corner
+    y: number;
+    width: number;
+    height: number;
+}
```
manzt: Maybe move this comment so that it is picked up by JSDoc.

```suggestion
/** 
 * The visual parameters that determine the shape of a linear track. 
 * Origin is the left top corner.
 */
interface LinearTrackShape {
    x: number;
    y: number;
    width: number;
    height: number;
}
```

src/core/gosling.schema.ts
```diff
@@ -254,16 +254,51 @@ interface RangeMouseEventData extends CommonEventData {
...
+
+/** The visual parameters that determine the shape of a circular track */
+interface CircularTrackShape {
+    cx: number;
+    cy: number;
+    innerRadius: number;
+    outerRadius: number;
+    // Anticlockwise, starts from 12 o'clock.
+    startAngle: number; // [0, 360]
+    endAngle: number; // [0, 360]
```
manzt: Each property can be annotated above with `/** ... */` and it will add to the documentation.

src/core/gosling.schema.ts
```diff
@@ -254,16 +254,51 @@ interface RangeMouseEventData extends CommonEventData {
...
+}
+
+/** The informatino for a mouse event on a track */
```
manzt:
```suggestion
/** The information for a track mouse event */
```

src/core/gosling.schema.ts
```diff
@@ -254,16 +254,51 @@ interface RangeMouseEventData extends CommonEventData {
...
+type TrackMouseEventData = {
+    /** ID of a source track, i.e., `track.id` */
+    id: string;
+
+    /** Source spec that is processed by the Gosling compiler, e.g., default properties filled in. */
```
manzt:
```suggestion
    /** Expanded track specification processed by the Gosling compiler, e.g., default properties filled in. */
```

src/core/gosling.schema.ts
```diff
@@ -254,16 +254,51 @@ interface RangeMouseEventData extends CommonEventData {
...
 export type _EventMap = {
     mouseOver: PointMouseEventData;
     click: PointMouseEventData;
     rangeSelect: RangeMouseEventData;
     rawData: CommonEventData;
+    trackMouseOver: TrackMouseEventData;
 };
 
 export type MouseEventsDeep = {
+    // TODO(Jul-19-2022): Due to cyclic definition of `SingleTrack`, `ts-json-schema-generator` throws an error
```
manzt: ah I see now why the change was necessary

src/core/gosling.schema.ts
```diff
@@ -254,16 +254,51 @@ interface RangeMouseEventData extends CommonEventData {
...
+    // see https://github.com/vega/ts-json-schema-generator/issues/192
+    // [Event in keyof Omit<_EventMap, 'rawData'>]?: boolean;
     /** Turn on and off individual mouse events. */
-    [Event in keyof Omit<_EventMap, 'rawData'>]?: boolean;
+    [Event in 'mouseOver' | 'click' | 'rangeSelect']?: boolean;
```
manzt: It might be easier if we just specified these normally.

```typescript
export interface MouseEventsDeep {
    /** Whether to enable mouseover events. */
    mouseOver: boolean;
    /** Whether to enable click events. */
    click: boolean;
    /** Whether to send range selection events. */
    rangeSelect: boolean;
    /** Group marks using keys in a data field. This affects how a set of marks are highlighted/selected by interaction. __Default__: `undefined` */
    groupMarksByField?: string;
    /** Determine whether all marks underneath the mouse point should be affected by mouse over. __Default__: `false` */
    enableMouseOverOnMultipleMarks?: boolean;
}
```
the original motivation was to reuse the `_EventMap` object, but if we can't then I think it's fine to just have a normal type definition.

src/core/utils/polar.ts
```diff
@@ -29,3 +29,10 @@ export function positionToRadian(x: number, y: number, cx: number, cy: number) {
         return Math.atan((y - cy) / (x - cx)) - Math.PI;
     }
 }
+
+/**
+ * Calculate a degree in the range of [0, 360) based on two points. Anticlockwise, starts from 12 o'clock.
+ */
+export function pointsToDegree(x: number, y: number, cx: number, cy: number) {
+    return ((Math.atan2(-(y - cy), x - cx) / Math.PI) * 180 + 270) % 360;
+}
```
manzt, line 38: Maybe add a couple unit tests?

src/gosling-mouse-event/polygon.ts
```diff
@@ -1,3 +1,31 @@
+import { pointsToDegree } from '../core/utils/polar';
+
+/**
+ * @param {Array} point Tuple of the form `[x,y]` to be tested.
```
manzt: It is usually preferred when using TS with JSDoc to omit the type definition (within `{ }`) in the JSDoc comment. e..,g

```
 * @param point Tuple of the form `[x,y]` to be tested.
 ```

This is because the TypeScript type is much more precise, and the JSDoc comment is ignored by the typechecker completely. We have some JSDoc comments like this in Viv, but that's because our documentation generator relies on the type in the JSDoc (and I'd prefer to get rid of it).

File: src/gosling-track/gosling-track.ts [hunk 11]
```diff
@@ -84,6 +85,8 @@ function GoslingTrack(HGC: import('@higlass/types').HGC, ...args: any[]): any {
         private mRangeBrush: LinearBrushModel;
         private _xScale!: ScaleLinear<number, number>;
         private _yScale!: ScaleLinear<number, number>;
+        private position!: [number, number]; // left and top offsets of a track
+        private dimensions!: [number, number]; // width and height of a track
```
manzt: Changing these to JSDoc comments means the information is accessible when hovering the property anywhere in the file:

```suggestion
        /** left and top offsets of a track */
        private position!: [number, number];
        /** width and height of a track */
        private dimensions!: [number, number];
```

image

src/gosling-track/gosling-track.ts
```diff
@@ -1233,6 +1236,39 @@ function GoslingTrack(HGC: import('@higlass/types').HGC, ...args: any[]): any {
...
+                    isPointInsideDonutSlice(
+                        [mouseX, mouseY],
+                        [width / 2.0, height / 2.0],
+                        [innerRadius, outerRadius],
+                        [startAngle, endAngle]
+                    )
+                ) {
+                    publish('trackMouseOver', {
+                        id: this.viewUid,
+                        spec: JSON.parse(JSON.stringify(this.options.spec)),
```
manzt: I believe you could just use the newish [`structuredClone`](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone) method here:

```suggestion
                        spec: structuredClone(this.options.spec),
```

src/gosling-track/gosling-track.ts
```diff
@@ -1233,6 +1236,39 @@ function GoslingTrack(HGC: import('@higlass/types').HGC, ...args: any[]): any {
...
+                        shape: { cx, cy, innerRadius, outerRadius, startAngle, endAngle }
+                    });
+                }
+            } else {
+                publish('trackMouseOver', {
+                    id: this.viewUid,
+                    spec: JSON.parse(JSON.stringify(this.options.spec)),
```
manzt: Likewise here re: structuredClone

manzt (approved):
Nice work! Just some minor comments. I don't feel too strongly about the naming, but I could imagine some type of grouping by event type that you can configure a listener for:

```typescript
export type MouseEventsDeep = {
   mouseOver?: "track" | "point" | "all"
```

and then the events could be prefixed: `track.mouseOver`

Not sure if I like that better than what we have, but just an idea.

package.json
sehilyi: It is just routine maintenance and is not specific to new features introduced in this PR.

sehilyi:
Thank you for all feedback! FYI, I added support for `click` on a linear track as well with small code modifications.

> I don't feel too strongly about the naming, but I could imagine some type of grouping by event type that you can configure a listener for:

I was also thinking the similar approach but not sure as well what is the best approach. I think we need to build a big picture regarding what mouse events we want to support.

PR #777, opened 2022-07-25
feat(api): get track information; correctly use assembly in zoom APIs

### LINKED ISSUE #544: `zoomTo` will jump to wrong position if assembly is set to hg19
I use hg19 instead of hg38 as genome reference. In the spec the assembly was set to hg19, it turns out that hg38 positions were calculated by zoomTo.

Zhen

sehilyi:
Fix #544

## Change List

- Added an API function `gosRef.current.api.getTrack('traci-1')` to get track information, such as its expanded spec after compiling and shape:

```ts
const info = gosRef.current.api.getTrack('traci-1');

console.log(info);

// { id: 'track-1', spec: { mark: 'rect', ...}, shape: { x: 0, y: 130, width: 400, height: 100 }
// or `undefined` if not found
```

- Added an API function `getTracks` that return information of all tracks.

- Use the `getTrackInfo` function internally for `zoomToExtent` and `zoomTo` functions to correctly use the assembly

Open to any suggestions.

## Checklist

### DIFF
--- a/src/core/api.ts
+++ b/src/core/api.ts
@@ -102,7 +102,8 @@ export function createApi(
 
             const assembly = getTrack(viewId)?.spec.assembly;
             const chr = position.split(':')[0];
-            const chrStart = GET_CHROM_SIZES(assembly).interval?.[chr]?.[0];
+            const info = GET_CHROM_SIZES(assembly);
+            const chrStart = info.interval?.[chr]?.[0];
 
             if (!chr || typeof chrStart === undefined) {
                 console.warn('Chromosome name is not valid', chr);
--- a/src/core/api.ts
+++ b/src/core/api.ts
@@ -1,4 +1,3 @@
-import type React from 'react';
 import * as PIXI from 'pixi.js';
 import type { TrackMouseEventData } from '@gosling.schema';
 import type { HiGlassApi } from './higlass-component-wrapper';
@@ -42,14 +41,14 @@ export interface GoslingApi {
 export function createApi(
     hg: HiGlassApi,
     hgSpec: HiGlassSpec | undefined,
-    trackInfos: React.MutableRefObject<TrackMouseEventData[]>,
+    trackInfos: TrackMouseEventData[],
     theme: Required<CompleteThemeDeep>
 ): GoslingApi {
     const getTracks = () => {
-        return trackInfos.current;
+        return trackInfos;
     };
     const getTrack = (trackId: string) => {
-        const trackInfoFound = trackInfos.current.find(d => d.id === trackId);
+        const trackInfoFound = trackInfos.find(d => d.id === trackId);
         if (!trackInfoFound) {
             console.warn(`[getTrack()] Unable to find a track using the ID (${trackId})`);
         }
--- a/src/core/api.ts
+++ b/src/core/api.ts
@@ -4,7 +4,7 @@ import type { TrackMouseEventData } from '@gosling.schema';
 import type { HiGlassApi } from './higlass-component-wrapper';
 import type { HiGlassSpec } from './higlass.schema';
 import { subscribe, unsubscribe } from './pubsub';
-import { GET_CHROM_SIZES } from './utils/assembly';
+import { GET_CHROM_SIZES, parseGenomicPosition } from './utils/assembly';
 import type { CompleteThemeDeep } from './utils/theme';
 import { traverseViewsInViewConfig } from './utils/view-config';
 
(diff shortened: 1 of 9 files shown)


src/core/api.ts
```diff
@@ -78,31 +93,31 @@ export function createApi(
...
             // Accepted input: 'chr1' or 'chr1:1-1000'
             if (!position.includes('chr')) {
                 console.warn('Genomic interval you entered is not in a correct form.');
                 return;
             }
 
+            const assembly = getTrack(viewId)?.spec.assembly;
             const chr = position.split(':')[0];
-            const chrStart = GET_CHROM_SIZES().interval?.[chr]?.[0];
+            const chrStart = GET_CHROM_SIZES(assembly).interval?.[chr]?.[0];
```
sehilyi: We will need to touch this part when refactoring the assembly

sehilyi:
Maybe do not include a track info object for the title, i.e.,:

```
{
    "id": "741e3e75-6749-43fe-816d-c769e39a4aaa",
    "spec": {
        "mark": "header",
        "width": 600,
        "height": 46,
        "title": "Visual Encoding",
        "subtitle": "Gosling provides diverse visual encoding methods",
        "id": "741e3e75-6749-43fe-816d-c769e39a4aaa"
    },
    "shape": {
        "cx": 300,
        "cy": 23
    }
}
```

src/core/api.ts
manzt: If we absolutely need this info, you should extract it once rather than calling `GET_CHROM_SIZES` multiple times in this funciton.

```suggestion
            const info = GET_CHROM_SIZES(assembly);
            const chrStart = info.interval?.[chr]?.[0];
```

src/core/api.ts
```diff
@@ -78,31 +93,31 @@ export function createApi(
...
 
             if (!chr || typeof chrStart === undefined) {
                 console.warn('Chromosome name is not valid', chr);
                 return;
             }
 
-            const [s, e] = position.split(':')[1]?.split('-') ?? [0, GET_CHROM_SIZES().size[chr]];
+            const [s, e] = position.split(':')[1]?.split('-') ?? [0, GET_CHROM_SIZES(assembly).size[chr]];
```
manzt: I think it would be best to write a utility function to parse this.

```typescript
function parsePosition(position: string): { name: string, start: number, end: number } {
  const [name, intervalString] = postion.split(":");
  const [start, end] = intervalString.split("-").map(s => +s.replaceAll(",", ""))
  return { name, start, end };
}
```

src/core/compile.ts
```diff
@@ -7,9 +7,11 @@ import type { CompleteThemeDeep } from './utils/theme';
 import { renderHiGlass as createHiGlassModels } from './create-higlass-models';
 import { manageResponsiveSpecs } from './responsive';
 
+export type compileCallback = (hg: HiGlassSpec, size: Size, gs: GoslingSpec, trackInfos: TrackMouseEventData[]) => void;
```
manzt:
```suggestion
export type CompileCallback = (hg: HiGlassSpec, size: Size, gs: GoslingSpec, trackInfos: TrackMouseEventData[]) => void;
```

src/core/compile.ts
```diff
@@ -58,5 +60,5 @@ export function compile(
     }
 
     // Make HiGlass models for individual tracks
-    createHiGlassModels(JSON.parse(JSON.stringify(specCopy)), trackInfos, setHg, theme);
+    createHiGlassModels(JSON.parse(JSON.stringify(specCopy)), trackInfos, callback, theme);
```
manzt: Probably should consider changing this API to something `async`/`await` based like most things in Gosling. Lots of HiGlass internals are based on callbacks but IMO it's confusing to switch between callbacks and async/await.

manzt:
```suggestion
    createHiGlassModels(structuredClone(specCopy), trackInfos, callback, theme);
```

manzt: For separate PR/consideration... but this cloning of the spec makes me assume the function is mutating the object in place. I think it would be preferred to have have the function avoid mutation of its arguments and instead clone the object internally and _return_ a new object (whether a clone or not).

src/core/create-higlass-models.ts
```diff
@@ -68,5 +68,23 @@ export function renderHiGlass(
...
+            shape:
+                d.track.layout === 'linear'
+                    ? d.boundingBox
+                    : {
+                          cx: d.boundingBox.x + d.boundingBox.width / 2.0,
+                          cy: d.boundingBox.y + d.boundingBox.height / 2.0,
+                          innerRadius: d.track.innerRadius!,
+                          outerRadius: d.track.outerRadius!,
+                          startAngle: d.track.startAngle!,
+                          endAngle: d.track.endAngle!
```
manzt, line 84: Will these fields always be defined? if not, we shouldn't tell the compiler they are and instead model the potential missing-ness in the `TrackMouseEventData` type.

src/core/gosling-component.tsx
```diff
@@ -54,7 +55,7 @@ export const GoslingComponent = forwardRef<GoslingRef, GoslingCompProps>((props,
     useEffect(() => {
         if (!ref || !hgRef?.current) return;
         const hgApi = hgRef.current;
-        const api = createApi(hgApi, viewConfig, theme);
+        const api = createApi(hgApi, viewConfig, trackInfos, theme);
```
manzt: Why not:

```suggestion
        const api = createApi(hgApi, viewConfig, trackInfos.current, theme);
```

And then we don't need to "fake" a ref with the api.

manzt:
I think this looks good. But I guess I'm missing the motivation for why we need to expose an API to access individual tracks by ID. What is the utility to the end user? It seems we can support custom navigation without adding the `getTrack`/`getTracks` api, and the track information is already in the original spec.

manzt (commented):
I think this looks good. But I guess I'm missing the motivation for why we need to expose an API to access individual tracks by ID. What is the utility to the end user? It seems we can support custom navigation without adding the `getTrack`/`getTracks` api, and the track information is already in the original spec.

src/core/api.ts
sehilyi: Agreed. I think similar parse is happening in several places, so we can reuse this function there.

src/core/compile.ts
sehilyi: Right, that I think makes more sense. I will make this change directly in this PR.

src/core/create-higlass-models.ts
sehilyi: Yes, it is. It is defined as the very first step of compiling (similar to https://github.com/gosling-lang/gosling.js/pull/779#discussion_r933222279). But, it would be best if we could figure out a nice way to define a separate type that we can internally use after compiling (e.g., a type that treats these properties required).

src/core/compile.ts
sehilyi: I agree. I would use the callback as is for now, but that is something that we can definitely change.

sehilyi:
One of the motivations of this PR is to enable end users to more easily integrate external graphical elements around Gosling visualizations, e.g., adding custom buttons and other UI elements to interact with Gosling tracks, etc. With the `shape` information, end users can more easily know where each track is displayed (e.g., cx, cy, startAngle, endAngle, outerRadius, innerRadius for circular tracks). Such information is difficult to accurately infer for end users based on their specs (e.g., need to consider cumulative heights, width, gaps, etc). The returned `spec` is an expanded spec that gives more detailed information about the given track (e.g., assembly of a track which must be missing in the original track definition).

PR #779, opened 2022-07-25
feat: make `width` and `height` of track optional

### LINKED ISSUE #778: Make `width` and `height` optional
I think there is no reason to make the `width` and `height` required.

sehilyi:
Fix #778

## Change List
 - Make width and height optional and set default values when missing:

   - 1D tracks → `width: 600, height: 130`
   - 2D tracks → `width: 600, height: 600`

## Checklist

### DIFF
--- a/src/core/defaults.ts
+++ b/src/core/defaults.ts
@@ -7,10 +7,9 @@ export const DEFAULT_SUBTITLE_HEIGHT = 20; // deprecated
 export const DEWFAULT_TITLE_PADDING_ON_TOP_AND_BOTTOM = 6;
 
 // default track size
-export const DEFAULT_TRACK_HEIGHT_LINEAR = 180;
-export const DEFAULT_TRACK_WIDTH_LINEAR = 360;
-export const DEFAULT_TRACK_HEIGHT_CIRCULAR = 400;
-export const DEFAULT_TRACK_WIDTH_CIRCULAR = 400;
+export const DEFAULT_TRACK_HEIGHT_LINEAR = 130;
+export const DEFAULT_TRACK_WIDTH_LINEAR = 600;
+export const DEFAULT_TRACK_SIZE_2D = 600;
 
 // gab between views
 export const DEFAULT_VIEW_SPACING = 10;
--- a/src/core/utils/bounding-box.ts
+++ b/src/core/utils/bounding-box.ts
@@ -185,14 +185,15 @@ function traverseAndCollectTrackInfo(
     const numTracksBeforeInsert = output.length;
 
     if ('tracks' in spec) {
-        const tracks = spec.tracks as Track[];
+        // following `traverseToFixSpecDownstream`, the width and height of each track are gaurenteed to be defined
+        const tracks = spec.tracks as (Track & { width: number; height: number })[];
 
         if (spec.orientation === 'vertical') {
             // This is a vertical view, so use the largest `height` of the tracks for this view.
-            cumHeight = Math.max(...tracks.map(d => d.height!));
+            cumHeight = Math.max(...tracks.map(d => d.height));
             tracks.forEach((track, i, array) => {
                 if (getNumOfXAxes([track]) === 1) {
-                    track.width! += HIGLASS_AXIS_SIZE;
+                    track.width += HIGLASS_AXIS_SIZE;
                 }
 
                 track.height = cumHeight;
@@ -202,7 +203,7 @@ function traverseAndCollectTrackInfo(
                     boundingBox: {
                         x: dx + cumWidth,
                         y: dy,
-                        width: track.width!,
+                        width: track.width,
                         height: cumHeight
                     },
                     layout: { x: 0, y: 0, w: 0, h: 0 } // Just put a dummy info here, this should be added after entire bounding box has been determined
@@ -211,7 +212,7 @@ function traverseAndCollectTrackInfo(
                 if (array[i + 1] && array[i + 1].overlayOnPreviousTrack) {
                     // do not add a height
                 } else {
-                    cumWidth += track.width!;
+                    cumWidth += track.width;
                     if (i !== array.length - 1) {
                         cumWidth += spec.spacing !== undefined ? spec.spacing : 0;
                     }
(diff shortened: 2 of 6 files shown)


src/core/defaults.ts
```diff
@@ -7,10 +7,9 @@ export const DEFAULT_SUBTITLE_HEIGHT = 20; // deprecated
 export const DEWFAULT_TITLE_PADDING_ON_TOP_AND_BOTTOM = 6;
 
 // default track size
-export const DEFAULT_TRACK_HEIGHT_LINEAR = 180;
-export const DEFAULT_TRACK_WIDTH_LINEAR = 360;
-export const DEFAULT_TRACK_HEIGHT_CIRCULAR = 400;
-export const DEFAULT_TRACK_WIDTH_CIRCULAR = 400;
```
sehilyi, line 13: These were unused

src/core/utils/bounding-box.ts
```diff
@@ -189,10 +189,10 @@ function traverseAndCollectTrackInfo(
 
         if (spec.orientation === 'vertical') {
             // This is a vertical view, so use the largest `height` of the tracks for this view.
-            cumHeight = Math.max(...tracks.map(d => d.height));
+            cumHeight = Math.max(...tracks.map(d => d.height!));
```
manzt: can we really assert that the width is defined? This seems dangerous....

```typescript
Math.max(1, 2, 3, undefined, 5, 6);
// NaN
```

manzt: It would be safer to filter the tracks by height, or create an object where all tracks have defined sizes.

src/core/utils/bounding-box.ts
```diff
@@ -189,10 +189,10 @@ function traverseAndCollectTrackInfo(
...
             tracks.forEach((track, i, array) => {
                 if (getNumOfXAxes([track]) === 1) {
-                    track.width += HIGLASS_AXIS_SIZE;
+                    track.width! += HIGLASS_AXIS_SIZE;
```
manzt: Same here. If these are truly defined, then our type system should show that somehow.

src/core/utils/bounding-box.ts
```diff
@@ -240,15 +240,15 @@ function traverseAndCollectTrackInfo(
...
+                        height: track.height!
                     },
                     layout: { x: 0, y: 0, w: 0, h: 0 } // Just put a dummy info here, this should be added after entire bounding box has been determined
                 });
 
                 if (array[i + 1] && array[i + 1].overlayOnPreviousTrack) {
                     // do not add a height
                 } else {
-                    cumHeight += track.height;
+                    cumHeight += track.height!;
```
manzt: many many non-null assertions that from the code in this context I cannot tell are "safe".

src/core/utils/spec-preprocess.ts
```diff
@@ -187,8 +188,12 @@ export function traverseToFixSpecDownstream(spec: GoslingSpec | SingleView, pare
...
         tracks.forEach((track, i, array) => {
             // If size not defined, set default ones
-            if (!track.width) track.width = DEFAULT_TRACK_WIDTH_LINEAR;
-            if (!track.height) track.height = DEFAULT_TRACK_HEIGHT_LINEAR;
+            if (!track.width) {
+                track.width = Is2DTrack(track) ? DEFAULT_TRACK_SIZE_2D : DEFAULT_TRACK_WIDTH_LINEAR;
+            }
+            if (!track.height) {
+                track.height = Is2DTrack(track) ? DEFAULT_TRACK_SIZE_2D : DEFAULT_TRACK_HEIGHT_LINEAR;
+            }
```
manzt, line 196: Ah I see, so here, we populate the spec with defined if they are missing. This is why a function that mutates the spec is unfortunate; we cannot capture that we've changed the type very easily. A function,

```typescript 
defineTrackSizes(spec); // mutates track sizes if missing
spec // spec still has same type (tracks could be missing)
```

vs

```typescript
const specWithCompleteTrackSizes = defineTrackSizes(spec); // does not mutate spec
```

src/core/utils/bounding-box.ts
manzt: At a minimum, this assertion 

```typescript
        const tracks = spec.tracks as (Omit<Track, 'width' | 'height'> & { width: number, height: number})[];
```

src/core/utils/spec-preprocess.ts
manzt: I also wonder if we should handle the case where it's a 2DTrack but only width _or_ height is defined. In that case, we probably want to use the defined dimension to infer the other? Actually not sure...

manzt (commented):
Looking good! I just a comment about non-null assertions and trying to avoid them generally.

src/core/utils/spec-preprocess.ts
sehilyi: `traverseToFixSpecDownstream` expands a spec by adding default values, and after calling this function, many optional properties are non-null. I had an idea of having another type of `GoslingSpec` that we internally use (something like `_GoslingSpec`), such as making some optional properties required and adding internally used properties (e.g., `startAngle`). But, I wasn't sure what would be the best and cleanest approach to create a new schema given that the `GoslingSpec` is fairly complex and deeply nested. At maximum, we could create a new type per each type we already have (e.g., `_SingleTrack`, `_Track`, `_CommonViewDef`, ..., `_GoslingSpec`).

sehilyi: > I also wonder if we should handle the case where it's a 2DTrack but only width or height is defined. In that case, we probably want to use the defined dimension to infer the other? Actually not sure...

I thought the same but was not sure either. That might be better if we expect that users will most frequently use matrices with the same `width` and `height`, and it does not confuse users. Being able to specify only one of them might be a bit handier.

manzt: > `traverseToFixSpecDownstream` expands a spec by adding default values, and after calling this function, many optional properties are non-null. 

We might be able to write a custom type helper for this. e.g,. `WithDefinedTracks<GoslingSpec>` and then just assert that the tracks are defined. But the `GoslingSpec` type is complicated recursive type so it might be challenging. I'd try to avoid maintaining `_GoslingSpec`, etc.

manzt:
I think the most practical solution for now is to try to have a single type-assertion with a comment (rather than many non-null assertions) in the same function. E.g,

```typescript
        // following `traverseToFixSpecDownstream`, the width and height of each track are gaurenteed to be defined
        const tracks = spec.tracks as (Track & { width: number, height: number})[];
```

However, ideally the `traverseToFixSpecDownstream` would have a type signature which captures all the transformations that we apply in that function:

```typescript

function traverseToFixSpecDownstream<TSpec extends GoslingSpec>(spec: TSpec): CompletedSpec<TSpec>;

```

PR #782, opened 2022-07-26
refactor: simplify workers

manzt:
There is a log of redundant caching of separate objects in both bam-worker.ts and vcf-worker.ts. This PR replaces uses of plain objects `{}` with ES6 `Map` and then creates a generic data source container (`DataSource<File, Options>`) for caching datasets. The data source takes care of translating chromsizes -> chrominfo and generating tilesetinfo for HiGlass.

### DIFF
--- a/src/data-fetchers/bam/bam-data-fetcher.ts
+++ b/src/data-fetchers/bam/bam-data-fetcher.ts
@@ -12,11 +12,8 @@ import { GET_CHROM_SIZES } from '../../core/utils/assembly';
 
 const DEBOUNCE_TIME = 200;
 
-// TODO: why both url/bamUrl & baiUrl/indexUrl
 interface DataConfig {
     url: string;
-    bamUrl?: string;
-    baiUrl?: string;
     indexUrl?: string;
     assembly: Assembly;
 }
@@ -40,16 +37,9 @@ class BamDataFetcher {
         this.uid = HGC.libraries.slugid.nice();
         this.toFetch = new Set();
 
-        const chromSizes = Object.entries(GET_CHROM_SIZES(dataConfig.assembly).size);
-
         this.worker = spawn<WorkerApi>(new Worker()).then(async worker => {
-            const bamUrl = dataConfig.bamUrl ?? dataConfig.url;
-            await worker.init(this.uid, {
-                ...dataConfig,
-                bamUrl,
-                baiUrl: dataConfig.baiUrl ?? dataConfig.indexUrl ?? `${bamUrl}.bai`,
-                chromSizes
-            });
+            const chromSizes = Object.entries(GET_CHROM_SIZES(dataConfig.assembly).size);
+            await worker.init(this.uid, dataConfig, chromSizes);
             return worker;
         });
     }
--- a/src/data-fetchers/bam/bam-worker.ts
+++ b/src/data-fetchers/bam/bam-worker.ts
@@ -1,4 +1,4 @@
-// This worker is heavily based on https://github.com/higlass/higlass-pileup/blob/master/src/bam-fetcher-worker.js
+// Adopted from https://github.com/higlass/higlass-pileup/blob/master/src/bam-fetcher-worker.js
 import { expose, Transfer } from 'threads/worker';
 import { BamFile as _BamFile } from '@gmod/bam';
 import QuickLRU from 'quick-lru';
--- a/src/data-fetchers/bam/bam-worker.ts
+++ b/src/data-fetchers/bam/bam-worker.ts
@@ -1,12 +1,13 @@
 // This worker is heavily based on https://github.com/higlass/higlass-pileup/blob/master/src/bam-fetcher-worker.js
 import { expose, Transfer } from 'threads/worker';
-import { BamFile } from '@gmod/bam';
+import { BamFile as _BamFile } from '@gmod/bam';
 import QuickLRU from 'quick-lru';
 
 import type { TilesetInfo } from '@higlass/types';
 import type { BamRecord } from '@gmod/bam';
 
-import { ExtendedChromInfo, RemoteFile, sizesToChromInfo } from '../utils';
+import { DataSource, RemoteFile } from '../utils';
+import type { ChromSizes } from '@gosling.schema';
 
 function parseMD(mdString: string, useCounts: true): { type: string; length: number }[];
 function parseMD(mdString: string, useCounts: false): { pos: number; base: string; length: 1; bamSeqShift: number }[];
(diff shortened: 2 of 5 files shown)


src/data-fetchers/bam/bam-worker.ts
```diff
@@ -1,12 +1,13 @@
 // This worker is heavily based on https://github.com/higlass/higlass-pileup/blob/master/src/bam-fetcher-worker.js
```
sehilyi: We could remove the line or change the comment?

```suggestion
// Adopted from https://github.com/higlass/higlass-pileup/blob/master/src/bam-fetcher-worker.js
```

src/data-fetchers/bam/bam-data-fetcher.ts
```diff
@@ -12,11 +12,8 @@ import { GET_CHROM_SIZES } from '../../core/utils/assembly';
...
 const DEBOUNCE_TIME = 200;
 
-// TODO: why both url/bamUrl & baiUrl/indexUrl
 interface DataConfig {
     url: string;
-    bamUrl?: string;
-    baiUrl?: string;
     indexUrl?: string;
     assembly: Assembly;
 }
```
sehilyi, line 19: We can instead reuse the BamData.

https://github.com/gosling-lang/gosling.js/blob/21edf06b4abdd2f6adcca8d323664147d876ca96/src/core/gosling.schema.ts#L1053-L1073

```diff
- import type { Assembly } from '../../core/gosling.schema';
+ import type { BamData } from '../../core/gosling.schema';
+ import { CommonDataConfig } from '../utils';
...

- interface DataConfig {
-    url: string;
-    indexUrl?: string;
-    assembly: Assembly;
- }
+ type DataConfig = BamData & CommonDataConfig; // ← has `assembly`
```

src/data-fetchers/bam/bam-worker.ts
```diff
@@ -228,78 +229,65 @@ const bamRecordToJson = (bamRecord: BamRecord, chrName: string, chrOffset: numbe
...
-    {
-        bamUrl,
-        baiUrl,
-        chromSizes,
-        loadMates = false,
-        maxInsertSize = 5000,
-        extractJunction = false,
-        junctionMinCoverage = 1
-    }: DataConfig
+    bam: { url: string; indexUrl?: string },
```
sehilyi, line 265: I think we can make `indexUrl` required since it is a required property in BamData ([code](https://github.com/gosling-lang/gosling.js/blob/21edf06b4abdd2f6adcca8d323664147d876ca96/src/core/gosling.schema.ts#L1060)). (This is my fault not to update this when I changed the grammar).

```suggestion
    bam: { url: string; indexUrl: string },
```

src/data-fetchers/bam/bam-worker.ts
```diff
@@ -228,78 +229,65 @@ const bamRecordToJson = (bamRecord: BamRecord, chrName: string, chrOffset: numbe
...
-            baiFilehandle: new RemoteFile(baiUrl)
-            // fetchSizeLimit: 500000000,
-            // chunkSizeLimit: 100000000 ,
-            // yieldThreadTime: 1000
-        });
-
-        // we have to fetch the header before we can fetch data
-        bamHeaders[bamUrl] = bamFiles[bamUrl].getHeader();
+    if (!bamFileCache.has(bam.url)) {
+        const bamFile = BamFile.fromUrl(bam.url, bam.indexUrl ?? `${bam.url}.bai`);
```
sehilyi, line 270: Likewise:

```suggestion
        const bamFile = BamFile.fromUrl(bam.url, bam.indexUrl);
```

src/data-fetchers/bam/bam-data-fetcher.ts
```diff
@@ -40,16 +37,9 @@ class BamDataFetcher {
...
         this.worker = spawn<WorkerApi>(new Worker()).then(async worker => {
-            const bamUrl = dataConfig.bamUrl ?? dataConfig.url;
-            await worker.init(this.uid, {
-                ...dataConfig,
-                bamUrl,
-                baiUrl: dataConfig.baiUrl ?? dataConfig.indexUrl ?? `${bamUrl}.bai`,
-                chromSizes
-            });
+            const chromSizes = Object.entries(GET_CHROM_SIZES(dataConfig.assembly).size);
+            await worker.init(this.uid, dataConfig, chromSizes);
```
sehilyi, line 42: I think the fourth parameter is missing which I think is the reason why certain values are not correctly generated in a BAM example:

Screen Shot 2022-07-26 at 16 31 55

I think we can either add the fourth parameter

```ts
await worker.init(this.uid, dataConfig, chromSizes, dataConfig);
```

or perhaps, we can combine the second and the fourth parameters? (Need to change the VCF file as well if we determine to change this).

```ts
await worker.init(this.uid, dataConfig, chromSizes);

// src/data-fetchers/bam/bam-worker.ts
import { DataConfig } from './bam-data-fetcher.ts';
...
const init = (
    uid: string,
    bam: DataConfig,
    chromSizes: ChromSizes
) => { ... }
```

src/data-fetchers/utils.ts
```diff
@@ -12,6 +12,16 @@ export type CommonDataConfig = {
     x1e?: string;
 };
 
+export class DataSource<File, Options> {
+    chromInfo: ExtendedChromInfo;
+    tilesetInfo: ReturnType<typeof tilesetInfoFromChromInfo>;
+
+    constructor(public file: File, chromSizes: ChromSizes, public options: Options) {
```
sehilyi, line 19: Didn't know about `public` key usage in the constructor, which looks handy!

sehilyi (commented):
Thanks, this is really great!! I can see codes are much more organized now. Also, good to have the cache working.

src/data-fetchers/bam/bam-data-fetcher.ts
manzt: I have addressed this in the changes. The workers all follow the following semantics when initializing the constructor:

```typescript
class DataFetcher {
  constructor(HGC: import("@higlass/types").HGC, config: Config) {
    // init
    const { url, indexUrl, assembly, ...options } = config;
    this.worker = spawn<WorkerApi>(new Worker()).then(async worker => {
      await worker.init(uid, { url, indexUrl }, chromSizes, options);
      return worker;
    })
  }
}

```

manzt:
It's important to note that the cache isn't exactly "working" as you'd expect. We still create a separate `Worker` for each `DataFetcher` instance, so their is no caching of shared resources yet.

PR #784, opened 2022-07-27
feat: run pre-commit hooks in parallel

manzt:
hmm the resulting `.json` files look weird. I wonder if running these scripts in parallel isn't stable?

### DIFF
--- a/package.json
+++ b/package.json
@@ -154,7 +154,7 @@
     },
     "husky": {
         "hooks": {
-            "pre-commit": "yarn changelog && yarn schema && yarn schema-higlass && yarn schema-theme && yarn schema-template && yarn format && git add .",
+            "pre-commit": "run-p changelog schema schema-higlass schema-theme schema-template format && git add .",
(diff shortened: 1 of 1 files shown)


schema/gosling.schema.json
```diff
@@ -415,9 +415,14 @@
         },
         "scaleOffset": {
           "description": "Whether to use offset of the domain proportionally. This is bound to brushes on the color legend. __Default__: `[0, 1]`",
-          "items": {
-            "type": "number"
-          },
+          "items": [
```
sehilyi: Maybe double-check to run `yarn`? I recently upgraded `ts-json-schema-generator`, and I wonder if the local version on your machine is the old one that makes the resulting schema different.

manzt:
Ah that did the trick! thanks @sehilyi

sehilyi (approved):

sehilyi:
Great! I will merge this now

PR #786, opened 2022-07-28
feat: reuse worker for same data-fetcher

manzt:
This PR adds a singleton worker per data fetcher, rather than creating a new worker each time. Ideally, this will allow us to reuse the tile cache for tracks which have the same data source.

I also refactored a bit of `bam-worker.ts` which was really difficult to follow. This implementation is more clear in my opinion, but there are still many odd bits translating between relative and absolute chromsizes. Ideally we figure out a way to deal with this across all data-fetchers.

### DIFF
--- a/src/data-fetchers/bam/bam-worker.ts
+++ b/src/data-fetchers/bam/bam-worker.ts
@@ -72,8 +72,6 @@ export type Segment = {
     start: number;
     end: number;
     md: string;
-    chrName: string;
-    chrOffset: number;
     cigar: string;
     mapq: string;
     strand: '+' | '-';
@@ -201,11 +199,7 @@ function getSubstitutions(segment: Segment, seq: string) {
     return substitutions;
 }
 
-/////////////////////////////////////////////////////
-/// End Chrominfo
-/////////////////////////////////////////////////////
-
-const bamRecordToJson = (bamRecord: BamRecord, chrName: string, chrOffset: number) => {
+const bamRecordToJson = (bamRecord: BamRecord) => {
     const seq = bamRecord.get('seq');
     const segment: Segment = {
         // if two segments have the same name but different id, they are paired reads.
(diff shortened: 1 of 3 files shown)


src/data-fetchers/bam/bam-worker.ts
```diff
@@ -285,129 +281,122 @@ const tilesetInfo = (uid: string) => {
...
     }
 
-    // flatten the array of promises so that it looks like we're getting one long list of value
-    return Promise.all(recordPromises).then(values => {
-        return values.flat();
+    return regions;
+}
+
+const tile = async (uid: string, z: number, x: number): Promise<JsonBamRecord[]> => {
+    const tileCacheKey = `${uid}.${z}.${x}`;
```
manzt, line 338: I'm not sure if this cache key is all that useful. The way that bam works there is a finite set of records at a single resolution. This caching strategy will redundantly cache records which we have already founds at a different zoom level. 

Ideally we could have some sort of interval cache, where if the interval (region) has already been visited, we know we have all the records for that region.

manzt:
This and the recent changes to the workers have started to make those visualizations feel sluggish. I'm not quite sure what is going on, but we should try to improve data fetching generally for our clientside loaders: this only makes the visualization a better user experience.

sehilyi:
Thanks for this update. I was not sure if this is ready to review, but one issue I found by running an editor on my machine is that the "Fetching... TILE1 TILE2 TILE3 ..." label on a BAM track does not seem to remove already fetched tiles from the list, making the label long after doing some zoom/pan interactions. I spend some time finding the source of this issue, but still uncertain. Somehow, `this.fetching` does not properly remove already-loaded tile IDs from the HiGlass (https://github.com/higlass/higlass/blob/38f0c4415f0595c3b9d685a754d6661dc9612f7c/app/scripts/TiledPixiTrack.js#L680).

Screen Shot 2022-08-04 at 14 43 54

manzt:
Hmm also can't figure out what is going on here with the HiGlass fetching... It might be overkill to implement this (and perhaps we should just rely on browser caching of requests for now).

PR #794, opened 2022-08-09
fix: allow custom chromnames in zoomTo api

manzt:
Currently the `zoomTo` API requires that the `position` string include a `chr` in the identifier. This prevents the API from being used with custom assemblies.

## Checklist

### DIFF
--- a/src/core/api.ts
+++ b/src/core/api.ts
@@ -94,10 +94,6 @@ export function createApi(
         unsubscribe,
         zoomTo: (viewId, position, padding = 0, duration = 1000) => {
             // Accepted input: 'chr1' or 'chr1:1-1000'
-            if (!position.includes('chr')) {
-                console.warn('Genomic interval you entered is not in a correct form.');
-                return;
-            }
             const assembly = getTrack(viewId)?.spec.assembly;
             const chromInfo = GET_CHROM_SIZES(assembly);
             const parsed = parseGenomicPosition(position);
--- a/src/core/utils/assembly.ts
+++ b/src/core/utils/assembly.ts
@@ -158,6 +158,12 @@ export function getChromTotalSize(chromSize: { [k: string]: number }) {
 
 export function parseGenomicPosition(position: string): { chromosome: string; start?: number; end?: number } {
     const [chromosome, intervalString] = position.split(':');
-    const [start, end] = intervalString.split('-').map(s => +s.replaceAll(',', ''));
-    return { chromosome, start, end };
+    if (intervalString) {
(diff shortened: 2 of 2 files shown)


manzt:
One question that comes to mind: why is the `id` for this API on the Track and not the *View*? 

By definition views are collections of tracks with shared genomic domains (and the first argument to `zoomTo` is `viewId`), but the id is for a track (calls `getTrack(viewId)` internally. Second, is it possible to define some default behavior where no view ID is provided? i.e., if no ID is provided... for _all_ views in the visualization try to navigate the provided position.

sehilyi:
The changes look good to me!

> Why is the id for this API on the Track and not the View?

That was mainly due to the way we handle Gosling tracks in the HiGlass, i.e., each Gosling track becomes a HiGlass view, so we use the ID to control the corresponding HiGlass view. I think it makes more sense to use the Gosling View IDs which will require us to add a view-level `"id"` and internally manage corresponding HiGlass view IDs for the APIs.

> is it possible to define some default behavior where no view ID is provided?

Yes, we should. I wonder what would be the good default behavior. We could give a better warning message w/o any navigation (like in HiGlass) or select the first or all visualization(s) by default to navigate to the position. If there is only one view, I think it would be handy if users can use an incorrect ID for the navigation. If there are multiple views (e.g., overview and detail views), it might be confusing if using an incorrect ID by mistake make all views navigate to a certain position.

sehilyi (approved):

manzt:
> That was mainly due to the way we handle Gosling tracks in the HiGlass, i.e., each Gosling track becomes a HiGlass view, so we use the ID to control the corresponding HiGlass view. I think it makes more sense to use the Gosling View IDs which will require us to add a view-level "id" and internally manage corresponding HiGlass view IDs for the APIs.

Ok that makes sense to me. It would be nice to "deprecate" track-level ids in favor of only view ids. This would help the mental model of tracks/views to be consistent (i.e., a `viewId` for a brush track would actually correspond to a view). 

> Yes, we should. I wonder what would be the good default behavior. 

Although I'm not a huge proponent of `null`, I think explicitly:

```typescript
api.zoomTo(null, position);
```

should have the behavior of trying to navigate _all_ views . I guess the only confusing part there is that `zoomTo(null)` navigates multiple views whereas `zoomTo('view-id')` only navigates one view. We could allow `zoomTo(['view-id', 'view-id2'])` to allow end users one api call to navigate multiple views.

> I think it would be handy if users can use an incorrect ID for the navigation.

I think I'd prefer that we just log an error if a view `id` is provided but we can't find a view with that name.

PR #795, opened 2022-08-09
refactor: use simple event emitter to listen for brush events

manzt:
This PR implements an idea for better composition of the different models in Gosling. Much of the node.js API is built around `Events` and objects extend from [`EventEmitter`](https://nodejs.org/api/events.html#class-eventemitter), where other objects can subscribe to various custom events emitted from objects:

```javascript

class Timer extends EventEmitter {
  constructor() {
    super();
    setInterval(() => {
      this.emit('tick')
    }, 100)
  }
}

let timer = new Timer();
timer.on('tick', () => console.log('tick!');
``` 

This is similar to the d3 API (except the `.on` methods are for various DOM events). Browsers don't have a native `EventEmitter`, so several libraries have made their own implementations:

- https://github.com/primus/eventemitter3
- https://github.com/EventEmitter2/EventEmitter2
- https://github.com/ai/nanoevents

This PR demonstrates the use of `nanoevents`, the most minimal and modern choice to show how to create custom emitters and add listeners. The library is tiny, and I think generally would encourage the ability to break apart large objects into smaller "emitting" components.

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,6 +76,7 @@
         "jspdf": "^2.3.1",
         "lodash-es": "^4.17.21",
         "mixwith": "^0.1.1",
+        "nanoevents": "^7.0.1",
         "pubsub-js": "^1.9.3",
         "quick-lru": "^6.1.1",
         "rbush": "^3.0.1",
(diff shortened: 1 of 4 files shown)


sehilyi (approved):
Nice. I didn't quite like the way I pass the function as a parameter of the constructor, and this approach looks better. I can imagine we can add more key--function pairs in `LinearBrushEvents` to enable listening to additional events, e.g., `'brushStart'`, etc.

PR #796, opened 2022-08-09
fix: support using custom chromosome names in genomic domains (`GenomicDomain`); fix `zoomTo()`

### LINKED ISSUE #793: `GenomicDomain` doesn't support custom chromsizes
The `Chromosome` type is hard-coded with human genome name literals (e.g. "chr1", "1", ...). In Gos, this causes an issue where you cannot initialize the domain for the visualization for the named chromosomes:

```python
gos.overlay(...).properties(
   assembly=[["NC_045512.2", 29903]],
   xDomain=gos.GenomicDomain(chromosome="NC_045512.2", interval=[1, 29903]),
 )
# Error!!! chromsome not one of "chr1" ... 
```

Chromosome should just be a `string`. I think we should try to avoid any magic with coercing chromosome names (e.g., (`name.replace("chr", "")`) since this gets quite complicated. Although as mentioned in #709, perhaps some fuzzy-ness is useful? In that case I would be extremely explicit about how coercion works and have a utility which does this (and can be tested carefully). i.e.,

```typescript
// returns valid key of assembly _or_ throws
function coerceChromIdentifier<
  Sizes extends Record<string, number>
>(name: string, sizes: Sizes): keyof Sizes {
  // exact match
  if (name in sizes) return name;
  
  let test;
  // strip chr prefix
  test = name.slice("chr".length)
  if (test in sizes) return test;

  // append chr prefix
  test = "chr" + name;
  if (test in sizes) return test;

  throw new Error("invalid identifier, could not be coerced");
}
```

sehilyi:
Fix #793

## Change List
 - Change the type of `chromosome` to `string` so that any custom chromosome names can be used.
 - Removes some magics about handling chromosome names which were really unnecessary.
 - Change all demo examples that used chromosome names without the "chr" prefix.

## Checklist

### DIFF
--- a/src/core/api.ts
+++ b/src/core/api.ts
@@ -3,7 +3,7 @@ import type { TrackMouseEventData } from '@gosling.schema';
 import type { HiGlassApi } from './higlass-component-wrapper';
 import type { HiGlassSpec } from './higlass.schema';
 import { subscribe, unsubscribe } from './pubsub';
-import { GET_CHROM_SIZES, parseGenomicPosition } from './utils/assembly';
+import { computeChromSizes, GenomicPositionHelper } from './utils/assembly';
 import type { CompleteThemeDeep } from './utils/theme';
 import { traverseViewsInViewConfig } from './utils/view-config';
 
@@ -95,25 +95,13 @@ export function createApi(
         zoomTo: (viewId, position, padding = 0, duration = 1000) => {
             // Accepted input: 'chr1' or 'chr1:1-1000'
             const assembly = getTrack(viewId)?.spec.assembly;
-            const chromInfo = GET_CHROM_SIZES(assembly);
-            const parsed = parseGenomicPosition(position);
-            const { chromosome: chr } = parsed;
-            let { start, end } = parsed;
-            if (!chr || !chromInfo.interval[chr]) {
-                console.warn('Chromosome name is not valid', chr);
-                return;
-            }
-            if (typeof start === 'undefined' || typeof end === 'undefined') {
-                // if only a chromosome name is specified, zoom to the extent of the chromosome
-                [start, end] = [0, chromInfo.size[chr]];
-            }
-            const chrOffset = chromInfo.interval[chr][0];
-            const relativeGenomicPositions = [start + chrOffset - padding, end + chrOffset + padding];
-            hg.api.zoomTo(viewId, ...relativeGenomicPositions, ...relativeGenomicPositions, duration);
+            const manager = GenomicPositionHelper.fromString(position);
+            const absCoordinates = manager.toAbsoluteCoordinates(assembly, padding);
+            hg.api.zoomTo(viewId, ...absCoordinates, ...absCoordinates, duration);
         },
         zoomToExtent: (viewId, duration = 1000) => {
             const assembly = getTrack(viewId)?.spec.assembly;
-            const [start, end] = [0, GET_CHROM_SIZES(assembly).total];
+            const [start, end] = [0, computeChromSizes(assembly).total];
             hg.api.zoomTo(viewId, start, end, start, end, duration);
         },
         zoomToGene: (viewId, gene, padding = 0, duration = 1000) => {
--- a/src/core/api.ts
+++ b/src/core/api.ts
@@ -99,7 +99,7 @@ export function createApi(
             const parsed = parseGenomicPosition(position);
             const { chromosome: chr } = parsed;
             let { start, end } = parsed;
-            if (!chr || chromInfo.interval[chr]) {
+            if (!chr || !chromInfo.interval[chr]) {
                 console.warn('Chromosome name is not valid', chr);
                 return;
             }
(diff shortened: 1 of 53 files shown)


src/core/api.ts
```diff
@@ -99,7 +99,7 @@ export function createApi(
             const parsed = parseGenomicPosition(position);
             const { chromosome: chr } = parsed;
             let { start, end } = parsed;
-            if (!chr || chromInfo.interval[chr]) {
+            if (!chr || !chromInfo.interval[chr]) {
```
sehilyi: I think this was a bug

sehilyi: I wonder if we can set up tests for the APIs not to break things.

src/core/gosling.schema.ts
```diff
@@ -766,18 +764,18 @@ export interface ChannelValue {
 export type AxisPosition = 'none' | 'top' | 'bottom' | 'left' | 'right';
 export type FieldType = 'genomic' | 'nominal' | 'quantitative';
 export type ValueExtent = string[] | number[];
-export type GenomicDomain = DomainInterval | DomainChrInterval | DomainChr | DomainGene;
```
sehilyi, line 769: This `DomainGene` was never supported, so let's remove this and add it back whenever we support it.

src/core/api.ts
manzt: > I wonder if we can set up tests for the APIs not to break things.

It would be challenging to test the API end-to-end since it requires the HiGlass API (running in the browser). I think breaking apart some of the logic here into separate functions which we can unit test separately would be best.


 i.e., a class which handles this behavior and we can test with different assemblies/chromInfos.

```typescript

class GenomicPosition {
  constructor(public chr: string, public start?: number, public end?: number) {}
  static fromString(s: string) { 
    const result = parseGenomicPosition(position);
    return new GenomicPosition(result.chromosome, result.start, result.end);
  }
  toAbsoluteCoordinates(assemby: Assembly, padding = 0): [number, number] {
     let info = GET_CHROM_SIZES(assembly);
     let size = info.size[this.chr];
     let interval = info.interval[this.chr];
     if (size === undefined || interval === undefined) {
       throw Error(`Chromosome name ${this.chr} is not valid`);
     }
     let { start, end } = this;
     if (start === undefined || end === undefined) {
         [start, end] = [0, size];
     }
     const offset = info.interval[0];
     return [start + offset - padding, end + offset + padding]
  }
}
```

manzt: then the zoomTo implementation would turn into:

```typescript

function zoomTo(viewId: string, positionString: string, padding = 0, duration = 1000) {
  const assembly = getTrack(viewId)?.spec.assembly;
  const position = GenomicPosition.fromString(positionString);
  let coordinates: [number, number];
  try {
    coordinates = position.toAbsoluteCoordinates(assembly);
  } catch (e) {
     console.warn("chromosome is not valid")
     return;
  }
  hg.api.zoomTo(viewId, ...coordinates, ...coordinates, duration);
}

```

manzt: and we could test the `GenomicPosition` class for a variety of positions and assemblies and assert the correct behavior

src/core/gosling.schema.ts
manzt: I agree with this decision. Genes are ultimately entities which correspond to intervals of genomic coordinates. There are many other genomic features which have this same property, and I don't see a motivating case to add specific support for "genes".

manzt: It should be simple enough to map _any_ table of features which are defined by to genomic coordinates (other columns in the table) with the current API.

src/core/utils/scales.ts
```diff
@@ -1,43 +1,32 @@
...
         return [
-            GET_CHROM_SIZES(assembly).interval[`chr${domain.chromosome}`][0] + 1,
-            GET_CHROM_SIZES(assembly).interval[`chr${domain.chromosome}`][1]
+            GET_CHROM_SIZES(assembly).interval[domain.chromosome][0] + 1,
+            GET_CHROM_SIZES(assembly).interval[domain.chromosome][1]
         ];
     } else if (IsDomainInterval(domain)) {
         return domain.interval;
     } else if (IsDomainChrInterval(domain)) {
-        const chrStart = GET_CHROM_SIZES(assembly).interval[`chr${domain.chromosome}`][0];
```
manzt: I would suggest calling `GET_CHROM_SIZES(assembly)` once in this function and then reusing the returned object. Remember that this function currently creates `URL` objects on every invocation.

I see this pattern a lot in the repo, I think that it stems from the `GET_CHROM_SIZES` name appearing to be a constant (where in fact it is a function that has non-trival overhead of computing). My I suggest renaming `GET_CHROM_SIZES` to `calculateChromsizes` or `computeChromsizes` so that this is more clear?

manzt (approved):

src/core/utils/scales.ts
sehilyi: Renaming the function is a good idea. I took the name `computeChromSizes`.

src/core/api.ts
sehilyi: I've updated the codes accordingly in [the last commit](https://github.com/gosling-lang/gosling.js/pull/796/commits/846da78d83480c44ac56aad687db1f90cd596d7f). I think we can expand this class and reuse it throughout the codes.

PR #801, opened 2022-08-11
refactor: register bam & vcf data-fetchers with HiGlass

manzt:
The `bam` and `vcf` data-loaders can now registered like others in `higlass-register` (since we don't need to intialize the Worker modules as of ). This PR refactors Gosling internals to register these data fetchers within `init()`.

## Checklist

### DIFF
--- a/src/core/init.ts
+++ b/src/core/init.ts
@@ -1,11 +1,11 @@
 // @ts-ignore
-import { default as higlassRegister } from 'higlass-register';
+import higlassRegister from 'higlass-register';
 // @ts-ignore
 import { TextTrack } from 'higlass-text';
 import { AxisTrack } from '../gosling-genomic-axis';
 import { BrushTrack } from '../gosling-brush';
-import { BigWigDataFetcher, CsvDataFetcher, JsonDataFetcher } from '../data-fetchers';
 import { GoslingTrack } from '../gosling-track/index';
+import * as dataFetchers from '../data-fetchers';
 
 let once = false;
 
@@ -67,12 +67,10 @@ export function init() {
     /**
      * Register data fetchers to HiGlassComponent
      */
-    higlassRegister({ dataFetcher: CsvDataFetcher, config: CsvDataFetcher.config }, { pluginType: 'dataFetcher' });
-    higlassRegister({ dataFetcher: JsonDataFetcher, config: JsonDataFetcher.config }, { pluginType: 'dataFetcher' });
-    higlassRegister(
-        { dataFetcher: BigWigDataFetcher, config: BigWigDataFetcher.config },
-        { pluginType: 'dataFetcher' }
-    );
+    for (const dataFetcher of Object.values(dataFetchers)) {
+        const { config } = dataFetcher;
+        higlassRegister({ dataFetcher, config }, { pluginType: 'dataFetcher' });
+    }
 
     once = true;
 }
--- a/src/gosling-track/gosling-track.ts
+++ b/src/gosling-track/gosling-track.ts
@@ -3,7 +3,15 @@ import PubSub from 'pubsub-js';
 import * as uuid from 'uuid';
 import { isEqual, sampleSize, uniqBy } from 'lodash-es';
 import type { ScaleLinear } from 'd3-scale';
-import type { SingleTrack, OverlaidTrack, Datum, EventStyle, GenomicPosition, Assembly } from '@gosling.schema';
+import type {
+    SingleTrack,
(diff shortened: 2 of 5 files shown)


src/core/init.ts
```diff
@@ -1,11 +1,11 @@
 // @ts-ignore
-import { default as higlassRegister } from 'higlass-register';
+import higlassRegister from 'higlass-register';
```
manzt, line 2: No need for `as` with default imports: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import#syntax

src/gosling-track/gosling-track.ts
```diff
@@ -63,6 +66,18 @@ interface GoslingTrackOption {
...
+    return typeof x === 'object' || x !== null;
+}
+
+function isTabularDataFetcher(dataFetcher: unknown): dataFetcher is BamDataFetcher | VcfDataFetcher {
+    return isObject(dataFetcher) && 'getTabularData' in dataFetcher;
+}
+
+function hasDataTransform(spec: SingleTrack | OverlaidTrack, type: DataTransform['type']) {
+    return (spec.dataTransform ?? []).some(d => d.type === type);
+}
```
sehilyi: Maybe move these util functions to another file like `gosling.schema.guards.ts`? (These functions are not exactly guards for the 'schema', but the file contains similar functions.)

src/gosling-track/gosling-track.ts
```diff
@@ -63,6 +66,18 @@ interface GoslingTrackOption {
     theme: CompleteThemeDeep;
 }
 
+function isObject(x: unknown): x is Record<PropertyKey, unknown> {
+    return typeof x === 'object' || x !== null;
```
sehilyi: Is the `||` correct or should it be `&&`? 

(Btw, didn't know that JS treats `null` as `object` which seems odd to me)

src/gosling-track/gosling-track.ts
```diff
@@ -63,6 +66,18 @@ interface GoslingTrackOption {
...
+}
+
+function isTabularDataFetcher(dataFetcher: unknown): dataFetcher is BamDataFetcher | VcfDataFetcher {
+    return isObject(dataFetcher) && 'getTabularData' in dataFetcher;
```
sehilyi: I think, later, we need to ensure that one implements `getTabularData` in worker-based data fetchers when we type out the classes (e.g., making `getTabularData` a required function).

src/gosling-track/gosling-track.ts
```diff
@@ -96,20 +111,6 @@ function GoslingTrack(HGC: import('@higlass/types').HGC, ...args: any[]): any {
...
-                try {
-                    if (options.spec.data?.type === 'bam') {
-                        context.dataFetcher = new BamDataFetcher(HGC, context.dataConfig);
-                    } else if (options.spec.data?.type === 'vcf') {
-                        context.dataFetcher = new VcfDataFetcher(HGC, context.dataConfig);
-                    }
-                } catch (e) {
-                    console.warn('Error loading tabular data-fetcher', e);
-                }
-            }
```
sehilyi, line 111: I wonder I why did not register these using `higlassRegister` before...🤷‍♂️

sehilyi (commented):
Nice 👍!

Other than a question regarding the logical operation of `isObject`, I think this is ready to merge.

src/gosling-track/gosling-track.ts
manzt: Ah yes, _very_ good catch!!!

> (Btw, didn't know that JS treats null as object which seems odd to me)

Yes, this is something that TypeScript enforces:

image

src/gosling-track/gosling-track.ts
manzt: I think it was because you needed to instantiate the `Worker` for each data loader. This was done in `gosling-track` constructor previously and the worker was stored on the track and accessed directly. Now the worker is hidden (within the data-fetcher) and we added an extra `getTabularData` fn which we inspect.

src/gosling-track/gosling-track.ts
manzt: Yes. I'd write an interface, and then have each class `implement` that interface:

```typescript
interface TabularDataFetcher<Tile> {
	getTabularData(tileIds: string[]): Promise<Tile[]>;
}
```

```typescript
class BamDataFetcher implements TabularDataFetcher<Segment | SegmentWithMate | Junction> { ... }
```

PR #802, opened 2022-08-11
fix: fix `displace` transform on non-higlass data

### LINKED ISSUE #800: Pileup transformation with non-higlass data throws an error
It looks like the source of the error is when combining tiles before applying the pileup transformation.

Source of the error ([code](https://github.com/gosling-lang/gosling.js/blob/a52d969d1fdde22ed97d7d9a1d97e46d2ef5fcc8/src/gosling-track/gosling-track.ts#L780-L786)):
```ts
// combineAllTilesIfNeeded()
tiles.forEach((t: any, i: number) => {
    // Combine data
    newData = [...newData, ...t.tileData]; // Uncaught (in promise) TypeError: c.tileData is not iterable

    // Flag to force using only one tile
    t.mergedToAnotherTile = i !== 0;
});
```

`tileData` is only available for the higlass data.

Minimal example:

```js
{
  "tracks": [
    {
      "data": {
        "type": "json",
        "values": [
          {"c": "chr1", "s": 100, "e": 200},
          {"c": "chr1", "s": 150, "e": 200}
        ],
        "chromosomeField": "c",
        "genomicFields": ["s", "e"]
      },
      "dataTransform": [
        {
          "type": "displace",
          "method": "pile",
          "boundingBox": {"startField": "s", "endField": "e"},
          "newField": "r"
        }
      ],
      "mark": "rect",
      "x": {"field": "s", "type": "genomic", "domain": {"interval": [1, 300]}},
      "xe": {"field": "e", "type": "genomic"},
      "row": {"field": "r", "type": "nominal"},
      "color": {"value": "red"},
      "stroke": {"value": "black"},
      "strokeWidth": {"value": 1}
    }
  ]
}
```

sehilyi:
Fix #800

## Change List
 - Fix a bug that `displace` transform does not work on non-higlass datasets (#800)
Screen Shot 2022-08-19 at 16 47 07

```ts
{
  "tracks": [
    {
      "data": {
        "type": "json",
        "values": [
          {"c": "chr1", "s": 150, "e": 200},
          {"c": "chr1", "s": 100, "e": 200},
          {"c": "chr1", "s": 125, "e": 160},
          {"c": "chr1", "s": 100, "e": 160}
        ],
        "chromosomeField": "c",
        "genomicFields": ["s", "e"]
      },
      "dataTransform": [
        {
          "type": "displace",
          "method": "pile",
          "boundingBox": {"startField": "s", "endField": "e"},
          "newField": "r"
        }
      ],
      "mark": "rect",
      "x": {"field": "s", "type": "genomic", "domain": {"interval": [1, 300]}},
      "xe": {"field": "e", "type": "genomic"},
      "row": {"field": "r", "type": "nominal"},
      "color": {"value": "red"},
      "stroke": {"value": "black"},
      "strokeWidth": {"value": 1}
    }
  ]
}
```

 - Type out `tile` and `tileData` more and remove a lot of `any` usage from `gosling-track.ts`
 - Remove `tile.gos` object entirely which was largely redundant with `tileData`. Instead, we are using a separate object that stores processed tabular data per tile:

```ts
private processedTileInfo: Record<string, ProcessedTileInfo>; // key is the `tile.tileId`

interface ProcessedTileInfo {
    mergedToAnotherTile: boolean;
    goslingModels: GoslingTrackModel[];
    tabularData: Record<string, any>[];
    tabularDataTransformed: Record<string, any>[];
}
```

## Checklist

### DIFF
--- a/src/core/mark/bar.test.ts
+++ b/src/core/mark/bar.test.ts
@@ -3,6 +3,7 @@ import type { SingleTrack } from '@gosling.schema';
 import { GoslingTrackModel } from '../gosling-track-model';
 import { getTheme } from '../utils/theme';
 import { drawBar } from './bar';
+import { Tile } from '@higlass/services';
 
 describe('mark:= bar', () => {
     it('x:=G, y:=V, ye:=V', () => {
@@ -25,7 +26,7 @@ describe('mark:= bar', () => {
             }
         };
         const tileMock = { graphics: new PIXI.Graphics(), tileData: {} };
-        drawBar(trackMock, tileMock as any, model);
+        drawBar(trackMock, tileMock as Tile, model);
         // Should render all data (https://github.com/gosling-lang/gosling.js/pull/791)
         expect(model.getMouseEventModel().size()).toEqual(d.length);
     });
--- a/src/core/mark/bar.test.ts
+++ b/src/core/mark/bar.test.ts
@@ -3,7 +3,7 @@ import type { SingleTrack } from '@gosling.schema';
 import { GoslingTrackModel } from '../gosling-track-model';
 import { getTheme } from '../utils/theme';
 import { drawBar } from './bar';
-import { Tile } from '@higlass/services';
+import type { Tile } from '@higlass/services';
 
 describe('mark:= bar', () => {
     it('x:=G, y:=V, ye:=V', () => {
--- a/src/core/mark/bar.test.ts
+++ b/src/core/mark/bar.test.ts
@@ -24,8 +24,8 @@ describe('mark:= bar', () => {
                 return { tileX: null, tileWidth: null };
             }
         };
-        const tileMock = { graphics: new PIXI.Graphics(), gos: {} };
-        drawBar(trackMock, tileMock, model);
+        const tileMock = { graphics: new PIXI.Graphics(), tileData: {} };
+        drawBar(trackMock, tileMock as any, model);
         // Should render all data (https://github.com/gosling-lang/gosling.js/pull/791)
         expect(model.getMouseEventModel().size()).toEqual(d.length);
     });
--- a/src/gosling-track/gosling-track.ts
+++ b/src/gosling-track/gosling-track.ts
@@ -68,10 +68,11 @@ interface GoslingTile extends Tile {
     mergedToAnotherTile: boolean;
     goslingModels?: GoslingTrackModel[];
     gos: Record<string, any> & {
+        tabularData?: Record<string, any>[];
         raw?: Datum[];
     };
     tileData: TileData & {
-        tabularData?: Record<string, any>;
+        tabularData?: Record<string, any>[];
     };
 }
 
@@ -855,7 +856,7 @@ function GoslingTrack(HGC: import('@higlass/types').HGC, ...args: any[]): any {
          * Construct tabular data from a higlass tileset and a gosling track model.
          * Return the generated gosling track model.
          */
-        preprocessTile(tile: any) {
+        preprocessTile(tile: GoslingTile) {
             if (tile.mergedToAnotherTile) {
                 tile.goslingModels = [];
                 return;
@@ -902,7 +903,7 @@ function GoslingTrack(HGC: import('@higlass/types').HGC, ...args: any[]): any {
                     });
                 }
 
-                tile.gos.tabularDataFiltered = Array.from(tile.gos.tabularData);
+                tile.gos.tabularDataFiltered = Array.from(tile.gos.tabularData!);
 
                 /*
                  * Data Transformation applied to each of the overlaid tracks.
@@ -993,13 +994,13 @@ function GoslingTrack(HGC: import('@higlass/types').HGC, ...args: any[]): any {
                 const gm = new GoslingTrackModel(resolved, tile.gos.tabularDataFiltered, this.options.theme);
 
                 // Add a track model to the tile object
-                tile.goslingModels.push(gm);
+                tile.goslingModels!.push(gm);
             });
 
             return tile.goslingModels;
         }
 
-        getIndicesOfVisibleDataInTile(tile: any) {
+        getIndicesOfVisibleDataInTile(tile: GoslingTile) {
             const visible = this._xScale.range();
 
             if (!this.tilesetInfo) return [null, null];
--- a/src/gosling-track/gosling-track.ts
+++ b/src/gosling-track/gosling-track.ts
@@ -34,6 +34,7 @@ import { flatArrayToPairArray } from '../core/utils/array';
 import { BamDataFetcher, VcfDataFetcher } from '../data-fetchers';
 import { LinearBrushModel } from '../gosling-brush/linear-brush-model';
 import { isPointInsideDonutSlice } from '../gosling-mouse-event/polygon';
+import type { Tile, TileData } from '@higlass/services';
 
 // Set `true` to print in what order each function is called
 export const PRINT_RENDERING_CYCLE = false;
@@ -61,6 +62,17 @@ interface GoslingTrackOption {
     spec: SingleTrack | OverlaidTrack;
     showMousePosition: boolean;
     theme: CompleteThemeDeep;
+}
+
+interface GoslingTile extends Tile {
+    mergedToAnotherTile: boolean;
+    goslingModels?: GoslingTrackModel[];
+    gos: Record<string, any> & {
+        raw?: Datum[];
+    };
+    tileData: TileData & {
+        tabularData?: Record<string, any>;
+    };
 }
 
 /**
@@ -265,7 +277,7 @@ function GoslingTrack(HGC: import('@higlass/types').HGC, ...args: any[]): any {
          * Do whatever is necessary before rendering a new tile. This function is called from `receivedTiles()`.
          * (Refer to https://github.com/higlass/higlass/blob/54f5aae61d3474f9e868621228270f0c90ef9343/app/scripts/HorizontalLine1DPixiTrack.js#L50)
          */
-        initTile(tile: any) {
+        initTile(tile: GoslingTile) {
             if (PRINT_RENDERING_CYCLE) console.warn('initTile(tile)');
 
             // super.initTile(tile); // This calls `drawTile()`
@@ -280,7 +292,7 @@ function GoslingTrack(HGC: import('@higlass/types').HGC, ...args: any[]): any {
         /**
          * Display a tile upon receiving a new one or when explicitly called by a developer, e.g., calling `this.draw()`
          */
-        drawTile(tile: any) {
+        drawTile(tile: GoslingTile) {
             if (PRINT_RENDERING_CYCLE) console.warn('drawTile(tile)');
 
             tile.drawnAtScale = this._xScale.copy(); // being used in `super.draw()`
@@ -496,8 +508,8 @@ function GoslingTrack(HGC: import('@higlass/types').HGC, ...args: any[]): any {
         /**
          * Return the set of all tiles which are both visible and fetched.
          */
-        visibleAndFetchedTiles() {
-            return this.visibleAndFetchedIds().map((x: any) => this.fetchedTiles[x]);
+        visibleAndFetchedTiles(): GoslingTile[] {
+            return this.visibleAndFetchedIds().map(x => this.fetchedTiles[x]);
         }
 
         // !! This is called in the constructor, `super(context, options)`. So be aware to use variables that is prepared.
@@ -638,13 +650,13 @@ function GoslingTrack(HGC: import('@higlass/types').HGC, ...args: any[]): any {
         /**
          * Convert tile positions to tile IDs
          */
-        tilesToId(xTiles: any[], yTiles: any[], zoomLevel: any) {
+        tilesToId(xTiles: number[], yTiles: number[], zoomLevel: number) {
             if (xTiles && !yTiles) {
                 // this means only the `x` axis is being used
                 return xTiles.map(x => [zoomLevel, x]);
             } else {
                 // this means both `x` and `y` axes are being used together
-                const tiles: any = [];
+                const tiles: [number, number, number][] = [];
                 xTiles.forEach(x => yTiles.forEach(y => tiles.push([zoomLevel, x, y])));
                 return tiles;
             }
@@ -693,7 +705,7 @@ function GoslingTrack(HGC: import('@higlass/types').HGC, ...args: any[]): any {
 
             this.tileSize = this.tilesetInfo?.tile_size ?? 1024;
 
-            tiles.forEach((t: any) => {
+            tiles.forEach(t => {
                 // A new object to store all datasets
                 t.gos = {};
 
@@ -702,7 +714,7 @@ function GoslingTrack(HGC: import('@higlass/types').HGC, ...args: any[]): any {
 
                 // Store objects first
                 keys.forEach(k => {
-                    t.gos[k] = t.tileData[k];
+                    t.gos[k] = t.tileData[k as keyof TileData];
                 });
 
                 // Store raw data
@@ -726,10 +738,8 @@ function GoslingTrack(HGC: import('@higlass/types').HGC, ...args: any[]): any {
         }
 
         shareScaleOffsetAcrossTracksAndTiles(scaleOffset: [number, number], channelKey: 'color' | 'stroke') {
-            const models: GoslingTrackModel[] = [];
-            this.visibleAndFetchedTiles().forEach((tile: any) => {
-                models.push(...tile.goslingModels);
-            });
+            const visibleTiles = this.visibleAndFetchedTiles();
+            const models = visibleTiles.flatMap(d => d.goslingModels ?? []);
             models.forEach(d => {
                 const channel = d.spec()[channelKey];
                 if (IsChannelDeep(channel)) {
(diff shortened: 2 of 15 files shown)


src/gosling-track/gosling-track.ts
```diff
@@ -63,6 +64,17 @@ interface GoslingTrackOption {
...
+    mergedToAnotherTile: boolean;
+    goslingModels?: GoslingTrackModel[];
+    gos: Record<string, any> & {
+        raw?: Datum[];
+    };
+    tileData: TileData & {
+        tabularData?: Record<string, any>;
+    };
+}
+
```
sehilyi, line 77: I think we should unify the use of `raw`, `tileData`, and `tabularData`.

src/gosling-track/gosling-track.ts
```diff
@@ -63,6 +64,18 @@ interface GoslingTrackOption {
...
+    goslingModels?: GoslingTrackModel[];
+    gos: Record<string, any> & {
+        tabularData?: Record<string, any>[];
+        raw?: Datum[];
+    };
+    tileData: TileData & {
+        tabularData?: Record<string, any>[];
+    };
+}
+
```
sehilyi: We should really unify the use of `tileData`, `tileData.tabularData`, `gos.raw`, and `gos.tabularData`.

sehilyi: We can remove the `gos` object and just use the `tileData`. I will do this in a separate PR.

src/missing-types.d.ts
```diff
@@ -52,12 +52,15 @@ declare module '@higlass/services' {
               bins_per_dimension?: number;
           }
     );
-    type TileData = {
+    export type TileData = Array<Record<string, string | number>> & {
```
manzt: Is this really  `Array<T> & { tileId: string, ... }`? I.e.,

```ts
declare const tile: TileData;

tile[0] // Record<string, string | number>
tile[1] // Record<string, string | number>
tile.tileId // string
```

When I `this.fetchedTiles` I don't see an array

src/missing-types.d.ts
```diff
@@ -52,12 +52,15 @@ declare module '@higlass/services' {
...
         dense: number[];
         shape: number[];
         tilePos: unknown[];
+        zoomLevel: number;
     };
-    type Tile = {
+    export type Tile = {
+        drawnAtScale: Scale;
+        graphics: PIXI.Graphics;
```
manzt, line 82: These types are missing. Add,

```typescript
    import type * as PIXI from 'pixi.js';

    type Scale = ScaleContinuousNumeric<number, number>;
```

to the top of `@higlass/services'

src/missing-types.d.ts
manzt: Screen Shot 2022-08-12 at 4 34 13 PM

manzt (commented):
I really like the idea of adding types for these data. However, I'm not sure I understand `GoslingTile` compared to HiGlass's `Tile`. It seems like we are mutating the `tile` returned from the datafetchers but I don't see where this occurs. I would avoid mutation if at all possible – this is a very easy way trick the type system and get weird bugs. 

I'd rather cautiously work with an untyped API than rely on types which are incorrect.

The main issue is that the current implementations "lies" to our type system. For example, `this.visibleAndFetchedTiles()` has type:

```typescript
visibleAndFetchedTiles(): GoslingTile[];
```

which declares that this method will always return an array of _complete_ `GoslingTile`s. This is incorrect. It isn't until we _mutate_ the individual tiles stored on `this.fetchedTiles` that they are transformed into `GoslingTile` (in `reorganizeTileInfo`).

```typescript
        reorganizeTileInfo() {
            // tiles: GoslingTile[] which is not correct the first time this executes. 
            const tiles = this.visibleAndFetchedTiles(); 

            this.tileSize = this.tilesetInfo?.tile_size ?? 1024;

	       // HERE the tileData is mutated!!!
            tiles.forEach(t => {
                // we are allowed to make this mutation from the type system since we've lied to TS
                // and said that `t.gos` exists.
                t.gos = {};

                const keys = Object.keys(t.tileData).filter(d => !+d && d !== '0'); // ignore array indexes

                keys.forEach(k => {
                    t.gos[k] = t.tileData[k as keyof TileData];
                });

                t.gos.raw = Array.from(t.tileData);
            });
        }
```

The first time this function runs it mutates each object in `this.fetchedTiles` to include the `gos` field. However, if I were to try to access `gos` the first time this function executed:

```ts
    reorganizeTileInfo() {
            const tiles = this.visibleAndFetchedTiles(); 
            tiles[0].gos.tabularData // runtime error! tabularData does not exist on type undefined
        }
```

We get a runtime error. The reason TypeScript doesn't catch this is because it believes us that the type is _always_ `GoslingTile`, since `this.fetchedTiles` is of type `any` and can be coerced into anything that we say it should be. It is trivial to see that `GoslingTile` doesn't actually exist if you prevent the tile mutation code from running and inspecting the console.

```typescript
    reorganizeTileInfo() {
            const tiles = this.visibleAndFetchedTiles(); // not actually `GoslingTile` until this function runs!
            console.log(tiles)
            return; // return early without mutatin
        }
```

TL;DR - currently we depend on certain functions to execute before others in ways that are hidden from our type system. This is dangerous and can lead to bugs which are very hard to track down (since we are likely to trust the type system). I'm not familiar with which pieces of the data are strictly necessary for Gosling and which are for HiGlass, but I hope there is a way to avoid mutation of tiles. 

I think ideally we add a type for `this.fetchedTiles` which is not `any` and models exactly what we can expect to be there from HiGlass. Then if we need to use that data in anyway, typescript will force us to handle conversion into Gosling-specific data structures.

src/missing-types.d.ts
sehilyi: Yes, it is. It's an 'array-like object' for the sparse data, e.g., beddb.

Screen Shot 2022-08-17 at 12 04 48

sehilyi:
Thank you for the detailed explanations. My type definition was incorrect since `gos` was required instead of optional, and making it optional gives more type errors. 

For context, these are the main data objects that we currently use.

- `tile.tabularData`: Tabular data received directly from Gosling data fetchers
- `tile.tileData`: Sparse HiGlass data (e.g., beddb)
- `tile.tileData.dense`: An array of numbers (i.e., `number[]`) of dense HiGlass data (e.g., multivec).

We currently store data from these different sources to `tile.gos.tabularData` for unified rendering after calling `getTabularData()`.

I think we can entirely remove the `tile.gos` object (which was largely redundant with `tile.tileData`) and instead add a new variable in `GoslingTrackClass` that stores the processed tabular data:

```ts
tabularizedDataByTileId: Record<string, Record<string, Datum[]>>;
```

We can use `tile.tileId` as a key.

src/missing-types.d.ts
sehilyi: https://github.com/higlass/higlass/blob/38f0c4415f0595c3b9d685a754d6661dc9612f7c/app/scripts/TiledPixiTrack.js#L653

manzt: Ah wow... In that case it might be good to have the tile typed with a discriminate union:

```typescript
interface TileDataBase {
	server: string;
	tileId: string;
	tilePos: number[];
	tilePositionId: string;
	tilesetUid: string;
	zoomLevel: number;
}

inferface DenseTileData extends TileDataBase {
	dense: number[];
}

type IntervalTile = { xStart: number, xEnd: number, chrOffset: number, importance: number, uid: string } 
type IntervalTileData = Array<IntervalTile> & TileDataBase;

type TileData = DenseTileData | IntervalTileData;
```

sehilyi:
@manzt I've made further changes to (1) remove `tile.gos` entirely, (2) do not mutation `tile` at all and instead use a separate variable `private processedTileInfo: Record<string, ProcessedTileInfo>;` which stores processed tabular data of tiles, (3) fix the `displace` issue (#800).

Would you be able to review again this PR to see if further updates make sense?

The most important changes are in `src/gosling-track/gosling-track.ts` and `src/missing-types.d.ts`.

src/core/mark/bar.test.ts
```diff
@@ -24,8 +24,8 @@ describe('mark:= bar', () => {
                 return { tileX: null, tileWidth: null };
             }
         };
-        const tileMock = { graphics: new PIXI.Graphics(), gos: {} };
-        drawBar(trackMock, tileMock, model);
+        const tileMock = { graphics: new PIXI.Graphics(), tileData: {} };
+        drawBar(trackMock, tileMock as any, model);
```
manzt: Perhaps `as Tile`? Just so we are being explicit.

src/missing-types.d.ts
```diff
@@ -52,14 +56,36 @@ declare module '@higlass/services' {
...
+        xStart: number;
+        xEnd: number;
+        chrOffset: number;
+        importance: number;
+        uid: string;
+        fields: string[];
     };
-    type Tile = {
+    interface TabularTileData extends TileDataBase {
+        tabularData: Record<string, any>[];
```
manzt: `TabularTileData` is something that we are introducing, correct? If so, this type should live outside of HiGlass.

src/gosling-track/gosling-track.ts
```diff
@@ -65,6 +66,21 @@ interface GoslingTrackOption {
     theme: CompleteThemeDeep;
 }
 
+interface ProcessedTileInfo {
+    mergedToAnotherTile: boolean;
+    /** Single tile can contain multiple gosling models if multiple tracks are superposed */
+    goslingModels: GoslingTrackModel[];
+    tabularData: Record<string, any>[];
+    tabularDataTransformed: Record<string, any>[];
```
manzt:
```suggestion
    tabularData: Datum[];
    tabularDataTransformed: Datum[];
```

Is this correct?

src/gosling-track/gosling-track.ts
```diff
@@ -764,38 +760,37 @@ function GoslingTrack(HGC: import('@higlass/types').HGC, ...args: any[]): any {
...
-                // Flag to force using only one tile
-                t.mergedToAnotherTile = i !== 0;
+                    // Flag to force using only one tile
+                    tileInfo.mergedToAnotherTile = i !== 0;
+                }
             });
 
-            tiles[0].gos.raw = newData;
+            const firstTileInfo = this.processedTileInfo[tiles[0].tileId];
+            firstTileInfo.tabularData = merged;
```
manzt: I'm a little confused about the logic here. So we are mapping over each higlass tile (regardless of tileId) and merging the results to a single array?

```
merged = [...tabularData0, ...tabularData1, ...tabularData2, ...];
```

but then using the `tileId` of the _first_ tile to keep track of that data? So the result is that in order to access the "merged" data we need to map over `processedTileInfo` and find an object where `mergedToAnotherTile === false` because  `tabularData` is the concatenated array?

src/gosling-track/gosling-track.ts
```diff
@@ -834,100 +829,112 @@ function GoslingTrack(HGC: import('@higlass/types').HGC, ...args: any[]): any {
...
-                tile.goslingModels = [];
-                return;
-            }
-
-            if (tile.goslingModels && tile.goslingModels.length !== 0) {
-                // already have the gosling models constructed
-                return tile.goslingModels;
+        preprocessTile(tile: Tile, force = false): GoslingTrackModel[] {
+            if (force || !this.processedTileInfo[tile.tileId]) {
+                this.processedTileInfo[tile.tileId] = { ...JSON.parse(JSON.stringify(DEFAULT_PROCESSED_TILE_INFO)) };
```
manzt: Hmm... Why not have a function:

```typescript
function initProcessedTileInfo() {
    return {
        mergedToAnotherTile: false,
        goslingModels: [],
        tabularData: [],
        tabularDataTransformed: []
   }
}

// ...
this.processedTileInfo[tile.tileId] = initProcessedTileInfo();
```

src/gosling-track/gosling-track.ts
```diff
@@ -834,100 +829,112 @@ function GoslingTrack(HGC: import('@higlass/types').HGC, ...args: any[]): any {
...
+                // we do not have this information ready yet, i.e., cannot calculate `tileX`
+                return [];
             }
 
-            // Single tile can contain multiple gosling models if multiple tracks are superposed.
-            tile.goslingModels = [];
-
-            const spec = JSON.parse(JSON.stringify(this.options.spec));
+            const tileInfo = this.processedTileInfo[tile.tileId];
+            const specCopy = JSON.parse(JSON.stringify(this.options.spec));
```
manzt:
```suggestion
            const specCopy = structuredClone(this.options.spec);
```

src/gosling-track/gosling-track.ts
```diff
@@ -836,100 +832,112 @@ function GoslingTrack(HGC: import('@higlass/types').HGC, ...args: any[]): any {
...
+            /* apply data transforms */
+            resolvedTracks.forEach(resolvedSpec => {
+                tileInfo.tabularDataTransformed = Array.from(tileInfo.tabularData);
 
                 /*
                  * Data Transformation applied to each of the overlaid tracks.
                  */
-                if (resolved.dataTransform) {
-                    resolved.dataTransform.forEach(t => {
+                if (resolvedSpec.dataTransform) {
```
manzt: could lose a level of indentation with `resolvedSpec.dataTransform?.forEach...`

src/gosling-track/gosling-track.ts
```diff
@@ -836,100 +832,112 @@ function GoslingTrack(HGC: import('@higlass/types').HGC, ...args: any[]): any {
...
                         tileY,
                         tileWidth,
                         tileHeight,
                         tileSize: this.tileSize
                     });
+                    if (tabularData) {
+                        tileInfo.tabularData = tabularData;
+                    }
                 }
+            });
```
manzt: Am I correct that `tileInfo.tabularData` is continuously re-written if `resolvedTracks.length >= 1`?

src/gosling-track/gosling-track.ts
```diff
@@ -836,100 +832,112 @@ function GoslingTrack(HGC: import('@higlass/types').HGC, ...args: any[]): any {
...
 
-                tile.gos.tabularDataFiltered = Array.from(tile.gos.tabularData);
+            this.combineAllTilesIfNeeded();
+
+            /* apply data transforms */
+            resolvedTracks.forEach(resolvedSpec => {
+                tileInfo.tabularDataTransformed = Array.from(tileInfo.tabularData);
```
manzt: Likewise here. We mutate `tileInfo` for each resolvedSpec?

manzt (commented):
Ok great! I think this is getting very close, just a few comments.

One thing I've been thinking about is that we are using a lot of very basic data types (simple objects) to model our internal tile data and then relying on flags or various properties to be present to decide what to do.

This means there is a lot of logic in `gosling-track.ts` to deal with synchronizing the `processedTileInfo` and it's current state and actions which must occur before creating `GoslingTrackModels`. For now I think it's better than before and about ready to merge, but one thing that would really help design this system more robustly would be to draw or write-up the data-flow in Gosling. 

Currently things are so tightly coupled to HiGlass that its challenging to understand the relationship between each other. We end up dealing with a lot of tileIds, and I wonder if some of that could be abstracted.

src/missing-types.d.ts
sehilyi: Yes, it is. I will move it to `gosling-track.ts`

src/gosling-track/gosling-track.ts
sehilyi: Yes, that is correct – When `displace` transform is used, we combine all tabular data into the first tile so that we apply the transformation at once. We iterate through tiles to render each tile, but since `mergedToAnotherTile` is `true` in the rest of the tiles, the rendering is skipped.

src/gosling-track/gosling-track.ts
```diff
@@ -766,38 +763,37 @@ function GoslingTrack(HGC: import('@higlass/types').HGC, ...args: any[]): any {
...
-            this.reorganizeTileInfo();
-
-            this.combineAllTilesIfNeeded();
+            this.tileSize = this.tilesetInfo?.tile_size ?? 1024;
 
-            this.visibleAndFetchedTiles().forEach((tile: any) => {
-                if (force) {
-                    tile.goslingModels = [];
-                }
+            this.visibleAndFetchedTiles().forEach(tile => {
```
manzt: you can use `Array.flatMap` here:

```typescript
const models = this.visibleAndFetchedTiles()
    .flatMap(tile => this.preprocessTile(tile, force));
```

src/gosling-track/gosling-track.ts
sehilyi: Oh, this is a really great catch. This should be called only once regardless of the `resolvedTracks.length`.

sehilyi:
Thanks for the feedback again! I fixed several issues you found (e.g., unintentionally overwriting transformed data). There are several improvements that we can make, and I think we can work on them in follow-up PRs.

Below is a high-level summary of how tiles are processed and rendered in gosling tracks. This may not explain everything, and I think we can gradually improve this as we make updates on the `gosling-track.ts` file.

Untitled-2022-08-23-1419

https://excalidraw.com/#json=_S-yFMT_aK9mcpPVHOV_t,BlurYdpBgXknzSGNpEN1bQ

**Steps**
1. Generate tabular data for all tiles (`generateTabularData()`)
2. If needed, Combine tabular data of all tiles and store it to the first tile (`combineAllTilesIfNeeded()`)
3. Apply data transformations to the tabular data and generate track models using it (`transformDataAndCreateModels(tile);`)
4. Each of the track models is rendered separately by `drawTile(tile);`

PR #803, opened 2022-08-18
feat: add TabularDataFetcher interface

manzt:
Adds `TabularDataFetcher` interface. Refactors `drawLoadingCue` to `showLoadingCue` which returns a `hideLoadingCue` callback.

### DIFF
--- a/src/data-fetchers/bam/bam-data-fetcher.ts
+++ b/src/data-fetchers/bam/bam-data-fetcher.ts
@@ -9,10 +9,13 @@ import type { BamData, Assembly } from '@gosling.schema';
 import type { ModuleThread } from 'threads';
 import type { WorkerApi, TilesetInfo, Tiles, Segment, SegmentWithMate, Junction } from './bam-worker';
 import { computeChromSizes } from '../../core/utils/assembly';
+import type { TabularDataFetcher } from '../utils';
 
 const DEBOUNCE_TIME = 200;
 
-class BamDataFetcher {
+type BamTile = Segment | SegmentWithMate | Junction;
+
+class BamDataFetcher implements TabularDataFetcher<BamTile> {
     static config = { type: 'bam' };
     dataConfig = {}; // required for higlass
     uid: string;
@@ -76,8 +79,8 @@ class BamDataFetcher {
         (await this.worker).fetchTilesDebounced(this.uid, tileIds).then(receivedTiles);
     }
 
-    async getTabularData(uid: string, tileIds: string[]): Promise<Segment[] | SegmentWithMate[] | Junction[]> {
-        const buf = await (await this.worker).getTabularData(uid, tileIds);
+    async getTabularData(tileIds: string[]): Promise<BamTile[]> {
+        const buf = await (await this.worker).getTabularData(this.uid, tileIds);
         return JSON.parse(new TextDecoder().decode(buf));
     }
 }
--- a/src/data-fetchers/bam/bam-data-fetcher.ts
+++ b/src/data-fetchers/bam/bam-data-fetcher.ts
@@ -7,15 +7,19 @@ import Worker from './bam-worker.ts?worker&inline';
 
 import type { BamData, Assembly } from '@gosling.schema';
 import type { ModuleThread } from 'threads';
-import type { WorkerApi, TilesetInfo, Tiles, Segment, SegmentWithMate, Junction } from './bam-worker';
+import type { WorkerApi, TilesetInfo, Tiles, Segment, Junction, SegmentWithMate } from './bam-worker';
 import { computeChromSizes } from '../../core/utils/assembly';
 import type { TabularDataFetcher } from '../utils';
 
 const DEBOUNCE_TIME = 200;
 
-type BamTile = Segment | SegmentWithMate | Junction;
+type InferTileType<Config extends BamData> = Config['extractJunction'] extends true
+    ? Junction
+    : Config['loadMates'] extends true
+    ? SegmentWithMate
+    : Segment;
 
-class BamDataFetcher implements TabularDataFetcher<BamTile> {
+class BamDataFetcher<Config extends BamData> implements TabularDataFetcher<InferTileType<Config>> {
     static config = { type: 'bam' };
     dataConfig = {}; // required for higlass
     uid: string;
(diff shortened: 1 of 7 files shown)


src/gosling-track/gosling-track.ts
```diff
@@ -640,36 +636,33 @@ function GoslingTrack(HGC: import('@higlass/types').HGC, ...args: any[]): any {
...
-                this.loadingText.visible = true;
-                this.loadingTextBg.visible = true;
-            } else {
+            this.loadingText.visible = true;
+            this.loadingTextBg.visible = true;
+            return () => {
                 this.loadingText.visible = false;
                 this.loadingTextBg.visible = false;
-            }
+            };
```
manzt: I wonder if this could be even further refactored to add/remove the loading text from `this.pLabel`.

src/data-fetchers/bam/bam-data-fetcher.ts
```diff
@@ -9,10 +9,13 @@ import type { BamData, Assembly } from '@gosling.schema';
 import type { ModuleThread } from 'threads';
 import type { WorkerApi, TilesetInfo, Tiles, Segment, SegmentWithMate, Junction } from './bam-worker';
 import { computeChromSizes } from '../../core/utils/assembly';
+import type { TabularDataFetcher } from '../utils';
 
 const DEBOUNCE_TIME = 200;
 
-class BamDataFetcher {
+type BamTile = Segment | SegmentWithMate | Junction;
```
sehilyi: Consistent with VCF, we can move `BamTile` to `bam-workder.ts` and remove the import of `Segment`, `SegmentWithMate`, and `Junction` from this file.

src/data-fetchers/bam/bam-data-fetcher.ts
```diff
@@ -76,8 +79,8 @@ class BamDataFetcher {
         (await this.worker).fetchTilesDebounced(this.uid, tileIds).then(receivedTiles);
     }
 
-    async getTabularData(uid: string, tileIds: string[]): Promise<Segment[] | SegmentWithMate[] | Junction[]> {
-        const buf = await (await this.worker).getTabularData(uid, tileIds);
+    async getTabularData(tileIds: string[]): Promise<BamTile[]> {
```
sehilyi: Since `Segment[] | SegmentWithMate[] | Junction[]` is a more scoped type than `BamTile[]`, we could define a dedicated type, e.g., `BamTiles`:

```ts
// maybe in gosling.schema.guards.ts
type UnionOfArray<T> = T extends infer I ? I[] : never;

// bam-worker.ts
type BamTile = Junction | SegmentWithMate | Segment;
type BamTiles = UnionOfArray<BamTile>; // Junction[] | SegmentWithMate[] | Segment[]

// bam-data-fetcher.ts
async getTabularData(tileIds: string[]): Promise<BamTiles> {
```

We could use `BamTiles` in https://github.com/gosling-lang/gosling.js/blob/83cd2350945c2f4cefa2bdc667cb2a777795948e/src/data-fetchers/bam/bam-worker.ts#L437 as well

```suggestion
    async getTabularData(tileIds: string[]): Promise<BamTiles[]> {
```

src/gosling-track/gosling-track.ts
```diff
@@ -430,15 +431,15 @@ function GoslingTrack(HGC: import('@higlass/types').HGC, ...args: any[]): any {
...
-            this.drawLoadingCue();
+
+            const hide = this.showLoadingCue(`Fetching... ${Array.from(this.fetching).join(' ')}`);
             const tabularData = await tabularDataFetcher.getTabularData(
-                this.dataFetcher.uid,
-                Object.values(this.fetchedTiles).map((x: any) => x.remoteId)
+                Object.values(this.fetchedTiles as { remoteId: string }[]).map(x => x.remoteId)
             );
-            this.drawLoadingCue();
+            hide();
```
sehilyi: When I test a BAM example on the gosling editor, the loading label is displayed only for a very short time and disappears. This may be because the `updateTileAsync` function is called multiple times while zooming and panning.

Instead of hiding the label right away, I think the text of the label should be updated based on the `this.fetching` object since another tile might be being loaded.

sehilyi: I think calling `drawLoadingCue()` function was not so clean previously. Since the purpose of this function is basically updating the label whenever there is an update on the `this.fetching` object, we can instead call `drawLoadingCue()` function when new tile IDs are added or removed from the object in HiGlass functions, e.g.,:

```ts
// gosling.track.ts
receivedTiles(loadedTiles: unknown) {
  // https://github.com/higlass/higlass/blob/38f0c4415f0595c3b9d685a754d6661dc9612f7c/app/scripts/TiledPixiTrack.js#L637
  super.receivedTiles(loadedTiles); // some items in this.fetching are removed
  this.drawLoadingCue();
}

// https://github.com/higlass/higlass/blob/38f0c4415f0595c3b9d685a754d6661dc9612f7c/app/scripts/TiledPixiTrack.js#L342
removeOldTiles() {
  // https://github.com/higlass/higlass/blob/38f0c4415f0595c3b9d685a754d6661dc9612f7c/app/scripts/TiledPixiTrack.js#L637
  super.removeOldTiles(); // some items are added to this.fetching
  this.drawLoadingCue();
}
```

sehilyi (commented):
Thanks for this update! The only issue I found is that the change on the `showLoadingCue` function makes the 'loading label' displayed for a very short time, making it hard to read.

This PR:
pr

Current Gosling Editor:

editor

Since it is not specific to the current PR, we can address the refactoring issue of the `showLoadingCue` function given that it might be rather complicated.

src/data-fetchers/bam/bam-data-fetcher.ts
manzt: I see what you are saying.  I have a slightly different idea using generics:

```typescript
type InferTileType<T extends BamData> =
	T['extractJunction'] extends true ? Junction :
	T['loadMates'] extends true ? SegmentWithMate :
	Segment;

class BamDataFetcher<T extends BamData = BamData> implements TabularDataFetcher<InferTileType<T>> {
```

This way we encode the relationship between config and the expected tile type.

```typescript
declare var fetcher: BamDataFetcher;
fetcher.getTabularData([]) // Promise<Segment[] | SegmentWithMate[] | Junction[]>;

declare var junctionFetcher: BamDataFetcher<{ extractJunction: true, ... }>;
fetcher.getTabularData([]) // Promise<Junction[]>;
```

src/data-fetchers/bam/bam-data-fetcher.ts
manzt: See below for how I decided to detail with this.

src/gosling-track/gosling-track.ts
manzt: I just implemented this, but added a check to ensure we have a tabular datafetcher before calling `this.drawLoadingCue`.

sehilyi (approved):
Changes look good to me. Feel free to merge.

PR #804, opened 2022-08-19
fix: ensure to use non-null `hgRef`

sehilyi:
## Change List
 - Sometimes, `api.exportPng()` was making errors since `hgRef` is null. I believe this addresses the issue.

Screen Shot 2022-08-19 at 16 34 48

## Checklist

### DIFF
--- a/src/core/api.ts
+++ b/src/core/api.ts
@@ -39,13 +39,13 @@ export interface GoslingApi {
 }
 
 export function createApi(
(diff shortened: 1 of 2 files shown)


manzt (commented):
I don't think this solution will work, unfortunately. https://stackoverflow.com/questions/60476155/is-it-safe-to-use-ref-current-as-useeffects-dependency-when-ref-points-to-a-dom

PR #805, opened 2022-08-21
fix: bind `hgApi` as a ES6 Proxy for `hgRef.current` to keep updated ref

manzt:
The issue with `hgRef` is that we cannot listen for changes to re-create the `api` in our `useEffect` (because it is a Ref not React State). I think I came up with a solution to keep the `hgRef` but make `hgApi` an ES6 Proxy to `hgRef.current`. This way we can just access `hgApi` as whatever the current ref is. 

Additionally, I replaced `useEffect` with [`useImperativeHandle`](https://reactjs.org/docs/hooks-reference.html#useimperativehandle) which seems to be the preferable hook for this use case.

### DIFF
--- a/src/core/gosling-component.tsx
+++ b/src/core/gosling-component.tsx
@@ -1,6 +1,6 @@
 /* eslint-disable react/prop-types */
 import { HiGlassApi, HiGlassComponentWrapper } from './higlass-component-wrapper';
-import React, { useState, useEffect, useMemo, useRef, forwardRef, useCallback } from 'react';
+import React, { useState, useEffect, useMemo, useRef, forwardRef, useCallback, useImperativeHandle } from 'react';
 import ResizeSensor from 'css-element-queries/src/ResizeSensor';
 import * as gosling from '..';
 import { getTheme, Theme } from './utils/theme';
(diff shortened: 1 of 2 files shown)


src/core/gosling-component.tsx
```diff
@@ -52,16 +52,16 @@ export const GoslingComponent = forwardRef<GoslingRef, GoslingCompProps>((props,
...
-            ref.current = { api, hgApi };
-        }
-    }, [hgRef.current, viewConfig, theme]);
+    useImperativeHandle(
+        ref,
+        () => {
+            const hgApi = refAsReadonlyProxy(hgRef);
+            const infos = refAsReadonlyProxy(trackInfos);
+            const api = createApi(hgApi, viewConfig, infos, theme);
+            return { api, hgApi };
```
manzt, line 61: I wonder if we should even expose the low-level hgApi in Gosling.

sehilyi (approved):
LGTM! And tested locally and works well.

src/core/gosling-component.tsx
sehilyi: I think this can be useful when ones want to use API functions that Gosling does not yet support, but I generally agree that it's better not to expose higlass-specific features in Gosling. We can make this deprecated at some point.

PR #808, opened 2022-08-23
feat: add type heirarchy for BarTrack

manzt:
This took a while but I think it will be worth it. This PR adds the types (and complete hierarchy) for HiGlass's `BarTrack`.

### DIFF
--- a/src/core/mark/legend.ts
+++ b/src/core/mark/legend.ts
@@ -5,7 +5,6 @@ import type { CompleteThemeDeep } from '../utils/theme';
 import type { Dimension } from '../utils/position';
 import { scaleLinear, ScaleLinear } from 'd3-scale';
 import { getTextStyle } from '../utils/text-style';
-import type { Tile } from '../../gosling-track/gosling-track';
 
 // Just the libraries necesssary fro this module
 type Libraries = Pick<typeof import('@higlass/libraries'), 'PIXI' | 'd3Selection' | 'd3Drag'>;
@@ -15,7 +14,7 @@ type LegendOffset = { offsetRight: number };
 export function drawColorLegend(
     HGC: { libraries: Libraries },
     trackInfo: any,
-    tile: Tile,
+    _tile: unknown,
     model: GoslingTrackModel,
     theme: Required<CompleteThemeDeep>
 ) {
@@ -32,10 +31,10 @@ export function drawColorLegend(
     if (IsChannelDeep(spec.color) && spec.color.legend) {
         switch (spec.color.type) {
             case 'nominal':
-                drawColorLegendCategories(HGC, trackInfo, tile, model, theme);
+                drawColorLegendCategories(HGC, trackInfo, _tile, model, theme);
                 break;
             case 'quantitative':
-                drawColorLegendQuantitative(HGC, trackInfo, tile, model, theme, 'color', offset);
+                drawColorLegendQuantitative(HGC, trackInfo, _tile, model, theme, 'color', offset);
                 break;
         }
     }
@@ -43,7 +42,7 @@ export function drawColorLegend(
     if (IsChannelDeep(spec.stroke) && spec.stroke.legend) {
         switch (spec.stroke.type) {
             case 'quantitative':
-                drawColorLegendQuantitative(HGC, trackInfo, tile, model, theme, 'stroke', offset);
+                drawColorLegendQuantitative(HGC, trackInfo, _tile, model, theme, 'stroke', offset);
                 break;
         }
     }
@@ -52,7 +51,7 @@ export function drawColorLegend(
 export function drawColorLegendQuantitative(
     HGC: { libraries: Libraries },
     trackInfo: any,
-    tile: Tile,
+    _tile: unknown,
     model: GoslingTrackModel,
     theme: Required<CompleteThemeDeep>,
     channelKey: 'color' | 'stroke',
@@ -292,7 +291,7 @@ export function drawColorLegendQuantitative(
 export function drawColorLegendCategories(
     HGC: { libraries: Libraries },
     trackInfo: any,
-    tile: Tile,
+    _tile: unknown,
     tm: GoslingTrackModel,
     theme: Required<CompleteThemeDeep>
 ) {
@@ -448,7 +447,7 @@ export function drawColorLegendCategories(
 export function drawRowLegend(
     HGC: { libraries: Libraries },
     trackInfo: any,
-    tile: Tile,
+    _tile: unknown,
     tm: GoslingTrackModel,
     theme: Required<CompleteThemeDeep>
 ) {
--- a/src/gosling-track/gosling-track.ts
+++ b/src/gosling-track/gosling-track.ts
@@ -42,8 +42,7 @@ import { LinearBrushModel } from '../gosling-brush/linear-brush-model';
 import { isPointInsideDonutSlice } from '../gosling-mouse-event/polygon';
 import type { Tile as _Tile, TileData, TileDataBase } from '@higlass/services';
 import type { TabularDataFetcher } from 'src/data-fetchers/utils';
-import { createPluginTrack, PluginTrackFactory } from 'src/core/utils/define-plugin-track';
-import type { TrackConfig } from '@higlass/types';
+import { createPluginTrack, PluginTrackFactory, TrackConfig } from '../core/utils/define-plugin-track';
 import { getTheme } from 'gosling-theme';
 
 // Set `true` to print in what order each function is called
@@ -107,12 +106,11 @@ const config: TrackConfig<GoslingTrackOptions> = {
         // barBorder: false,
         // sortLargestOnTop: true,
         // axisPositionHorizontal: 'left',
-        theme: getTheme('light'),
+        theme: getTheme('light')
     }
 };
 
 const factory: PluginTrackFactory<Tile, GoslingTrackOptions> = (HGC, context, options) => {
-
     // Services
     const { tileProxy } = HGC.services;
 
@@ -467,7 +465,6 @@ const factory: PluginTrackFactory<Tile, GoslingTrackOptions> = (HGC, context, op
             this.forceDraw();
         }
 
-
         /**
          * This is currently for testing the new way of rendering visual elements.
          */
@@ -675,7 +672,7 @@ const factory: PluginTrackFactory<Tile, GoslingTrackOptions> = (HGC, context, op
             if (!yTiles) {
                 // this means only the `x` axis is being used
                 return xTiles.map(x => [zoomLevel, x]);
-            } 
+            }
             // this means both `x` and `y` axes are being used together
             const tiles: [number, number, number][] = [];
             xTiles.forEach(x => yTiles.forEach(y => tiles.push([zoomLevel, x, y])));
@@ -897,7 +894,7 @@ const factory: PluginTrackFactory<Tile, GoslingTrackOptions> = (HGC, context, op
                 // generate tabular data
                 const { tileX, tileY, tileWidth, tileHeight } = this.getTilePosAndDimensions(
                     tile.tileData.zoomLevel,
-                    tile.tileData.tilePos!,
+                    tile.tileData.tilePos!
                 );
 
                 const sparse = 'length' in tile.tileData ? Array.from(tile.tileData) : [];
@@ -918,7 +915,7 @@ const factory: PluginTrackFactory<Tile, GoslingTrackOptions> = (HGC, context, op
             this.processedTileInfo[tile.tileId] = tileInfo;
         }
 
-         /**
+        /**
          * Apply data transformation to each of the overlaid tracks and generate gosling models
          */
         transformDataAndCreateModels(tile: Tile) {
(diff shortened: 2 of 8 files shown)


manzt:
I will wait to continue on this until #802 is merged and I can update the tile types as well.

sehilyi:
Just merged https://github.com/gosling-lang/gosling.js/pull/802. It looks like there are some conflicts, probably due to changes I additionally made on the merged PR. Let me know if you need any help when resolving the conflicts.

manzt:
Ok, should be ready for review!

src/core/mark/legend.ts
```diff
@@ -15,7 +14,7 @@ type LegendOffset = { offsetRight: number };
 export function drawColorLegend(
     HGC: { libraries: Libraries },
     trackInfo: any,
-    tile: Tile,
+    _tile: unknown,
```
manzt, line 17: This is just to fix our tests (which provide a tile as a `null` value). These functions should ideally define an interface which does not include `tile` as property,

```typescript
interface DrawColorLegendOptions {
  trackInfo: any;
  model: GoslingModel;
  theme: Required<CompleteThemeDeep>;
}

export function drawColorLegend(HGC, opts: DrawColorLegendOptions);

drawColorLegend(HGC, { trackInfo, model, theme }) // ok
drawColorLegend(HGC, { trackInfo, model, theme, track }) // ok
drawColorLegend(HGC, { trackInfo, theme, track }) // error
```

src/gosling-track/gosling-track.ts
```diff
@@ -89,47 +90,77 @@ function initProcessedTileInfo(): ProcessedTileInfo {
...
+        // TODO: Are any of these used?
+        // labelPosition: 'none',
+        // labelColor: 'black',
+        // labelTextOpacity: 0.4,
+        // trackBorderWidth: 0,
+        // trackBorderColor: 'black',
+        // backgroundColor: 'white',
+        // barBorder: false,
+        // sortLargestOnTop: true,
+        // axisPositionHorizontal: 'left',
```
manzt, line 108: These don't seem to have any effect?

src/gosling-track/gosling-track.ts
```diff
@@ -89,47 +90,77 @@ function initProcessedTileInfo(): ProcessedTileInfo {
...
+        private svgData: unknown[] = [];
+        private textGraphics: unknown[] = [];
+        /** used to improve the performance of text rendering */
+        private textsBeingUsed = 0;
+        private loadingStatus = { loading: 0, processing: 0, rendering: 0 };
+
+        private prevVisibleAndFetchedTiles?: Tile[];
+        private oldDimensions?: [number, number];
+        private xDomain?: number[];
+        private xRange?: number[];
```
manzt, line 154: most of these properties are not used by this class. Think we can remove.

src/gosling-track/gosling-track.ts
```diff
@@ -89,47 +90,77 @@ function initProcessedTileInfo(): ProcessedTileInfo {
...
+        // Brushes on the color legend
+        private gLegend? = HGC.libraries.d3Selection.select(context.svgElement).append('g');
+        private pMouseHover = new HGC.libraries.PIXI.Graphics();
+        private pMouseSelection = new HGC.libraries.PIXI.Graphics();
+        private isRangeBrushActivated = false;
+        private gBrush = HGC.libraries.d3Selection.select(context.svgElement).append('g');
+
+        private loadingTextStyleObj = new HGC.libraries.PIXI.TextStyle(loadingTextStyle);
+        private loadingTextBg = new HGC.libraries.PIXI.Graphics();
+        private loadingText = new HGC.libraries.PIXI.Text('', loadingTextStyle);
```
manzt, line 143: I moved all this deterministic setup logic outside of the constructor.

src/gosling-track/gosling-track.ts
```diff
@@ -89,47 +90,77 @@ function initProcessedTileInfo(): ProcessedTileInfo {
...
+        private xRange?: number[];
+
+        private mouseDownX = 0;
+        private mouseDownY = 0;
+
+        constructor() {
             super(context, options);
 
             context.dataFetcher.track = this;
-            this.context = context;
```
manzt, line 130: We have `context` available in the closure. I don't think adding even more to the class hierarchy is very useful.

src/gosling-track/gosling-track.ts
```diff
@@ -144,43 +175,34 @@ function GoslingTrack(HGC: import('@higlass/types').HGC, ...args: any[]): any {
...
-            this.pMask.mousedown = (e: InteractionEvent) =>
-                this.onMouseDown(
-                    e.data.getLocalPosition(this.pMain).x,
-                    e.data.getLocalPosition(this.pMain).y,
-                    e.data.originalEvent.altKey
-                );
-            this.pMask.mouseup = (e: InteractionEvent) =>
-                this.onMouseUp(e.data.getLocalPosition(this.pMain).x, e.data.getLocalPosition(this.pMain).y);
-            this.pMask.mousemove = (e: InteractionEvent) => this.onMouseMove(e.data.getLocalPosition(this.pMain).x);
-            this.pMask.mouseout = () => this.onMouseOut();
```
manzt, line 183: These were all type errors. Looking through the `pixi` docs, the `.on` listeners seem to be the preferred syntax. This is nice because it also mirrors our `LinearBrushModel.on` listener api!

src/gosling-track/gosling-track.ts
```diff
@@ -303,8 +315,8 @@ function GoslingTrack(HGC: import('@higlass/types').HGC, ...args: any[]): any {
                 return;
             }
 
-            tile.graphics.clear();
-            tile.graphics.removeChildren();
+            tile.graphics?.clear();
+            tile.graphics?.removeChildren();
```
manzt, line 319: is graphics always defined? if so, I can remove this...

src/gosling-track/gosling-track.ts
```diff
@@ -385,7 +397,7 @@ function GoslingTrack(HGC: import('@higlass/types').HGC, ...args: any[]): any {
 
             if (this.gLegend) {
                 this.gLegend.remove();
-                this.gLegend = null;
+                this.gLegend = undefined;
```
manzt, line 400: prefer `undefined` to `null`

src/gosling-track/gosling-track.ts
```diff
@@ -542,67 +542,84 @@ function GoslingTrack(HGC: import('@higlass/types').HGC, ...args: any[]): any {
...
-                        this.yTiles = tileProxy.calculateTiles(
-                            this.zoomLevel,
+                        yTiles = tileProxy.calculateTiles(
+                            zoomLevel,
                             this._yScale,
                             this.tilesetInfo.min_pos[1],
                             this.tilesetInfo.max_pos[1],
                             this.tilesetInfo.max_zoom,
-                            this.tilesetInfo.max_width1 || this.tilesetInfo.max_width
+                            // @ts-expect-error what is max_width1?
```
manzt, line 592: yeah not sure about this!

src/gosling-track/gosling-track.ts
```diff
@@ -542,67 +542,84 @@ function GoslingTrack(HGC: import('@higlass/types').HGC, ...args: any[]): any {
...
+        private get binsPerTile() {
+            let maybeValue: number | undefined;
+            if (this.tilesetInfo) {
+                maybeValue =
+                    'bins_per_dimension' in this.tilesetInfo
+                        ? this.tilesetInfo.bins_per_dimension
+                        : this.tilesetInfo.tile_size;
+            }
+            return maybeValue ?? 256;
+        }
```
manzt, line 612: This logic was repeated all over the place. 

Moving it to a private class property means that we can modify the logic in the future and not worry about needing to keep everything in sync.

src/gosling-track/gosling-track.ts
```diff
@@ -542,67 +542,84 @@ function GoslingTrack(HGC: import('@higlass/types').HGC, ...args: any[]): any {
...
+
         /**
          * Get the tile's position in its coordinate system.
          */
-        getTilePosAndDimensions(zoomLevel: number, tilePos: [number, number], binsPerTileIn?: number) {
-            const binsPerTile = binsPerTileIn || this.tilesetInfo.bins_per_dimension || 256;
```
manzt, line 603: this local is moved to `this.binsPerTile`

src/gosling-track/gosling-track.ts
```diff
@@ -612,7 +629,7 @@ function GoslingTrack(HGC: import('@higlass/types').HGC, ...args: any[]): any {
 
                 const [xTilePos, yTilePos] = tilePos;
 
-                const tileWidth = chosenResolution * binsPerTile;
+                const tileWidth = chosenResolution * this.binsPerTile;
```
manzt, line 632: see use of `this.binsPerTile` here.

src/gosling-track/gosling-track.ts
```diff
@@ -995,22 +1012,18 @@ function GoslingTrack(HGC: import('@higlass/types').HGC, ...args: any[]): any {
...
-                tile.tileData.zoomLevel,
-                tile.tileData.tilePos,
-                this.tilesetInfo.bins_per_dimension || this.tilesetInfo?.tile_size
-            );
+            const { tileX, tileWidth } = this.getTilePosAndDimensions(tile.tileData.zoomLevel, tile.tileData.tilePos);
 
             const tileXScale = HGC.libraries.d3Scale
                 .scaleLinear()
-                .domain([0, this.tilesetInfo?.tile_size || this.tilesetInfo?.bins_per_dimension])
+                .domain([0, this.binsPerTile])
```
manzt, line 1026: `this.binsPerTile` here as well

src/gosling-track/gosling-track.ts
```diff
@@ -1377,73 +1396,10 @@ function GoslingTrack(HGC: import('@higlass/types').HGC, ...args: any[]): any {
                     return `<div>${content}</div>`;
                 }
             }
+            return '';
         }
     }
-    return new GoslingTrackClass(args);
-}
-
-const goslingIcon = `<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 400 400" width={30} height={30}>
```
manzt, line 1385: This is used anywhere in Gosling? This seems like a custom HiGlass thing, and just adds a lot of text to our library. If it's useful I can add it back, but I really don't think it's necessary.

sehilyi: Removing it sounds good. As far as I know, it's useful only if we use the HiGlass modal view, but we do not use it.

src/core/mark/legend.ts
sehilyi: Maybe we gotta remove `_tile` from the argument since it seems like the variable is not used inside the functions.

src/gosling-track/gosling-track.ts
sehilyi: Yes, let's remove the commented-out properties.

src/gosling-track/gosling-track.ts
sehilyi: Yes, I think we can remove them except for `textGraphics`, `textsBeingUsed`, and `prevVisibleAndFetchedTiles`. `textGraphics` and `textsBeingUsed` are used in `mark/text.ts`.

src/gosling-track/gosling-track.ts
sehilyi: Hmm, not sure about it, too. I see it is used in HeatmapTiledPixiTrack.js of the higlass only.

src/gosling-track/gosling-track.ts
sehilyi: I believe they are always defined. Many `mark/*.ts` files also consider `tile.graphics` to be defined.

src/gosling-track/gosling-track.ts
```diff
@@ -89,47 +90,77 @@ function initProcessedTileInfo(): ProcessedTileInfo {
...
-        private pMouseSelection: Graphics;
+
         private assembly?: Assembly;
-        private fetchedTiles: FetchedTiles;
         private processedTileInfo: Record<string, ProcessedTileInfo>;
         // TODO: add members that are used explicitly in the code
-
-        constructor(params: any[]) {
-            const [context, options] = params;
+        //
```
sehilyi, line 133: We can remove this comment now? (i.e., `// TODO: add members that are used explicitly in the code`)

sehilyi (approved):
This is SUPER!

src/core/mark/legend.ts
manzt: That was my thought, but I wasn't sure if other functions had a similar signature (ie, `draw*()`) and it would be useful to allow for `tile` to be passed somehow when calling.

manzt: something for a follow up PR I think.

src/gosling-track/gosling-track.ts
manzt: > textGraphics, textsBeingUsed, and prevVisibleAndFetchedTiles. 

Then we should mark these properties as `public` not private

src/gosling-track/gosling-track.ts
manzt: Ok I'll add it to that type...

src/gosling-track/gosling-track.ts
manzt: Done!

PR #828, opened 2022-11-26
fix: improve within link performance by reducing the number of event points

sehilyi:
Improve `withinLink` rendering performance by reducing the number of mouse event points. We previously saved more event points than what are displayed. This was, initially, to make the mouse events more fine-grained, but this is not necessary anymore.


src/core/mark/withinLink.ts
```diff
@@ -324,13 +324,10 @@ export function drawWithinLink(g: PIXI.Graphics, trackInfo: any, model: GoslingT
                                         : Math.min(xe - x + trackWidth * MIN_HEIGHT, trackWidth) / trackWidth) *
                                     (flipY ? -1 : 1);
 
-                            if (step % 20 === 0 || step === NUM_STEPS) {
```
sehilyi, line 327: This was unnecessary.

PR #834, opened 2022-12-08
fix: correctly display legends when multiple tracks are overlaid

sehilyi:
The issue is that if two tracks are overlaid and the last (2nd) track does not include a color legend (i.e., `legend: false`), no legends are displayed even if the first track contains the color legend. This PR fixes this issue.


sehilyi:
We still need to find a way to display multiple legends without visual occlusion.

PR #850, opened 2023-03-22
chore: upgrade to TS 5.0

manzt:
TypeScript doesn't follow semver, so 5.0 isn't semantically meaninfful from 4.9 for example. However, this update did introduce some changes that required fixing. 

The primary issue is that there are a lot of bits of the current code that compare `string` to `number` and don't verfy that we are only comparing numbers. I added some type gaurds to narrow the types, but I bet we could use the type system more effectively to avoid needing to check the types in these paths (i.e., statically verify the types rather than at runtime). 

The second big change was migrating to `verbatimModuleSyntax` TS config option, which makes the type-only imports more explicit. This is the newer version of `importTypeAsValues`, which has been deprecated. See comment in the `tsconfig.json` for more info.

Fix #
Toward #

## Change List
 -

## Checklist


tsconfig.json
```diff
@@ -20,7 +20,7 @@
     "strict": true,
     "strictNullChecks": true,
     "resolveJsonModule": true,
-    "importsNotUsedAsValues": "error",
+    "verbatimModuleSyntax": true,
```
manzt, line 23: has the same behavior as before but is even more strict (`importNotUsedAsValues`) is now deprecated. This makes us be explicit when importing types:

```ts
import { ScaleLinear } from "d3-scale"; // error
import { scaleLinear, ScaleLinear } from "d3-scale"; // error, previously ok

import type { ScaleLinear } from "d3-scale"; // ok
import { scaleLinear, type ScaleLinear } from "d3-scale"; // ok
```

sehilyi (approved):
> I bet we could use the type system more effectively to avoid needing to check the types in these paths (i.e., statically verify the types rather than at runtime)

Your change on the type check looks good, but I agree

tsconfig.json
sehilyi: Nice

manzt:
> Your change on the type check looks good, but I agree

Just a general comment. The current changes improve the type safety with minimal refactoring, which is a good first step. Ideally we also try to systematically improve our types (removing uses of `any`, or type-casting with `as number[]` etc) so that if it makes sense to make a larger refactor, we can do so with much more confidence.

Issue #853, opened 2023-03-27
Improve the use of TypeScript types


sehilyi:
There is room for improvement regarding the use of TypeScript types. This will enable us to maintain and improve the package in a more robust way.

- HiGlass does not provide types, so we are partly implementing them in our repo (https://github.com/gosling-lang/gosling.js/blob/master/src/missing-types.d.ts).

  - https://github.com/gosling-lang/gosling.js/issues/702

- Many `any` types can be removed by inferring/specifying proper types

  - https://github.com/gosling-lang/gosling.js/issues/695

sehilyi:
we have several fine-grained issues, so we can close this rather too broad one

Issue #650, opened 2022-03-03
Git repo size


manzt:
Not a concern with Gosling itself, but I recently re-cloned the repo and found it to be quite slow (~3 min). Upon inspection, the `.git/` folder is very large (1.2 GB). I think this can occur from checking in many large binary objects (e.g., images).

```
gosling.js on  master is 📦 v0.9.17 via  v17.6.0
❯ du -h -d1

4.0K    ./embed
5.6M    ./img
336K    ./schema
 16K    ./public
 68K    ./logo
8.0K    ./scripts
 28K    ./.github
1.2G    ./.git
 36K    ./notebooks
608K    ./editor
996K    ./src
1.2G    .
```

for comparison, the `.git` folder for higlass (with >5000 commits) is 227M.

manzt:
I think at some point the bundle for Gosling was checked into this repo, and was ~42Mb?

from https://stackoverflow.com/a/14329983/11008641

```
gosling.js on  master is 📦 v0.9.17 via  v17.6.0
❯ git rev-list --all --objects | \
    sed -n $(git rev-list --objects --all | \
    cut -f1 -d' ' | \
    git cat-file --batch-check | \
    grep blob | \
    sort -n -k 3 | \
    tail -n40 | \
    while read hash type size; do
         echo -n "-e s/$hash/$size/p ";
    done) | \
    sort -n -k1
42667313 main.js
42723724 main.js
42723724 main.js
42724139 main.js
42724200 main.js
42724507 main.js
42724663 main.js
42726288 main.js
42726560 main.js
42726877 main.js
42726964 main.js
42727343 main.js
42727343 main.js
42728427 main.js
42762390 main.js
42763888 main.js
42763888 main.js
42773344 main.js
42773344 main.js
42773344 main.js
42792513 main.js
42792513 main.js
42793592 main.js
42793592 main.js
42793756 main.js
42794184 main.js
42794185 main.js
42832661 main.js
42832665 main.js
42832722 main.js
42854024 main.js
42854024 main.js
42904025 main.js
42904153 main.js
42904311 main.js
42904311 main.js
43096900 main.js
43096901 main.js
43096946 main.js
43097807 main.js
```

manzt:
Ok, so these large files seem to be coming from the `gh-pages` branch which I guess makes sense. A simple fix would be to delete and recreate the `gh-pages` branch, and then run `git gc --prune=now --aggressive` which will cleanup all the objects. 

[...]

PR #859, opened 2023-03-31
fix(editor): reduce asset size 

### LINKED ISSUE #696: Reduce asset size for the initial load of Editor
Currently, the Editor requires ~100MB of data to be transferred for the initial page load. The demo example image files are the ones that make the loading too slow. We will need to optimize the loading, e.g., compress image files, reduce their sizes given that we only use them for showing thumbnails, and load images only when a user clicks on the example gallery.

image

Image directory: https://github.com/gosling-lang/gosling.js/tree/master/editor/example/thumbnails

etowahadams:
Fix #696 
Toward:  #853, #650

## Change List
 - Resize the example thumbnails so the maximum with is 500px. The total combined size of the thumbnails is now ~10 MB
 - Pull out the editor examples into its own component (EditorExamples) and add types 
 - Make thumbnail images lazy load 

## Checklist


editor/editor.tsx
```diff
@@ -1,30 +1,31 @@
...
-import './editor.css';
+import { ICONS, type ICON_INFO } from './icon';
+import { getHtmlTemplate } from './html-template';
+import ErrorBoundary from './error-boundary';
+import { traverseTracksAndViews } from '../src/core/utils/spec-preprocess';
+import { examples, type Example } from './example';
+import EditorPanel, { type EditorLangauge } from './EditorPanel';
+import EditorExamples from './EditorExamples';
+
+import './Editor.css';
```
etowahadams, line 28: Reorder the imports. We should consider having a eslint rule to do this automatically

editor/editor.tsx
```diff
@@ -1381,84 +1388,14 @@ function Editor(props: RouteComponentProps) {
...
-                </div>
-            </div>
+            {showExamples && (
+                <EditorExamples
+                    setShowExamples={setShowExamples}
+                    closeDescription={closeDescription}
+                    setIsImportDemo={setIsImportDemo}
+                    setDemo={setDemo}
+                />
+            )}
```
etowahadams, line 1398: Pull out the editor examples into its own component. The conditional rendering is much more clear this way

editor/EditorExamples.css
```diff
@@ -0,0 +1,130 @@
...
+    height: 30px;
+    text-overflow: ellipsis;
+    white-space: nowrap;
+    overflow: hidden;
+    display: block;
+}
+
+.example-card-bg:hover {
+    background-position: bottom right;
+}
```
etowahadams, line 130: Pulled out from Editor.css

editor/EditorExamples.css
etowahadams, line 1: Styles for editor examples were moved from Editor.css to EditorExamples.css

editor/editor.tsx
sehilyi: It would be good to have!

editor/editor.tsx
sehilyi: Thanks for making this change!

sehilyi (approved):
Looks great!

Do you want to write instructions in `CONTRIBUTING.md` about how you compressed the images? We will be adding more examples to Editor, and having that will enable people (incl. us) to make consistent thumbnails.

etowahadams:
I'll add instructions about compressing to CONTRIBUTING.md! I just used the built-in Preview to batch resize the images, but in theory imagemagick, ffmpeg etc could be used

CONTRIBUTING.md
etowahadams, line 1: Let me know if these instructions seem sufficient

sehilyi: Looks great! Thanks for adding this!

PR #864, opened 2023-04-06
feat: support optional INFO fields of VCF files

### LINKED ISSUE #858: Test parsing optional columns in VCF files
Not sure we supported this at some point, but we need to support this and include that in the documentation.

sehilyi:
Fix #858

## Change List
 - Support parsing INFO fields of VCF files. 

### A tooltip showing the two of the INFO fields 
Screenshot 2023-04-06 at 11 17 57

### A single line of a VCF file

> chr1	118168808	.	G	A	62.19	PASS	ECNT=1;FS=5.392;HCNT=1;MAX_ED=.;MIN_ED=.;NLOD=8.42;NLODF=4.13;PV=0.0028;PV2=0.0028;SOR=1.397;TLOD=41.24	GT:AD:AF:AFDP:ALTHC:ALT_F1R2:ALT_F2R1:BaseQRankSumPS:ClippingRankSumPS:DPHC:FOXOG:MQRankSumPS:NBQPS:QSS:REF_F1R2:REF_F2R1:ReadPosEndDistPS:ReadPosRankSumPS	0/

### Demo on Local Browser

[Link](HTTP://localhost:3000/?full=false&spec=(M%27layout4%27linear%27%2CM%27arrangement4%27vertical%27%2CM%27centerRadius40.8%2CM%27viewsXM*(M**%27tracksXM***(2%27dataTransformX2*(%27Qfilter7JoneOfXnull%5D)2%5DH%27id4%27track-1%27H%27data4(2*%27url4%27hFK~Qvcf~indexUrl4%27hFK.tbi~sampleLength450002)H%27tooltipX2*(%27JQnominal7format4%27s1%27)H*W%27)2%5DH%27mark4%27point%27H%27x4WZcolor4(2*%27jSUBTYPE~Qnominal~legend4trueH*%27domainX%27C%3EA7C%3EG7C%3ET7T%3EA7T%3EC7T%3EG%27%5D2)H%27y4(%27JQquantitativeZopacity4(%27value40.9)H%27width4600H%27height4130M***)M**%5DM*)M%5D%0A)*%20%202M****4!%207%27%2C%20%27Fttps%3A%2F%2Fs3.amazonaws.com%2Fgosling-lang.org%2Fdata%2FSV%2FH%2C2JjDISTPREVLOGE7KSNV_test_tumor_normal_with_panel.vcf.gzM%0A*Qtype4%27W(%27jPOS7QgenomicX4%5BZ7axis4%27top%27)H%27jfield4%27~%27H*%27%01~jZXWQMKJHF742*_)

## Checklist

### DIFF
--- a/src/data-fetchers/vcf/vcf-data-fetcher.test.ts
+++ b/src/data-fetchers/vcf/vcf-data-fetcher.test.ts
@@ -1,5 +1,5 @@
-import { recordToTile } from './vcf-worker';
-import type { VcfRecord } from './vcf-worker';
+import { recordToTile } from './vcf-data-fetcher';
+import type { VcfRecord } from './vcf-data-fetcher';
 
 describe('VCF file parser', () => {
     it('Convert a VCF record to a tile data', () => {
--- a/src/data-fetchers/vcf/vcf-data-fetcher.test.ts
+++ b/src/data-fetchers/vcf/vcf-data-fetcher.test.ts
@@ -21,6 +21,28 @@ describe('VCF file parser', () => {
         const prevPos = record.POS - 100;
         const tile = recordToTile(record, chrPos, prevPos);
 
-        expect(tile).toMatchInlineSnapshot();
+        expect(tile).toMatchInlineSnapshot(`
+          {
+            "AF": "0.5, DB, H2",
+            "ALT": "A",
+            "CHROM": "20",
+            "DISTPREV": 100,
+            "DISTPREVLOGE": 4.605170185988092,
+            "DP": "14",
+            "FILTER": "PASS",
+            "ID": [
+              "rs6054257",
+            ],
+            "INFO": "{\\"NS\\":[3],\\"DP\\":[14],\\"AF\\":[\\"0.5\\",\\"DB\\",\\"H2\\"]}",
+            "MUTTYPE": "substitution",
+            "NS": "3",
+            "ORIGINALPOS": 14370,
+            "POS": 137827,
+            "POSEND": 137828,
+            "QUAL": 29,
+            "REF": "G",
+            "SUBTYPE": "C>T",
+          }
+        `);
     });
 });
--- a/src/data-fetchers/vcf/vcf-data-fetcher.test.ts
+++ b/src/data-fetchers/vcf/vcf-data-fetcher.test.ts
@@ -22,7 +22,7 @@ describe('VCF file parser', () => {
         const tile = recordToTile(record, chrPos, prevPos);
 
         expect(tile.MUTTYPE).toBe('substitution');
-        expect(tile.SUBTYPE).toBe('G>A');
+        expect(tile.SUBTYPE).toBe('C>T');
         expect(tile.DISTPREV).toBe(chrPos + record.POS - prevPos);
         expect(tile.NS).toBe(3);
     });
--- a/src/data-fetchers/vcf/vcf-data-fetcher.test.ts
+++ b/src/data-fetchers/vcf/vcf-data-fetcher.test.ts
@@ -21,9 +21,6 @@ describe('VCF file parser', () => {
         const prevPos = record.POS - 100;
         const tile = recordToTile(record, chrPos, prevPos);
 
-        expect(tile.MUTTYPE).toBe('substitution');
-        expect(tile.SUBTYPE).toBe('C>T');
-        expect(tile.DISTPREV).toBe(100);
-        expect(tile.NS).toBe('3'); // always string
+        expect(tile).toMatchInlineSnapshot();
     });
 });
--- a/src/data-fetchers/vcf/vcf-data-fetcher.test.ts
+++ b/src/data-fetchers/vcf/vcf-data-fetcher.test.ts
@@ -1,4 +1,4 @@
-import { recordToTile } from './vcf-data-fetcher';
+import { recordToTile } from './utils';
 import type { VcfRecord } from './vcf-data-fetcher';
 
 describe('VCF file parser', () => {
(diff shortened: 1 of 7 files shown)


src/data-fetchers/vcf/vcf-worker.ts
```diff
@@ -184,10 +184,14 @@ const tile = async (uid: string, z: number, x: number): Promise<void[]> => {
...
+            // Add optional INFO columns
             Object.keys(vcfRecord.INFO).forEach(key => {
                 const val = vcfRecord.INFO[key];
-                if (Array.isArray(val)) return [key, val[0]] as const;
-                return [key, val] as const;
+                if (Array.isArray(val)) {
+                    data[key] = val.join(', ');
+                } else {
+                    data[key] = val;
+                }
```
manzt: `parseLineStoreData` has several responsibilities:

1.) parse a `VcfRecord` from a string using the `parser`
2.) convert a `VcfRecord` to `VcfTile`
3.) add the result to a global cache


Let's move the logic for 2 to a separate function, e.g., `recordToTile`. All this custom logic for converting a record to a tile can live in a function (which we could also unit test).

```typescript
       const parseLineStoreData = (line: string, prevPos?: number) => {
            const vcfRecord: VcfRecord = parser.parseLine(line);
            const data = recordToTile(vcfRecord, prevPos);

            // Store this column
            tileValues[CACHE_KEY] = tileValues[CACHE_KEY].concat([data]);

            // Return current POS
            return vcfRecord.POS;
        };
```

manzt (approved):
This looks good, just some minor comments. I think on a deeper level we should consider refactoring the `tiles` implementation. The logic is tricky to reason about because there the are so many moving parts. It combines calculating genomic coordinates, using the vcf parser, converting to a vcf tile, and then caching the data on a global object. The callbacks don't return anything, so keeping track of where data are is really tricky.

It seems a lot of this could be simplified if we moved all the vcf parsing and tile conversion behind a different parser interface that returned `Promise<VcfTile[]>`,

```ts
 let tilePromises = cumPositions.map(async cumPos => {
      const chromName = cumPos.chr;
      const chromStart = cumPos.pos;
      const chromEnd = cumPos.pos + chromLengths[chromName];
      // ...
      const tiles = await source.file.getTiles(chromName, chromStart, chromEnd); // new API, uses Parser internally
      return tiles;
});
let allTiles = await Promise.all(tilePromises);
tileValues[CACHE_KEY] = allTiles.flat();
```

src/data-fetchers/vcf/vcf-worker.ts
sehilyi: Thank you for this suggestion. I made this change and also added a unit test.

sehilyi:
@manzt It is indeed tricky to follow the flow of the data, and I like your suggestion. Tried to address this, but it looks like this will make some structural changes that I think would be better to work on a separate PR. Let me open a PR.

manzt:
Hmm, it seems like there were a bunch of changes added on top of this PR rather than a merge or rebase? This will make the git history challenging to resolve.

src/data-fetchers/vcf/vcf-data-fetcher.test.ts
```diff
@@ -0,0 +1,29 @@
...
+            }
+        };
+        const chrPos = 123456;
+        const prevPos = record.POS - 100;
+        const tile = recordToTile(record, chrPos, prevPos);
+
+        expect(tile.MUTTYPE).toBe('substitution');
+        expect(tile.SUBTYPE).toBe('C>T');
+        expect(tile.DISTPREV).toBe(100);
+        expect(tile.NS).toBe('3'); // always string
```
manzt: I think this would be a really good choice to use a [snapshot test](https://vitest.dev/guide/snapshot.html#inline-snapshots). Snapshot testing is especially useful for this kind of thing, where you have a transformation in mind and just want to make sure the data "looks right". Checking in the entire snapshot helps to surface changes in the current behavior. It also means you can write many fewer assertions. 

To create a snapshot, you can replace the above with the following in the source code:

```suggestion
        expect(tile).toMatchInlineSnapshot();
```

Then, run `yarn test`. The very first time it runs, `vitest` will take a "snapshot" of `tile`, and generate a string _inline_. For example, when I run the tests, the following is generated:

```typescript
        expect(tile).toMatchInlineSnapshot(`
          {
            "AF": "0.5, DB, H2",
            "ALT": "A",
            "CHROM": "20",
            "DISTPREV": 100,
            "DISTPREVLOGE": 4.605170185988092,
            "DP": "14",
            "FILTER": "PASS",
            "ID": [
              "rs6054257",
            ],
            "INFO": "{\\"NS\\":[3],\\"DP\\":[14],\\"AF\\":[\\"0.5\\",\\"DB\\",\\"H2\\"]}",
            "MUTTYPE": "substitution",
            "NS": "3",
            "ORIGINALPOS": 14370,
            "POS": 137827,
            "POSEND": 137828,
            "QUAL": 29,
            "REF": "G",
            "SUBTYPE": "C>T",
          }
        `);
```

Once the snapshot is correct. Make a commit checking in the snapshot to repo. In subsequent test runs, `vitest` will assert that `tile` matches this snapshot exactly. This will prevent regressions from whatever we previously checked into the repo.

If you notice that something is wrong or need fixing, it either means we need to 1.) update the snapshot or 2.) we have a regression that needs fixing. This lets you refactor with much more confidence that you haven't changed the current behavior.

cc: @etowahadams, useful testing feature of vitest that might be good to know

sehilyi: This is really cool! Thanks for sharing this.

manzt: yeah! might be useful for testing a variety of spec transformations

PR #866, opened 2023-04-06
feat: upgrade to HiGlass 1.12.2

sehilyi:
Toward https://github.com/gosling-lang/gos/pull/122

## Change List
 - Upgrade to latest HiGlass (v1.12.2)
 - Make changes regarding the use of the latest d3-related APIs to reflect the breaking changes introduced in the recent HiGlass (e.g., https://github.com/higlass/higlass/pull/1111)
    - The circular brushes were not working in https://github.com/gosling-lang/gos/pull/122 because we are using HiGlass' d3 library (`HGC.libraries.d3Selection.event.sourceEvent`) for the circular brushes, but we were using a syntax that works with previous d3 version.

## Checklist


src/gosling-brush/linear-brush-model.ts
```diff
@@ -42,7 +42,7 @@ export class LinearBrushModel {
     private data: LinearBrushData;
 
     /* drag */
-    private startEvent: typeof D3Selection.event;
+    private startEvent: any; // D3Drag.D3DragEvent.sourceEvent which is `any`
```
manzt: could we just make this explicit?

```suggestion
    private startEvent: D3Drag.D3DragEvent.sourceEvent;
```

Although it is `any`, if it were to ever change we would automatically get the correct type.

manzt (approved):
Looks great! Just a minor comment.

manzt:
Do you think we will be able to make a release soon so we can also upgrade gos?

sehilyi:
I received a report that there is an error with `Buffer` when using bigwig files (https://github.com/gosling-lang/gosling.js/issues/876). I think we can release a version once we address it.

PR #874, opened 2023-04-13
docs: add documentation and refactor gosling track

etowahadams:
Fix #
Toward #875 

## Change List
 - Make fields and methods that are only used within `GoslingTrackClass` [private](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields)
 - Add documentation to some functions 
 - Remove unused function parameters 

## Checklist


src/core/mark/background.ts
```diff
@@ -1,17 +1,10 @@
-import type { Tile } from '../../gosling-track/gosling-track';
 import { isUndefined } from 'lodash-es';
 import type { GoslingTrackModel } from '../gosling-track-model';
 import { IsChannelDeep } from '../gosling.schema.guards';
 import colorToHex from '../utils/color-to-hex';
 import type { CompleteThemeDeep } from '../utils/theme';
 
-export function drawBackground(
-    HGC: import('@higlass/types').HGC,
```
etowahadams, line 9: `HGC` was an unused parameter

src/core/mark/index.ts
```diff
@@ -147,22 +138,18 @@ export function drawPreEmbellishment(
...
     (['x', 'x1', 'x1e', 'xe'] as const).forEach(d => {
-        // const c = tm.spec()[d as keyof typeof ChannelTypes];
-        // if(IsChannelDeep(c) && c.type === 'genomic') {
         model.setChannelScale(d, trackInfo._xScale);
-        // }
     });
 
-    if (CIRCULAR) {
-        drawCircularOutlines(HGC, trackInfo, tile, model, theme);
+    const isCircular = model.spec().layout === 'circular';
```
etowahadams, line 146: Rename `CIRCULAR` to `isCircular`

src/core/mark/outline-circular.ts
```diff
@@ -3,15 +3,8 @@ import { IsChannelDeep } from '../gosling.schema.guards';
 import { cartesianToPolar, valueToRadian } from '../utils/polar';
 import colorToHex from '../utils/color-to-hex';
 import type { CompleteThemeDeep } from '../utils/theme';
-import type { Tile } from '../../gosling-track/gosling-track';
 
-export function drawCircularOutlines(
-    HGC: import('@higlass/types').HGC,
```
etowahadams, line 9: Remove unused parameters

src/gosling-track/gosling-track.ts
```diff
@@ -133,54 +134,51 @@ const factory: PluginTrackFactory<Tile, GoslingTrackOptions> = (HGC, context, op
     /* Custom loading label */
     const loadingTextStyle = getTextStyle({ color: 'black', size: 12 });
 
+    /**
+     * The main plugin track in Gosling. This is a versetile plugin track for HiGlass which relies on GoslingTrackModel
+     * to keep track of mouse event and channel scales.
+     */
     class GoslingTrackClass extends BarTrack<Tile, typeof options> {
-        // This is tracking the xScale of an entire view, which is used when no tiling concepts are used
-        drawnAtScale = HGC.libraries.d3Scale.scaleLinear();
```
etowahadams, line 138: Did not s

src/gosling-track/gosling-track.ts
```diff
@@ -133,54 +134,51 @@ const factory: PluginTrackFactory<Tile, GoslingTrackOptions> = (HGC, context, op
...
-        scalableGraphics: Record<string, PIXI.Graphics> = {};
-        extent = { min: Number.MAX_SAFE_INTEGER, max: Number.MIN_SAFE_INTEGER };
```
etowahadams, line 140: Removed, was unused as far as I could tell!

src/gosling-track/gosling-track.ts
```diff
@@ -133,54 +134,51 @@ const factory: PluginTrackFactory<Tile, GoslingTrackOptions> = (HGC, context, op
...
-        private pMouseHover = new HGC.libraries.PIXI.Graphics();
-        private pMouseSelection = new HGC.libraries.PIXI.Graphics();
-        private isRangeBrushActivated = false;
-        private gBrush = HGC.libraries.d3Selection.select(context.svgElement).append('g');
-
-        private loadingTextStyleObj = new HGC.libraries.PIXI.TextStyle(loadingTextStyle);
-        private loadingTextBg = new HGC.libraries.PIXI.Graphics();
-        private loadingText = new HGC.libraries.PIXI.Text('', loadingTextStyle);
-
-        private svgData: unknown[] = [];
```
etowahadams, line 161: Removed, was unused

src/gosling-track/gosling-track.ts
```diff
@@ -133,54 +134,51 @@ const factory: PluginTrackFactory<Tile, GoslingTrackOptions> = (HGC, context, op
...
-        private textGraphics: unknown[] = [];
-        /** used to improve the performance of text rendering */
-        private textsBeingUsed = 0;
-        private loadingStatus = { loading: 0, processing: 0, rendering: 0 };
```
etowahadams, line 165: Removed, was unused

src/gosling-track/gosling-track.ts
```diff
@@ -133,54 +134,51 @@ const factory: PluginTrackFactory<Tile, GoslingTrackOptions> = (HGC, context, op
...
-
-        private prevVisibleAndFetchedTiles?: Tile[];
-        private oldDimensions?: [number, number];
-        private xDomain?: number[];
-        private xRange?: number[];
```
etowahadams, line 170: Removed, seemed unused

src/gosling-track/gosling-track.ts
```diff
@@ -516,36 +472,9 @@ const factory: PluginTrackFactory<Tile, GoslingTrackOptions> = (HGC, context, op
...
-
-            const tileK = drawnAtScaleExtent / xScaleExtent;
-            const newRange = xScale.domain().map(drawnAtScale);
-
-            const posOffset = newRange[0];
-            graphics.forEach(g => {
-                g.scale.x = tileK;
-                g.position.x = -posOffset * tileK;
-            });
-        }
```
etowahadams, line 537: Unused function

etowahadams:
When I tested things in the browser, I was getting a private field access error, and figured out that it was `pMouseHover` that should not be set to private. It was unexpected to me that `pMouseHover` needed to be a public field given that it was only called inside of `gosling-tracks.ts`. 

@sehilyi Is it obvious to you why it needed to be public? I didn't see any mentions if it in HiGlass either.

src/gosling-track/gosling-track.ts
```diff
@@ -133,54 +134,51 @@ const factory: PluginTrackFactory<Tile, GoslingTrackOptions> = (HGC, context, op
...
-
-        /** Store the color legends added so far so that we can avoid overlaps and redundancy */
-        private displayedLegends: DisplayedLegend[] = [];
+        /* *
+         *
+         *  Properties
+         *
+         * */
+
+        tileSize: number;
```
etowahadams, line 147: This has to be public or things break...but I don't see it being used in HiGlass

src/gosling-track/gosling-track.ts
```diff
@@ -133,54 +134,51 @@ const factory: PluginTrackFactory<Tile, GoslingTrackOptions> = (HGC, context, op
...
+        mRangeBrush: LinearBrushModel;
```
etowahadams, line 148: `mRangeBrush` strangely has to be public or things break

src/gosling-track/gosling-track.ts
```diff
@@ -133,54 +134,51 @@ const factory: PluginTrackFactory<Tile, GoslingTrackOptions> = (HGC, context, op
...
+        #processedTileInfo: Record<string, ProcessedTileInfo>;
+        // Used in mark/legend.ts
+        gLegend? = HGC.libraries.d3Selection.select(context.svgElement).append('g');
+        displayedLegends: DisplayedLegend[] = []; // Store the color legends added so far so that we can avoid overlaps and redundancy
+        // Used in mark/text.ts
+        textGraphics: unknown[] = [];
+        textsBeingUsed = 0;
+        // Mouse fields
+        pMouseHover = new HGC.libraries.PIXI.Graphics();
+        #pMouseSelection = new HGC.libraries.PIXI.Graphics();
```
etowahadams: Not sure why `pMouseSelection` doesn't have to be public when `pMouseHover` is public

src/gosling-track/gosling-track.ts
sehilyi: Thanks for removing these. I think we were using these three mainly to test different/performant rendering options which we do not use now, e.g., `scaleScalableGraphics()`.

src/core/gosling.schema.guards.ts
```diff
@@ -185,6 +185,9 @@ export function IsDataDeepTileset(
     );
 }
 
+/**
+ * Checks if a channel is of type ChannelDeep
+ */
 export function IsChannelDeep(channel: ChannelDeep | ChannelValue | undefined): channel is ChannelDeep {
     return typeof channel === 'object' && !('value' in channel);
```
manzt: Just having a look through is. I think `isObject` is now a guard we could re-use in this module.

```suggestion
    return isObject(channel) && !('value' in channel);
```

src/core/gosling-track-model.ts
```diff
@@ -862,7 +862,7 @@ export class GoslingTrackModel {
     }
 
     /**
-     * Set a new scale for a certain channel.
+     * Set a new scale for a certain channel. Usually called to Replace the scale of a genomic axis with the one that is generated by the HiGlass data fetcher.
```
manzt:
```suggestion
    * Set a new scale for a certain channel. Usually called to replace the scale of a genomic axis with the one that is generated by the HiGlass data fetcher.
```

manzt:
op, sorry noticing this is a draft. don't worry about my comments!

manzt:
> It was unexpected to me that pMouseHover needed to be a public field given that it was only called inside of `gosling-tracks.ts`.

Think this could be the culprit:

https://github.com/gosling-lang/gosling.js/blob/a1977b99c7b2687f1bee8990ce2ca18760572d4d/src/gosling-track/gosling-track.ts#L204-L205


`this.pMain` is coming from HiGlass

src/gosling-track/gosling-track.ts
sehilyi: It's odd that only one of them cannot be private. Does the IDE give you an error when you change it to private? It seems like I can change it to private on my vscode.

```ts
#pMouseHover = new HGC.libraries.PIXI.Graphics();
```

src/gosling-track/gosling-track.ts
```diff
@@ -1041,51 +949,34 @@ const factory: PluginTrackFactory<Tile, GoslingTrackOptions> = (HGC, context, op
...
-                Math.round(tileXScale.invert(this._xScale.invert(visible[1])))
-            );
-
-            return [start, end];
-        }
-
         /**
-         * Returns the minimum in the visible area (not visible tiles)
+         * Returns the minimum in the visible area (not visible tiles).
+         * Overrides method in Tiled1DPixiTrack
```
sehilyi: Thanks for adding this info throughout the code!

src/gosling-track/gosling-track.ts
```diff
@@ -390,31 +367,24 @@ const factory: PluginTrackFactory<Tile, GoslingTrackOptions> = (HGC, context, op
...
-            //     );
-            // }
-
             this.processAllTiles(true);
             this.draw();
             this.forceDraw();
         }
 
-        clearMouseEventData() {
+        private clearMouseEventData() {
```
sehilyi: Should we consider using `#` keywords for methods as well?

src/gosling-track/gosling-track.ts
etowahadams: Interestingly the IDE doesn't give me an error! I only noticed the error when I looked at the HiC matrix in the editor. @manzt is right though, that pMouseHover is being using in pMain which a field in the parent class ` this.pMain.addChild(this.pMouseHover);`. So should probably set `pMouseSelection` to public too.

src/gosling-track/gosling-track.ts
etowahadams: I didn't realize it could be used for methods too but I see now that it can. I think it would be a good idea. I'll change them now.  
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields#private_methods

src/gosling-track/gosling-track.ts
sehilyi: Ah, I see that matrix and `rule` examples break. Interesting that it still works with a `Mouse Events` example which actually uses `pMouseHover`.

src/gosling-track/gosling-track.ts
manzt: We could make this explicit and required throughout the code base with the [`--noImplicitOverride` complier option](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-3.html#override-and-the---noimplicitoverride-flag)

etowahadams: I didn't know this was a thing! I'll also add `overrides` to the ones in this track.

src/gosling-track/gosling-track.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
```
sehilyi: Do we need this? (Or did you mean to `eslint-enable` this?) I don't see `!` non-null assertion operator, but I may miss it.

sehilyi (approved):
Thanks for making this change! I tested locally, and all examples seem to work well.

src/gosling-track/gosling-track.ts
etowahadams: Will remove! I had just wanted to temporarily get rid of some warnings.

Issue #870, opened 2023-04-13
Support more data formats & options (e.g., FASTA, GFF, etc)


sehilyi:
Since the range of data types we supports is closely related to the usability of Gosling, we need to support additional data types that we discussed. These include

- #583
- #184 
- #764
- https://github.com/gosling-lang/gosling.js/issues/573
- Support loading multiple data fetchers in a single track (I somehow cannot find a dedicated issue).

Additionally, we can explore ways to support other cloud services:

- Google Cloud in IGV: https://googlegenomics.readthedocs.io/en/latest/use_cases/browse_genomic_data/igv.html

sehilyi:
We can also support BEDPE (https://bedtools.readthedocs.io/en/latest/content/general-usage.html#bedpe-format), and something to consider is predefined missing values, e.g., `chrom2` (`.`), `start2` (`-1`), `end2` (`-1`)

manzt:
As we implement more data-fetchers, I think it would be worth it to reduce the boilerplate/weirdness of implementing the HiGlass interfaces. Right now, we have `any` sprinkled about and could use a more type-safe wrappers that hide this weirdness consistently. Something I toyed around with this afternoon:

```typescript

// Represents the "modern" async interface we require all our loaders to implement
interface DataFetcher<Tile = unknown> {
    info(): Promise<Record<string, any>>;
    tiles(tileIds: string[]): Promise<Tile[]>;
}

// Represents the legacy callback interface that all HiGlass data fetchers implement
interface LegacyDataFetcher<Tile = unknown> {
    tilesetInfo(callback?: (info: unknown) => void): Promise<Record<string, any>>;
    fetchTilesDebounced(receivedTiles: (tiles: Record<string, Tile>[]) => void, tileIds: string[]): Promise<Tile[]>;
}

interface DataFetcherFactory<Args extends any[]> {
    (HGC: import('@higlass/types').HGC, ...args: Args): DataFetcher;
}

interface DataFetcherClass<Config, Args extends any[]> {
    new (HGC: import('@higlass/types').HGC, ...args: Args): LegacyDataFetcher;
    config: Config;
}

type DataFetcherArgs = [Record<string, unknown>];

function defineDataFetcher<Config>(
    config: Config,
    factory: DataFetcherFactory<DataFetcherArgs>
): DataFetcherClass<Config, DataFetcherArgs> {
[...]

PR #877, opened 2023-04-19
feat: add bed file datafetcher

### LINKED ISSUE #583: feat: Make BED v1 a primitive data format
### Motivation

BED (Browser Extensible Data) format provides a flexible way to define the data lines that are displayed in an annotation track. It has recently been formalized in the [v1 specification](https://github.com/samtools/hts-specs/pull/570). 

Gosling currently support BED via CSV, but it is quite verbose and users can define any field names they'd like for standard BED fields:

##### Specifying **BED12+1** in Gosling as CSV

```json
{
  "type": "csv",
  "url": "https://localhost:8080/data.bed",
  "headerNames": ["chrom", "chromStart", "chromEnd", "name", "score", "strand", "thickStart", "thickEnd", "itemRgb", "blockCount", "blockSizes", "myField"],
  "chromosomeField": "chrom",
  "genomicFields": ["chromStart", "chromEnd"],
  "quantitativeFields": ["score", "thickStart", "thickEnd", "blockCount"],
  "separator": "\t"
}
```

### Proposal

Add `BED` as a new data-type in Gosling. `BED` is designed for this exact use case, and should be the preferred format for representing text-based genomic annotation data (over a custom CSV capturing identical information). Using BED will make specifications less verbose and more reusable. Using BED has the additional side-effect of ensuring datasets behind a Gosling visualization are more likely to be interoperable with other genomics tools.

```typescript
interface BED {
  type: "bed";
  url: string;
  customFields?: string;
  separator?: string;
}
```

##### Specifying **BED12+1** in Gosling as CSV

```json
{
  "type": "bed",
  "url": "https://localhost:8080/data.bed",
  "customFields": ["myField"]
}
```

etowahadams:
Fix #583
Toward #870

## Change List
 - Add support for BED files
 - Refactor CSV data fetcher 

Test out using this gist: http://localhost:3000/?gist=etowahadams/cd49a7f308e4859e095d209056a564d3

[Documentation PR](https://github.com/gosling-lang/gosling-website/pull/63)

image

## Checklist


tsconfig.json
```diff
@@ -29,7 +29,7 @@
       "zlib": ["./src/alias/zlib.ts"]
     },
     "baseUrl": ".",
-    "types": ["vitest/globals", "vite/client"]
+    "types": ["vitest/globals", "vite/client", "d3-dsv"]
```
etowahadams: Had to add this so that `csv-data-fetcher` could get types from d3-dsv

manzt: Just had a look at the `package.json` and I don't see the types for d3-dsv included there. I think this also could be resolved with `yarn add -D @types/d3-dsv`. D3 doesn't ship it's own types, so we need to install them separately from the community (from DefinitelyTyped). This is the pattern we've used for the other d3 sub-packages.

etowahadams: I'll try this thanks!

package.json
```diff
@@ -155,7 +157,7 @@
     },
     "husky": {
         "hooks": {
-            "pre-commit": "run-p changelog schema schema-higlass schema-theme schema-template format && git add .",
```
etowahadams: Sometimes I only want to commit things that are staged. When the pre-commit hook would run, it would add all my unstaged files

src/data-fetchers/bed/bed-parser.ts
```diff
@@ -0,0 +1,136 @@
...
+"BED12"
+    (
+    string chrom;      "The name of the chromosome (e.g. chr3, chrY, chr2_random) or scaffold (e.g. scaffold10671)."
+    uint   chromStart; "The starting position of the feature in the chromosome or scaffold. The first base in a chromosome is numbered 0."
+    uint   chromEnd;   "The ending position of the feature in the chromosome or scaffold. The chromEnd base is not included in the display of the feature. For example, the first 100 bases of a chromosome are defined as chromStart=0, chromEnd=100, and span the bases numbered 0-99."
+    string   name;   "Defines the name of the BED line."
+    float   score;   "Feature score, doesn't care about the 0-1000 limit as in bed"
+    char   strand;   "Defines the strand. Either '.' (=no strand) or '+' or '-'"
+    uint thickStart; "The starting position at which the feature is drawn thickly (for example, the start codon in gene displays). When there is no thick part, thickStart and thickEnd are usually set to the chromStart position."
+    uint thickEnd; "The ending position at which the feature is drawn thickly (for example the stop codon in gene displays)."
```
etowahadams, line 22: Interestingly gmod/bed-js doesn't use this as their default schema, it just uses one with with the first three columns.

package.json
sehilyi: I see. Would the generated schema files and the changelogs still be included by the pre-commit hooks?

src/core/init.ts
```diff
@@ -69,6 +69,7 @@ export function init() {
      */
     for (const dataFetcher of Object.values(dataFetchers)) {
         const { config } = dataFetcher;
+        console.warn(dataFetcher);
```
sehilyi: Is this intended to be left here?

src/data-fetchers/bed/bed-worker.ts
```diff
@@ -0,0 +1,281 @@
...
+            this.#parser = await new BedParser(this.#customFields, n_columns);
+        }
+        return this.#parser;
+    }
+    /**
+     * Function to calculate the number of columns in the BED file. Needed when custom column names are supplied
+     * Relatively inefficient because the gmod/tabix API only has a way to retrieve lines based on genomic coordinates,
+     * but I believe this is better than fetching the BED file itself because then it only has to be fetched and
+     * unzipped once.
+     * @param uid A string which is the UID associated worker. Used to retrieve the chromosome information
```
sehilyi: This seems to be removed from the parameter at the end?

```suggestion
```

src/data-fetchers/bed/bed-worker.ts
```diff
@@ -0,0 +1,281 @@
...
+    }
+    /**
+     * Creates a parser. Parser cannot be created in the constructor because it relies on chromosome information
+     * which gets created by DataSource
+     * @returns A BED parser
+     */
+    async getParser() {
+        if (!this.#parser) {
+            const n_columns = this.#customFields ? await this.#calcNColumns() : undefined;
+            this.#parser = await new BedParser(this.#customFields, n_columns);
```
sehilyi: It looks like both parameters (i.e., `customFields` and `n_columns`) are either undefined or defined at the same time. If so, it would be better to make it more explicit, e.g.,

```diff
- constructor(customFields?: string[], n_columns?: number) {
+ constructor(opt?: { customFields: string[], n_columns: number }) {
```

src/data-fetchers/bed/bed-parser.ts
```diff
@@ -0,0 +1,136 @@
...
+            } else {
+                throw new Error(
+                    `Expected ${
+                        REQUIRED_COLS + this.#customFields.length
+                    } columns (${REQUIRED_COLS} required columns and ${
+                        this.#customFields.length
+                    } custom columns) but found ${this.#n_columns} columns`
+                );
+            }
+        }
```
sehilyi: It would be beneficial to add comments inside some of if statements here. It was a bit challenging to follow the logic.

src/data-fetchers/bed/bed-parser.test.ts
```diff
@@ -0,0 +1,91 @@
...
+                "chromEnd": 14510,
+                "chromStart": 11969,
+                "custom6": "+",
+                "custom7": "11868",
+                "name": "ENST00000456328.2",
+                "score": 0,
+                "strand": 0,
+              }
+            `);
+    });
```
sehilyi, line 90: Just out of curiosity, can this actually happen? (i.e., there are 12 columns defined in a BED file, but Gosling reads only the first few columns).

src/data-fetchers/bed/bed-parser.ts
sehilyi: How about simpler logic?

```suggestion
        if(this.#n_columns - this.#customFields.length < REQUIRED_COLS) {
            throw new Error(
                `Expected ${
                    REQUIRED_COLS + this.#customFields.length
                } columns (${REQUIRED_COLS} required columns and ${
                    this.#customFields.length
                } custom columns) but found ${this.#n_columns} columns`
            );
        } else if(this.#n_columns !== BED12Fields.length + this.#customFields.length) {
            throw new Error(`BED file error: unexpected number of custom fields. Found ${this.#n_columns} columns 
                    which is different from the expected ${BED12Fields.length + this.#customFields.length}`);
        } else {
            allFields = BED12Fields.slice(0, this.#n_columns - this.#customFields.length).concat(
                customFieldsWithTypes
            );
        }
```

src/data-fetchers/bed/shared-types.ts
```diff
@@ -0,0 +1,22 @@
...
+    itemRgb?: string;
+    blockCount?: number;
+    blockSizes?: number[];
+    blockStarts?: number[];
+    [customField: string]: string | number | number[] | undefined;
+}
+
+export interface EmptyTile {
+    tilePositionId: string;
+}
```
sehilyi: Is this used in somewhere?

sehilyi (approved):
This is awesome! I left several minor comments.

Also, do you want to add a simple example to our example gallery? That would be really great to include.

package.json
etowahadams: Ah you're right. I'll add it back.

src/core/init.ts
etowahadams: Will remove!

package.json
sehilyi: @manzt Do you aware of any better way to deal with this?

src/data-fetchers/bed/bed-parser.test.ts
etowahadams: This would happen when there is a custom column and the number of columns in the first row doesn't match the number of columns in the other rows. In this instance the column which contains "extra" would get ignored. 
```
chr1    23    234   custom
chr1    24    277   custom2    extra
```

src/data-fetchers/bed/bed-parser.ts
etowahadams: Much cleaner! Thank you.

src/data-fetchers/bed/shared-types.ts
etowahadams: It was being used in the bed fetcher. I think it'll be better if I move it this interface to the worker so I'll do that.

package.json
manzt: I think this is something we could probably check in CI? For example, we could have a GitHub action that runs a `package.json` script and assert that there are no git changes. If there are, fail and require the developer to run locally in order to pass CI. This way we allow much more flexibility when creating PRs, but ensure the changelog and schemas must be in sync in order to be merged into `master`.

I personally really dislike pre-commit hooks because I commit often.

sehilyi: I like the idea! I also dislike waiting for the pre-commit hooks to run scripts.

src/data-fetchers/bed/bed-worker.ts
etowahadams: I couldn't think of a good way to do this, that makes sense!

etowahadams:
Finished making edits, just need to add the example.

etowahadams:
Added example, ready to merge

PR #879, opened 2023-04-24
fix: alias lodash to lodash-es for smaller bundle

manzt:
Fix #
Toward #

I was just curious and ran `npx vite-bundler-vizualizer` to see what largest contributors to the gosling bundle are currently.

I was surprised to see `lodash` as such a large contributor (531kb).

Digging deeper, I discovered that `lodash` is not a direct dependency of Gosling (it's not listed in `package.json`), but `yarn`'s module scoping doesn't prevent tools like `typescript` or `vite` from digging into `node_modules` and finding `lodash` if it's used by a different dependency. This means that we ended up including `lodash` (a commonjs, non-treeshakeable module) in all of our builds.

The fix here makes the use of `lodash-es` consistent in the code base and removes this issue from the build.

## Change List
 - use only `lodash-es` in the source code (not `lodash`)

## Checklist

### DIFF
--- a/src/core/utils/omit-deep.ts
+++ b/src/core/utils/omit-deep.ts
@@ -1,9 +1,10 @@
 import type { GoslingSpec } from '@gosling.schema';
-import { cloneDeepWith, type CloneDeepWithCustomizer } from 'lodash';
+import { cloneDeepWith } from 'lodash-es';
+import { isObject } from '../gosling.schema.guards';
 
 export function omitDeep(spec: GoslingSpec, omitKeys: string[]) {
-    return cloneDeepWith(spec, (value: CloneDeepWithCustomizer<GoslingSpec>) => {
-        if (value && typeof value === 'object') {
+    return cloneDeepWith(spec, (value: unknown) => {
+        if (isObject(value)) {
             omitKeys.forEach(key => {
                 delete value[key];
             });
--- a/editor/example/json-spec/matrix.ts
+++ b/editor/example/json-spec/matrix.ts
@@ -1,6 +1,6 @@
 import type { PredefinedColors, SingleTrack } from '@gosling.schema';
 import type { GoslingSpec } from 'gosling.js';
-import { random } from 'lodash';
+import { random } from 'lodash-es';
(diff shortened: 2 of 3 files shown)


src/core/utils/omit-deep.ts
```diff
@@ -1,9 +1,10 @@
 import type { GoslingSpec } from '@gosling.schema';
-import { cloneDeepWith, type CloneDeepWithCustomizer } from 'lodash';
+import { cloneDeepWith } from 'lodash-es';
+import { isObject } from '../gosling.schema.guards';
 
 export function omitDeep(spec: GoslingSpec, omitKeys: string[]) {
-    return cloneDeepWith(spec, (value: CloneDeepWithCustomizer<GoslingSpec>) => {
```
manzt, line 5: The type of `CloneDeepWithCustomizer<GoslingSpec>` is as function (not an object), so previously the type was `never` in:

```typescript
        if (value && typeof value === 'object') {
            // value: never
            omitKeys.forEach(key => {
                delete value[key];
            });
```

So typescript never complained. This new typing narrows the type of value to `Record<PropertyKey, unknown>`, which is more likely what we wanted.

src/core/utils/omit-deep.ts
```diff
@@ -1,9 +1,10 @@
...
-        if (value && typeof value === 'object') {
+    return cloneDeepWith(spec, (value: unknown) => {
```
etowahadams, line 6: I didn't know about this `unknown` type, seems quite useful!

etowahadams (commented):
This vite-bundle-visualizer seems like a good tool to run periodically, thanks for doing this!

src/core/utils/omit-deep.ts
manzt: Yup, basically a [type-safe counterpart to `any`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#new-unknown-top-type). Anything can be assigned to `unknown`, but then you must work to convince the compiler that you can access particular properties. There's been some discussion about changing the return type of `Response.json()` to `Promise<unknown>`.

sehilyi (approved):

PR #893, opened 2023-05-05
ci: move pre-commit checks to CI

manzt:
Implements https://github.com/gosling-lang/gosling.js/pull/877#discussion_r1186531055

## Change List
 - move schema generation to a separate script (`scripts/generate-schemas.mjs`)

## Checklist


manzt:
Hmm, can't quite figure it out but now when I run `yarn changelog` locally, there are no changes (but it is working in CI).

.github/workflows/ci.yml
```diff
@@ -7,6 +7,21 @@ on:
...
+        name: Lint Repo
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v3
+            - uses: actions/setup-node@v3
+              with:
+                  node-version: 16
+                  cache: 'yarn'
+            - run: |
+                yarn install
```
manzt: I think ideally this action can run very quickly, and fail early (to prevent other jobs from running).

Maybe we could use `npx` here (rather than yarn installing everything for the repo).

scripts/generate-schemas.mjs
```diff
@@ -0,0 +1,34 @@
...
+  ["Theme", "theme.schema.json"],
+  ["TemplateTrackDef", "template.schema.json"],
+];
+
+const generator = tsj.createGenerator({
+  path: path.resolve(__dirname, "../src/index.ts"),
+  tsconfig: path.resolve(__dirname, "../tsconfig.json"),
+  skipTypeCheck: true,
+  encodeRefs: false,
+});
```
manzt, line 26: Running `ts-json-schema-generator` from the command line multiple times is unnecessarily expensive, since it needs to run TS and parse the AST each time for the whole Gosling code base. We can re-use the same schema generator for each of our schemas in this single script.

manzt:
Ah, actually I see what's going on. We needed to use `fetch-depth: 0` with the `actions/checkout` to get the whole git history in order for `conventional-changelog` to work. It relies on the previous tags to understand what need to be written to the CHANGELOG.md. 

Everything is working as it should now, it's just this action is _extremely_ slow because:

- It needs to checkout the entire gosling history to correctly run `conventional-changelog`, and the repo is massive (1.37GB) https://github.com/gosling-lang/gosling.js/issues/650
- It needs to install **all** of the yarn dependencies for gosling.js to run these commands (really it just needs contentional-changelog and ts-json-schema-generator)

In theory, this "Lint Repo" job should be extremely fast and fail early since it's so minimal, but as it stands it takes the longest out of all the other jobs (including tests and building the editor). 

Not sure how to proceed because although this is kind of nicer than pre-commit hooks, its very costly in CI.

EDIT: I guess 2.5 minutes isn't horrible, but reducing the size of the repo and making the yarn install much faster would probably make the job take seconds.

package.json
```diff
@@ -155,7 +153,6 @@
     },
     "husky": {
         "hooks": {
-            "pre-commit": "run-p changelog schema schema-higlass schema-theme schema-template format && git add .",
             "commit-msg": "commitlint -E HUSKY_GIT_PARAMS"
```
manzt: Also, since our "git flow" is squash & merge PRs, idk if it makes sense to lint every commit for conventional commits. 

Instead, something like a [GitHub action to lint the _PR title_](https://github.com/vitejs/vite/blob/main/.github/workflows/semantic-pull-request.yml) would probably make more sense

Not sure if either of you have thoughts on this: @etowahadams @sehilyi

.github/workflows/release.yml
```diff
@@ -26,9 +26,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+        with:
+            fetch-depth: 0
       - uses: actions/setup-node@v3
         with:
           node-version: 16
       - run: npx changelogithub
```
manzt, line 34: I saw this action was added in https://github.com/gosling-lang/gosling.js/pull/881. This action also won't work unless you set `fetch-depth: 0`.

package.json
sehilyi: It would make sense to use the `semantic-pull-request` you shared and remove the `commitlint` we use for every commit.

We were using [another semantic PR](https://github.com/zeke/semantic-pull-requests) which is deprecated and does not work anymore. We were using `commitlit` to handle a case where there is only one commit in a PR because in this case [GitHub used the commit message and ignored the PR title](https://github.com/refined-github/refined-github/issues/276). But, [using PR title by default is now supported](https://github.blog/changelog/2022-05-11-default-to-pr-titles-for-squash-merge-commit-messages/), so I don't think we need `commitlint` anymore.

sehilyi:
Regarding the costly CI, would the CI speed be more reasonable if we remove `conventional-changelog` from the CI? Since I think we can check `conventional-changelog` only at the release and not in every PR

.github/workflows/release.yml
sehilyi: Ah, got it! Thanks for finding this

package.json
etowahadams: I agree! No need to have every conventional commit name be linted.

sehilyi (approved):
I am open to merging this unless this will likely surpass the available usage of GitHub actions.

manzt:
I don't think we are likely to surpass available usage of GitHub actions (famous last words), but it does mean that currently it will take CI longer to pass (which is annoying for PRs).

I like the idea of only running the `contentional-changelog` in the release action. However, we then need to make sure the `CHANGELOG.md` is checked into the repo. Right now it will just be generated in the action and never added with `git commit`.

manzt:
Should be all good to merge. The main change now is that when you bump the version `npm version patch|minor|major`, the `package.json` "version" script will run, executing `conventional-changelog` and checking in the updated CHANGELOG.md.

The release "flow" is as described in the conventional-changelog docs: https://github.com/conventional-changelog/conventional-changelog/blob/master/packages/conventional-changelog-cli/README.md#with-npm-version

concretely,

```
npm version patch
git push --follow-tags
```

Issue #897, opened 2023-05-11
Make Gosling.js monorepo


sehilyi:
#### Related comments

- In both the editor and the document, examples can be consistent using, e.g., `gosling.js/examples`
- `gosling-theme` can be moved to here #860 
- We need to consider moving examples in `gosling-lang/gosling-react-example` to this repo under `src/examples`
> One thing to consider would be moving an examples/ folder to gosling.js and managing each of the examples (rather than every app needing react-dom-router) by @manzt

sehilyi:
As part of a post-1.0 cleanup of Gosling issues, we are closing stale issues that are no longer planned. If this is still relevant, feel free to reopen or start a discussion. Thanks!

PR #901, opened 2023-05-16
refactor: import from `index.ts` instead of internal files

sehilyi:
Toward #897 

## Change List
 - Use `@gosling-track`, `@gosling-mouse-event`, `@gosling-brush`, and `@data-fetchers` to import objects instead of their internal files (e.g., `import { XXX} from data-fetchers/utils.ts')
- `\core` folder is now big with many components contained. This can be divided into multiple folders, e.g., `schema`, `assembly`, `utils`, etc.

## Checklist


manzt:
Just came across this. Great to start working towards separate packages, but we probably want to adopt a convention for how they will be named on NPM. The `@` usage is specialized for a specific organization on NPM, so we would need create a different organization for each package as is.

I think adopting a convention with `@gosling-lang/*` would likely be preferable for when/if we refactor to a monorepo.

For example `@gosling-track` -> `@gosling-lang/track`. 

We could also remove `@gosling.schema` and `@higlass.schema` in favor of a `@gosling/schema` repo that is types-only package.

sehilyi:
I will use the `@gosling-lang/*` convention consistently!

> We could also remove @gosling.schema and @higlass.schema in favor of a @gosling/schema repo that is types-only package.

I assume you meant `@gosling-lang/schema`?

manzt:
yes!

Issue #909, opened 2023-05-24
Gosling - Higlass track API improvements


ThHarbig:
Currently, the gosling API provides a function to [retrieve the tracks](https://github.com/gosling-lang/gosling.js/blob/67eb1bad45468975dc6b7113aec35a6d051d7976/src/core/api.ts#L47-L49). This can be used, for example, to get the size of the tracks as a variable for other purposes, like aligning the gosling component with another custom component.

This function returns the tracks with their definition as higlass views. This has some issues:
- Overlaying tracks: In the Higlass spec there is the fields `"overlay":[]` within a view and `overlayOnPreviousTrack` to overlay tracks. Gosling tracks are sometimes transformed to elements in `"overlay":[]` of a higlass view. Therefore IDs that are set for gosling tracks cannot be found and the higlass ID for a view with the `"overlay":[]` field is undefined.
- I think, ideally, users would not get in contact with higlass specs at all when interacting with the API but they would be hidden. 

To improve this we could do something like this
- Make sure every higlass view has an ID
- Map between gosling track ID and highlass view ID
- Define what the information actually is that users need and provide this information either in `getTracks()` or implement other functions for this.

sehilyi:
> Map between gosling track ID and highlass view ID

The complexity comes from the fact that the structure of overlaid Gosling tracks are not consistent in the compiled HiGlass view for technical reasons (e.g., Gosling tracks in a circular view belong to the same HiGlass view, keeping only the first track ID at the end in HiGlass). This results in losing some of track IDs defined by users. I think we can keep IDs of every tracks at every level in the mapping table and connect them to the most relevant HiGlass view.

Example spec:
```ts
tracks: [
   { ..., id: '1' }, // ← becomes HiGlass view '1'
   { ..., alignment: 'overlay', tracks: [
       { ..., id: '2' }, // ← becomes HiGlass view '2'
       { ..., id: '3' }, // ← added to HiGlass view '2' (we lose ID currently in HiGlass)
   ]
]
```

Example mapping table:
```ts
{
   1: '1',
   2: '2',
   3: '2'
}
```

I think this mapping table will be most helpful in cases where we use HiGlass APIs internally, e.g., `gosRef.current.api.zoomToGene('3', 'MYC')`.

sehilyi:
We need to rename parameters/variables related to view IDs for the clarifications, refer to https://github.com/gosling-lang/gosling.js/pull/939#issuecomment-1636098302

PR #912, opened 2023-05-31
feat(api): rename API for getting track IDs

sehilyi:
Toward #909

## Change List
 - The IDs we use in the APIs correspond to Gosling track's ID, e.g., `{ mark: 'point', id: 'track-id-1', ...}`. This PR renames the API function from `getViewIds` to `getTrackIds`, making it consistent and more understandable. I left the `getViewIds` as is for backward compatibility, but at some point, it will be deprecated (added JS comment statements).

## Checklist


sehilyi:
I will close this and open another bigger PR that addresses this issue and also implement the mapping table @ThHarbig 

I found that would be more ideal.

PR #898, opened 2023-05-11
feat: support React 18

sehilyi:
## Change List
 - Upgrade HiGlass to support the latest React. This does not solve the entire dependency issue (#852), but it will still be a helpful workaround for users who want to use v18 React.
 - Replace `react-split-pane` with `allotment` (`react-split-pane` is not well maintained and having dependency issues with latest React types)

Screenshot 2023-05-11 at 17 42 53

## Checklist
 - [x] Ensure the PR works with all demos on the online editor
 - [ ] Unit tests added or updated
 - [ ] Examples added or updated
 - [ ] [Documentation](https://github.com/gosling-lang/gosling-website) updated (e.g., added API functions)
 - [ ] Screenshots for visual changes (e.g., new encoding support or UI change on Editor)

PR #913, opened 2023-05-31
fix: revert breaking React 18 changes

manzt:
Changes from #898 are breaking to end users who use React 16 or 17 and `gosling.embed` (e.g., Gos). The peer-dependencies in Gosling allow for `react-dom` v16/17/18, but only `react-dom` v18 has `react-dom/client`. The top-level `ReactDOM.render()` will eventually be [deprecated](https://react.dev/reference/react-dom/render), but it's totally fine to use for the moment.

This PR reverts use of `react-dom/client` (and `createRoot`), which is specific to react-dom v18, for the backward compatible `ReactDOM.render`.

I'm happy to discuss narrowing support for React versions in Gosling, but the previous changes seemed a bit too restrictive and would likely break many existing Gosling usage for folks who allow for upgrading patches in their dependencies.

## Change List
 - Reverts use of `react-dom/client` to `react-dom` to avoid breaking change in patch release.

## Checklist

### DIFF
--- a/editor/index.tsx
+++ b/editor/index.tsx
@@ -1,13 +1,13 @@
 import React from 'react';
-import ReactDOM from 'react-dom/client';
+import ReactDOM from 'react-dom';
 import { BrowserRouter, Route } from 'react-router-dom';
 import Editor from './Editor';
 import './index.css';
 import 'higlass/dist/hglib.css';
 
-const root = ReactDOM.createRoot(document.getElementById('root') as HTMLElement);
-root.render(
+ReactDOM.render(
     <BrowserRouter>
         <Route component={Editor} />
-    </BrowserRouter>
+    </BrowserRouter>,
+    document.getElementById('root') as HTMLElement
 );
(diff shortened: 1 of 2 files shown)


editor/index.tsx
```diff
@@ -1,13 +1,13 @@
...
 import 'higlass/dist/hglib.css';
 
-const root = ReactDOM.createRoot(document.getElementById('root') as HTMLElement);
-root.render(
+ReactDOM.render(
     <BrowserRouter>
         <Route component={Editor} />
-    </BrowserRouter>
+    </BrowserRouter>,
+    document.getElementById('root') as HTMLElement
```
sehilyi, line 12: Should this also have to be reverted?

sehilyi (approved):
Ah, I apologize. I overlooked the case of using lower React versions. I will cut the version upon merging this.

editor/index.tsx
manzt: It doesn't have to be, but my thinking was to be consistent in the editor with how we expect most will consume Gosling.js package. When we upgrade react (and it does break in the future), the editor will also break, signaling we need to narrow the react compatibility versions.

In other words, there's no tests in place to ensure `gosling-embed.ts` works. It would probably be useful to do so, but unifying how we use react in the library in the editor could help surface issues faster.

sehilyi: Make sense. I am happy with the current version.

PR #914, opened 2023-05-31
feat: preserve track IDs for JS API

sehilyi:
Toward #909 

## Change List
 - Rename `getViewIds()` API to `getTrackIds()`, consistent to terms used in Gosling. "View" was a HiGlass concept that we want to hide from users.
 - Preserve track IDs for using JS API functions (e.g., `zoomTo()`). Previously, some IDs of individual overlaid tracks were lost in compiled HiGlass views (because we only took the first ID, ignoring the rest of IDs). This PR now preserves an ID mapping table.

```ts
{ alignment: 'overlay', tracks: [ 
   { id: 'a', ...}, // we only took this previously
   { id: 'b', ...} // previously lost in a HiGlass view
]}
```

mapping table:

```ts
{ 'a': 'a', 'b', 'a' }
```

## Checklist


src/gosling-higlass/id.ts
```diff
@@ -0,0 +1,36 @@
...
+    /** A mapping table between Gosling track IDs to HiGlass view IDs */
+    #mappingTable: Record<string, string> = {};
+
+    getMappingTable() {
+        return this.#mappingTable;
+    }
+    addMapping(gtId: string, hvId: string) {
+        if (this.#mappingTable[gtId]) {
+            console.warn('The given track ID already exists.');
+        }
```
ThHarbig, line 16: This warning appears every time, multiple times for all the examples that I tried in the editor, is this intentional?

src/core/api.ts
```diff
@@ -21,11 +20,19 @@ interface GeneSuggestion {
...
+     * This is for the backward compatibility and will be deprecated. Use `getTrackIds()` instead.
+     */
     getViewIds(): string[];
+    /**
+     * Get an array of all available track IDs that are either specified by users or generated by the compiler.
+     * This can be used to call other API functions, e.g., `getTrack('track-1')`.
+     */
+    getTrackIds(): string[];
     getTracks(): TrackMouseEventData[];
     getTrack(trackId: string): TrackMouseEventData | undefined;
```
ThHarbig, line 37: For some reason there is more IDs in getTrackIds() than the number of tracks in getTracks(). I think for overlaid tracks the first track is not included in getTracks().

src/gosling-track/gosling-track.ts
```diff
@@ -480,7 +491,7 @@ const factory: PluginTrackFactory<Tile, GoslingTrackOptions> = (HGC, context, op
 
             const flatTileData = ([] as Datum[]).concat(...models.map(d => d.data()));
             if (flatTileData.length !== 0) {
-                publish('rawData', { id: context.viewUid, data: flatTileData });
+                this.options.siblingIds.forEach(id => publish('rawData', { id, data: flatTileData }));
```
ThHarbig, line 494: This basically means that the rawData event is fired multiple times with the same data for different track Ids?

ThHarbig (commented):
I think there is something wrong with overlaid tracks. It seems like the first track in a set of overlaid tracks is not returned by getTracks()

src/core/api.ts
sehilyi: Thanks for catching this. They should be consistent! Let me double-check.

src/gosling-track/gosling-track.ts
ThHarbig: This seems to be resolved!

sehilyi:
I will open another PR for this!

PR #923, opened 2023-07-03
feat(data-fetcher): GFF3 with tabix

### LINKED ISSUE #764: Support for GFF
Several folks at ISMB asked about support for [GFF](https://en.wikipedia.org/wiki/General_feature_format) for gene annotation data. It's well supported by many [genome browsers](https://jbrowse.org/docs/tutorial.html) and would be useful in Gos (probably indexed). 

One thing I'm thinking about Gos is to automatically index indexable files for the visualization if they are missing.

etowahadams:
Fix #764 
Toward #870 

## Change List
 - Add support for GFF3 

Human genome: http://localhost:3000/?gist=etowahadams/380ada008a2c39077065793f41446caf
E coli genome: http://localhost:3000/?gist=etowahadams/4caa11352a4da2c299f7869d1c3684ea

[Documentation PR](https://github.com/gosling-lang/gosling-website/pull/66)

image

image

## Checklist
 - [x] Examples added or updated


etowahadams:
Currently, each tile contains all of the data parsed from a given line. 

For example, the following line gets parsed as follows. Note that the attributes column has the format `tag=value;`, which gets parsed into an object. 
```
U00096.3	Genbank	gene	352706	354592	.	+	.	ID=gene-b0335;Dbxref=ASAP:ABE-0001154,ECOCYC:G6200;Name=prpE;gbkey=Gene;gene=prpE;gene_biotype=protein_coding;gene_synonym=ECK0332,yahU;locus_tag=b0335
```
```
{
  Name: "prpE"
  ​​​attributes: Object { ID: (1) […], Dbxref: (2) […], Name: (1) […], … }
  ​​​child_features: Array []
  ​​​derived_features: Array []
  ​​​end: 354592
  ​phase: null
  ​​​score: null
  ​​​seq_id: "U00096.3"
  ​​​source: "Genbank"
  ​​​start: 352706
  ​​​strand: "+"
  ​​​type: "gene"
  }
```

When `extractAttributes` is set to `true`, then the keys and values in the attributes object become added the parent object. This is useful because often times you want to color or name things based on an attribute, and now it can be accessed as a field. 

```
{
  Dbxref: Array [ "ASAP:ABE-0001154", "ECOCYC:G6200" ]
  ID: "gene-b0335"
  Name: "prpE"
  attributes: Object { ID: (1) […], Dbxref: (2) […], Name: (1) […], … }
  child_features: Array []
  derived_features: Array []
  end: 354592
  gbkey: "Gene"
  gene: "prpE"
  gene_biotype: "protein_coding"
  gene_synonym: Array [ "ECK0332", "yahU" ]
  locus_tag: "b0335"
  phase: null
  score: null
  seq_id: "U00096.3"
  source: "Genbank"
  start: 352706
  strand: "+"
  type: "gene"
}
```
This creates an interesting problem though. Because not all features have the same attributes, if `extractAttributes` is `true`, then not all tile objects have the same keys. This can lead to some unexpected bugs, such as when coloring a mark by an extracted attribute value, where all tiles are expected to share the same field. This raises a question: do we want to preserve that all tiles have the same fields, or should we allow tiles have different number of fields? 

I think the ideal answer is that we require that all tiles have the same number of fields, but this seems unrealistic. In the case of the GFF file, we will not know what all of the possible attributes are unless we parse the entire file. And parsing the entire file is something we want to avoid for large GFF files. So the easier solution is to have tiles that have differing fields, or have users define which attributes they would like to have extracted. 

What do you think of the idea to define which attributes get added to the parent object? If only GFF was a true columnar data format! These issues would be avoided.

src/data-fetchers/gff/gff-worker.ts
```diff
@@ -0,0 +1,354 @@
...
+                    Object.keys(attributes).forEach(key => {
+                        const attVal = attributes[key];
+                        if (Array.isArray(attVal)) {
+                            cleanAtt[key] = attVal.length == 1 ? attVal[0] : attVal;
+                        }
+                    });
+                    return { ...tile, ...cleanAtt };
+                });
+            }
+            console.warn(tiles);
```
sehilyi:
```suggestion

```

src/core/gosling.schema.ts
```diff
@@ -1101,6 +1102,28 @@ export interface BamData {
...
+
+    /** URL link to the tabix index file */
+    indexUrl: string;
+
+    /** The maximum number of rows to be loaded from the URL. __Default:__ `1000` */
+    sampleLength?: number;
+
+    /** Determines whether to extract the feature attributes, which have the format in the format `tag=value` in column 9
+     * If true, the tag will become available as a field. If there is an existing field with the same name, then
+     * if will be overwritten by the attribute tag. __Default:__ `false`
```
sehilyi:
```suggestion
     * it will be overwritten by the attribute tag. __Default:__ `false`
```

src/core/gosling.schema.ts
```diff
@@ -1101,6 +1102,28 @@ export interface BamData {
...
+    /** URL link to the GFF file */
+    url: string;
+
+    /** URL link to the tabix index file */
+    indexUrl: string;
+
+    /** The maximum number of rows to be loaded from the URL. __Default:__ `1000` */
+    sampleLength?: number;
+
+    /** Determines whether to extract the feature attributes, which have the format in the format `tag=value` in column 9
```
sehilyi: Is it possible that `value` is an array of items instead of a single item?

src/core/gosling.schema.ts
```diff
@@ -1101,6 +1102,28 @@ export interface BamData {
...
+export interface GffData {
+    type: 'gff';
+    /** URL link to the GFF file */
+    url: string;
+
+    /** URL link to the tabix index file */
+    indexUrl: string;
+
+    /** The maximum number of rows to be loaded from the URL. __Default:__ `1000` */
+    sampleLength?: number;
```
sehilyi, line 1119: Could you add a little more description about how the rows are sampled (e.g., randomly sampled)? I learned recently that this is important information for users.

src/data-fetchers/gff/gff-worker.ts
```diff
@@ -0,0 +1,354 @@
...
+
+            const totalBins = Math.floor(diff / windowSize);
+            const selectedBins = makeRandomSortedArray(n_samples, totalBins);
+            const binLines = selectedBins.map(bin => {
+                const binStart = minX + bin * windowSize;
+                const binEnd = binStart + windowSize;
+                return this.#getLinePromises(binStart, binEnd);
+            });
+            return binLines.flat();
+        }
```
sehilyi, line 98: This is interesting logic. My understanding is that (compared to sampling data at the very end after parsing entire lines) this basically controls the number of lines parsed so that the data fetcher does not spend much time parsing an overly large number of lines that will be mostly unused at the end.

As a side note, it would be an interesting/useful side project to benchmark data loading for vis. given that we now have more number of data loaders.

src/core/gosling.schema.ts
sehilyi: Maybe clarify here what the parsed data look like.

sehilyi:
> What do you think of the idea to define which attributes get added to the parent object?

I like the idea! In general, it looks to me that the `attributes` will be useful data to show in visualization, and making such data easier to use seems desirable. I think we can let users define what attributes to add, and we can probably also allow them to specify default values to fill in if missing.

sehilyi (commented):
Looks great to me, and my comments are mostly about clarifying user parameters that will be shown as a part of the documentation.

One thing to consider is how we want to deal with `Array` in the child attributes. It looks like the parsed values with multiple items will be shown as something like `[object Object]` (e.g., in a tooltip). It might be better to at least make this a string value, e.g., `ECK0332, yahU` instead of `["ECK0332", "yahU"]` so that it can be used as a text label or shown in a tooltip.

src/data-fetchers/gff/gff-worker.ts
etowahadams: Your understanding is correct! I discovered that when I zoomed out, it would take forever for all of the lines to get fetched and parsed. And most of the parsed lines wouldn't be actually displayed anyway. 

My goal was to reduce the number of lines that need to be loaded into memory and parsed. To do that, I randomly sampled a hundred (`n_samples`) 50 kilobase windows  (`windowsSize`) from the entire region that I wanted to display. I found empirically this was a good tradeoff. It would be interesting to benchmark all of the options. I agree I think that could be a fun side project.

etowahadams:
> I think we can let users define what attributes to add, and we can probably also allow them to specify default values to fill in if missing.

I replaced the `extractAttributes` option with a new option called `attributesToFields` which allows for users to specify which attributes they want accessible, and what the default value for that attribute should be. I think this is a lot cleaner and helps avoid the problem of unintentionally overriding default field names. 

> It looks like the parsed values with multiple items will be shown as something like [object Object] (e.g., in a tooltip). It might be better to at least make this a string value, e.g., ECK0332, yahU

Added! And updated schema documentation to reflect this change. 

Also check out my new E coli genome visualization! After a lot of head scratching I finally figured out how to color by gene type. It was because the gosling gene template applies a data transform to filter out non "-" or "+" values, so I couldn't color by values that were not among those. So the solution was not to use the template. 

http://localhost:3000/?gist=etowahadams/4caa11352a4da2c299f7869d1c3684ea

src/core/gosling.schema.ts
etowahadams: Added!

sehilyi (approved):
The changes in this repo and the docs look great! And I like your new visualization and its colors

PR #933, opened 2023-07-07
ci: specify allowed scopes in semantic-pull-request.yml

sehilyi:
Specify the allowed scopes for the PR title so that we can make sure to use scope names consistently.

I have added five categories for `feat` and `fix` updates

- `core`: any general updates related to the library, such as the grammar and new rendering
- `track: updates specific to tracks in this library, such as gosling track and axis
- `data-fetcher`: any updates related to data fetchers
- `editor`: UI and other updates to the online editor
- `api`: the Gosling APIs

Usage:

```sh
feat(core): added A, B, C to the grammar
fix(data-fetcher): fix the bam data fetcher to ...
```

~~Feel free to suggest adding a new category. Also, maybe we can set `requireScope` to `true` and add a general scope (e.g., `core`) so that we make sure not to specify a proper scope. I found that I did not put a scope when I fixed the bam data fetcher!~~

### DIFF
--- a/.github/workflows/semantic-pull-request.yml
+++ b/.github/workflows/semantic-pull-request.yml
@@ -18,8 +18,10 @@ jobs:
         with:
           # Configure which scopes are allowed.
           scopes: |
+            core
+            track
             data-fetcher
             api
             editor
           # Configure if a scope must always be provided.
-          requireScope: false
+          requireScope: true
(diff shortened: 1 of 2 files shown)


etowahadams (approved):
Looks good to me! I just changed the order so that information about commits/PRs comes up earlier in the doc. I think that's a little more logical? Going from the general advice to the specific. 

I agree about making the scopes required. I think `core` is a good addition. If we continue this pattern, perhaps we should have a `track` scope as well.

sehilyi:
Thank you for the suggestion! I made further changes so that

- a scope is forced for `feat` and `fix` (and disallowed for other types of updates)
- added `core` and `track` as you suggested
- Updated the CONTRIBUTING.md document

Feel free to update this rule as you work on the repo!

PR #938, opened 2023-07-11
fix(data-fetcher): Bump tabix-js version to fix errors when loading BED files

### LINKED ISSUE #936: BED datafetcher works locally but not when deployed
BED datafetcher is not working properly because tabix-js isn't using explicit Buffer imports. I just submitted a [PR](https://github.com/GMOD/tabix-js/pull/140) to tabix-js. So hopefully its resolved soon. 

I happened to look at the [deployed BED data fetcher](https://gosling.js.org/?example=doc_bed) and noticed it wasn't working properly. The data does not load and the following error shows up in the console: 
image

However, everything seems to work fine when I test things locally. 

I believe this is the same issue that was encountered before with [Buffer needing to be polyfilled](https://github.com/GMOD/generic-filehandle/pull/102), thus necessitating to explicit Buffer imports.

etowahadams:
Fix #936
Toward #

## Change List
 - Updated to most recent version of [tabix-js](https://github.com/GMOD/tabix-js). This version uses explicit Buffer imports, which allows the Buffer polyfill to be used. 

Tested using the deployment version. Thanks for the advice Sehi! 
 
image

## Checklist

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -334,7 +334,17 @@
   resolved "https://registry.yarnpkg.com/@gmod/bed/-/bed-2.1.2.tgz#02e27a3a75269dec06ecc22f0ff7eb10b8affc42"
   integrity sha512-LnCmA+jb0xfbSWO7isi1dVvqbQi8Icqaj8FeUcnCc8t4jiNe1eFoe1YU8Chn7a8EDGFqS06kMNpO1vodO+q4IA==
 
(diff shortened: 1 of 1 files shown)


sehilyi (approved):

sehilyi:
I will merge this and make a patch. And the new features (e.g., GFF & APIs) will be included in a new minor version.

Issue #183, opened 2021-02-05
Allow showing arrangement preview in the editor


sehilyi:
![image](https://user-images.githubusercontent.com/9922882/107083339-78ecca80-67c3-11eb-84c1-cc526aaf05b2.png)

As @ngehlenborg suggested, it would be helpful to show how tracks are arranged in the output visualization in the editor.

We can add a toggle button on the title bar of the editor and turning it on will overlay rectangles and text labels on top of the Gosling visualization for such visual information.

sehilyi:
Should work w/ circular layouts as well.

sehilyi:
This is now supported via the "Show Views" button.

PR #939, opened 2023-07-11
feat(core, api, editor): support using  view IDs

### LINKED ISSUE #929: Idea: add IDs to views
Currently, only tracks have IDs, but I think it might be helpful to associate views with ID, for example, to get a view by its ID and retrieve its height.

sehilyi:
Fix #929 
Toward #183

## Change List
 - In the grammar, enable adding an `id` to a view at any level.
 - Using the `id`, users can get the bounding boxes and specs of views as well as tracks.
 - In the Editor, add a toggle button to display the hierarchy of tracks and views.

### APIs Added/Updated
```ts
const tracksAndViews = gosRef.api.current.getTracksAndViews();
const views = gosRef.api.current.getViews();
const tracks = gosRef.api.current.getTracks();

console.log(tracksAndViews[0]);

// { type: 'track', shape: { x: 0, y: 12, width: 400, height: 100 }, spec: { id: 'track-1', mark: 'point', ...} }
```

### Screenshot

The blue rectangles represent views, and the red rectangles represent tracks.

Screenshot 2023-07-13 at 12 16 13

Screenshot 2023-07-13 at 13 04 25

Screenshot 2023-07-13 at 13 04 35

## Checklist

### DIFF
--- a/src/core/api.ts
+++ b/src/core/api.ts
@@ -1,5 +1,5 @@
 import * as PIXI from 'pixi.js';
-import type { TrackMouseEventData } from '@gosling.schema';
+import type { TrackMouseEventData, ViewApiData } from '@gosling.schema';
 import type { HiGlassApi } from './higlass-component-wrapper';
 import type { HiGlassSpec } from '@higlass.schema';
 import { subscribe, unsubscribe } from './pubsub';
@@ -28,6 +28,8 @@ export interface GoslingApi {
     getViewIds(): string[];
     getTracks(): TrackMouseEventData[];
     getTrack(trackId: string): TrackMouseEventData | undefined;
+    getViews(): ViewApiData[];
+    getView(viewId: string): ViewApiData | undefined;
     exportPng(transparentBackground?: boolean): void;
     exportPdf(transparentBackground?: boolean): void;
     getCanvas(options?: { resolution?: number; transparentBackground?: boolean }): {
@@ -42,6 +44,7 @@ export function createApi(
     hg: Readonly<HiGlassApi>,
     hgSpec: HiGlassSpec | undefined,
     trackInfos: readonly TrackMouseEventData[],
+    views: readonly ViewApiData[],
     theme: Required<CompleteThemeDeep>
 ): GoslingApi {
     const getTracks = () => {
@@ -54,6 +57,16 @@ export function createApi(
         }
         return trackInfoFound;
     };
+    const getViews = () => {
+        return [...views];
+    }
+    const getView = (viewId: string) => {
+        const view = getViews().find(d => d.id === viewId);
+        if (!view) {
+            console.warn(`Unable to find a view with the ID of ${viewId}`);
+        }
+        return view;
+    }
     const getCanvas: GoslingApi['getCanvas'] = options => {
         const resolution = options?.resolution ?? 4;
         const transparentBackground = options?.transparentBackground ?? false;
@@ -120,6 +133,8 @@ export function createApi(
         },
         getTracks,
         getTrack,
+        getView,
+        getViews,
         getCanvas,
         exportPng: transparentBackground => {
             const { canvas } = getCanvas({ resolution: 4, transparentBackground });
--- a/src/core/api.ts
+++ b/src/core/api.ts
@@ -1,5 +1,5 @@
 import * as PIXI from 'pixi.js';
-import type { TrackMouseEventData, ViewApiData } from '@gosling.schema';
+import type { TrackApiData, VisUnitApiData, ViewApiData } from '@gosling.schema';
 import type { HiGlassApi } from './higlass-component-wrapper';
 import type { HiGlassSpec } from '@higlass.schema';
 import { subscribe, unsubscribe } from './pubsub';
@@ -25,9 +25,10 @@ export interface GoslingApi {
     zoomToExtent(viewId: string, duration?: number): void;
     zoomToGene(viewId: string, gene: string, padding?: number, duration?: number): void;
     suggestGene(viewId: string, keyword: string, callback: (suggestions: GeneSuggestion[]) => void): void;
+    getTracksAndViews(): VisUnitApiData[];
     getViewIds(): string[];
-    getTracks(): TrackMouseEventData[];
-    getTrack(trackId: string): TrackMouseEventData | undefined;
+    getTracks(): TrackApiData[];
+    getTrack(trackId: string): TrackApiData | undefined;
     getViews(): ViewApiData[];
     getView(viewId: string): ViewApiData | undefined;
     exportPng(transparentBackground?: boolean): void;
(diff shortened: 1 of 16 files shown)


src/core/utils/bounding-box.ts
```diff
@@ -16,13 +16,6 @@ export interface Size {
     height: number;
 }
 
-export interface GridInfo extends Size {
-    columnSizes: number[];
-    rowSizes: number[];
-    columnGaps: number[];
-    rowGaps: number[];
-}
```
sehilyi, line 24: Not used anymore.

sehilyi:
- [ ] I still need to update the documentation

src/core/api.ts
```diff
@@ -25,9 +25,12 @@ export interface GoslingApi {
     zoomToExtent(viewId: string, duration?: number): void;
     zoomToGene(viewId: string, gene: string, padding?: number, duration?: number): void;
     suggestGene(viewId: string, keyword: string, callback: (suggestions: GeneSuggestion[]) => void): void;
+    getTracksAndViews(): VisUnitApiData[];
     getViewIds(): string[];
```
etowahadams, line 29: Do we want to keep this API method? It could be confusing because `getViewIds` returns the `view.uid` created by HiGlass, not `view.id` as the name would suggest.

etowahadams, line 29: Do we want to keep the `getViewIds` API method? It could be confusing because `getViewIds` returns the HiGlass view `viewIds`, not the ids of Gosling views.

src/core/gosling.schema.ts
```diff
@@ -155,9 +157,8 @@ export interface CommonViewDef {
 export type Track = SingleTrack | OverlaidTrack | DataTrack | TemplateTrack;
 
 export interface CommonTrackDef extends CommonViewDef {
-    // !! TODO: we can check if the same id is used multiple times.
-    // !! TODO: this should be track-specific and not defined in views.
-    id?: string; // Assigned to `uid` in a HiGlass view config, used for API and caching.
+    /** Assigned to `uid` in a HiGlass view config, used for API and caching. */
+    id?: string;
```
etowahadams, line 161: Since `CommonViewDef` also now has an `id?` property, should this be removed?

src/core/gosling.schema.ts
```diff
@@ -254,10 +255,10 @@ interface RangeMouseEventData extends CommonEventData {
 }
 
 /**
- * The visual parameters that determine the shape of a linear track.
+ * The visual parameters that determine the shape of a linear track or a view.
  * Origin is the left top corner.
  */
-interface LinearTrackShape {
```
etowahadams, line 260: Glad this name got changed :)

src/core/track-and-view-ids.ts
etowahadams, line 1: `src/core` is starting to get quite cluttered. Do you think it would make sense to have a folder called `api` which contains `api.ts`, `track-and-view-ids.ts` and maybe `pubsub.ts`? Perhaps a question for a larger reorganization conversation.

src/core/gosling.schema.ts
etowahadams: I think I see how the two are different--is my understanding correct? 
- `track.id`:  Set in the schema by the user, or is generated by HiGlass. An instance of GoslingTrackClass knows its `id` by calling `context.viewUid`
- `view.id` Set by the user in the schema or is generated by Gosling

src/core/utils/spec-preprocess.ts
```diff
@@ -187,6 +192,11 @@ export function traverseToFixSpecDownstream(spec: GoslingSpec | SingleView, pare
 
         const linkID = uuid.v4();
         tracks.forEach((track, i, array) => {
+            // ID should be assigned to each view and track for an API usage
+            if (!track.id) {
+                track.id = uuid.v4();
+            }
+
```
etowahadams, line 199: Is this equivalent to what happens in the `goslingToHiGlass` function? 

https://github.com/gosling-lang/gosling.js/blob/526882fb82463ad4262568d9569840431c709fc9/src/core/gosling-to-higlass.ts#L38-L40

src/core/gosling.schema.ts
etowahadams, line 161:
```suggestion

```
Since `CommonViewDef` also now has an `id?` property, should this be removed?

etowahadams (approved):

etowahadams:
To clarify some terms and check my understanding: 
- **Gosling track ID**: ID that is associated with each track. It is set by the user or is autogenerated in Gosling. 
Autogenerated here unless set in the schema (although I think the new location for track ID generation makes more sense) :https://github.com/gosling-lang/gosling.js/blob/526882fb82463ad4262568d9569840431c709fc9/src/core/gosling-to-higlass.ts#L38-L40 
- **Gosling view ID**: This is a new concept. Each Gosling view now has an ID.  
- **HiGlass track ID**: The HiGlass track ID is the same as the Gosling track ID with `-track` appended to it. 
https://github.com/gosling-lang/gosling.js/blob/526882fb82463ad4262568d9569840431c709fc9/src/core/gosling-to-higlass.ts#L76
- **HiGlass view ID**: Each HiGlass view has an ID, and *this ID is the same as the Gosling track ID.* An instance of GoslingTrackClass knows what the Gosling track ID is by calling `context.viewUid`, which is the HiGlass view ID which is the same as the Gosling track ID. The HiGlass view ID is set in this line
https://github.com/gosling-lang/gosling.js/blob/526882fb82463ad4262568d9569840431c709fc9/src/core/gosling-to-higlass.ts#L175


There are many variables called `viewId` in Gosling currently, and most of the refer HiGlass view IDs (and indirectly Gosling track IDs), not Gosling view IDs. I think it would be useful to distinguish between these two. 

For example, the usage of `viewId` here refers to HiGlass view IDs. I think we should change the variable name to be `trackId` to refer to Gosling track IDs? 
https://github.com/gosling-lang/gosling.js/blob/526882fb82463ad4262568d9569840431c709fc9/src/core/api.ts#L24-L27

And another usage of `viewId`, which is referring a HiGlass view ID (I think?)  

https://github.com/gosling-lang/gosling.js/blob/526882fb82463ad4262568d9569840431c709fc9/src/core/utils/linking.ts#L12

src/core/api.ts
sehilyi: Ah, good catch! I tried to change the name to `getTrackIds` in another PR (https://github.com/gosling-lang/gosling.js/pull/914) to make any HiGlass concept hidden from the users. I will rename it in this PR.

sehilyi: Actually, let me change that in another PR. I will work on a PR to address https://github.com/gosling-lang/gosling.js/issues/909. In the PR, I will try to make the higlass concept hidden from users in the APIs (like the HiGlass view IDs) and better manage the connection between Gosling track IDs and HiGlas view IDs.

src/core/track-and-view-ids.ts
sehilyi: Glad you bring this up. I think we should make a large reorganization. Let's discuss this during the next meeting and come up with a proposal together.

src/core/utils/spec-preprocess.ts
sehilyi: It is. I think these lines can be removed.

src/core/gosling.schema.ts
sehilyi: It can be removed, but I didn't since I wanted to keep the separate comments for the track `id` and the view `id` generated in `schema/gosling.schema.json`.

sehilyi:
Your understanding is correct! One thing to add (which increases the complexity) is that, in an overlaid track (like below), the `id` defined at the root level of an overlaid track is used as the HiGlass' view ID. This is an exception to limit the number of `gosling-track` instances generated (i.e., multiple tracks that are overlaid can share the same data, hence reducing the number of data requests).

```ts
tracks: [
   { ..., id: '1' }, // ← This is a track in Gosling. This becomes a HiGlass view '1'
   { ..., alignment: 'overlay', id: '2', // ← This is actually a view in Gosling, but this becomes a HiGlass view '2'
      tracks: [ 
         { ..., id: '3' }, // ← This track is included in a HiGlass view '2'
         { ..., id: '4' }, // ← This track is included in a HiGlass view '2'
   ]}
]
```

This is the motivation of https://github.com/gosling-lang/gosling.js/issues/909 where we want to make this complexity hidden from users, e.g., allow users to use track IDs (like `3` in the example) to call API functions, like `zoomTo()`, which is not possible now.

> For example, the usage of viewId here refers to HiGlass view IDs. I think we should change the variable name to be trackId to refer to Gosling track IDs?

Yes, we should rename this! I will change these names in the separate PR for https://github.com/gosling-lang/gosling.js/issues/909

> And another usage of viewId, which is referring a HiGlass view ID (I think?)

You are correct that it is a HiGlass view ID. To make it clear, should we name it `hgViewId`?

sehilyi:
I think it would also be good to explain those four concepts in our `CONBRIBUTING.md` file. I made an issue of that. https://github.com/gosling-lang/gosling.js/issues/941

sehilyi:
I will merge this to master and work on a PR.

etowahadams:
> You are correct that it is a HiGlass view ID. To make it clear, should we name it hgViewId?

That seems like a good name to me! 

> I think it would also be good to explain those four concepts in our CONBRIBUTING.md file. I made an issue of that

I agree, and I can do that

PR #943, opened 2023-07-17
feat(api): onNewTrack, onNewView

### LINKED ISSUE #926: onNewView/onNewTrack functionality
Create an event that is dispatched when a new view/track has been created to get the data and other information/e.g. dimensions, specification etc. about the track/view.

See also view API of vega: https://vega.github.io/vega/docs/api/view/

etowahadams:
Fix #926
Toward #

## Change List
 - Add `onNewTrack` API
 - Add `onNewView` API

```
// New Track
gosRef.current.api.subscribe('onNewTrack', (type, eventData) => {
    console.warn(eventData.id, eventData.shape);
});
// New View
gosRef.current.api.subscribe('onNewView', (type, eventData) => {
     console.warn(eventData.id, eventData.shape);
});
```

## Checklist

### DIFF
--- a/src/core/gosling.schema.ts
+++ b/src/core/gosling.schema.ts
@@ -248,6 +248,12 @@ interface OnNewTrackEventData {
     /** The shape of the source track */
     shape: BoundingBox | CircularTrackShape;
 }
+interface OnNewViewEventData {
+    /** Source visualization ID, i.e., `track.id` */
+    id: string;
+    /** The shape of the source track */
+    shape: BoundingBox;
+}
 
 interface PointMouseEventData extends CommonEventData {
     /** A genomic coordinate, e.g., `chr1:100,000`. */
@@ -320,6 +326,7 @@ export type _EventMap = {
     trackMouseOver: TrackApiData;
     trackClick: TrackApiData; // TODO (Jul-25-2022): with https://github.com/higlass/higlass/pull/1098, we can support circular layouts
     onNewTrack: OnNewTrackEventData;
+    onNewView: OnNewViewEventData;
 };
 
 /** Options for determining mouse events in detail, e.g., turning on specific events only */
--- a/src/core/gosling.schema.ts
+++ b/src/core/gosling.schema.ts
@@ -242,6 +242,12 @@ export interface GenomicPosition {
     chromosome: string;
     position: number;
 }
+interface OnNewTrackEventData {
+    /** Source visualization ID, i.e., `track.id` */
+    id: string;
+    /** The shape of the source track */
+    shape: BoundingBox | CircularTrackShape;
+}
 
 interface PointMouseEventData extends CommonEventData {
     /** A genomic coordinate, e.g., `chr1:100,000`. */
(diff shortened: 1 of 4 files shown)


etowahadams:
I'll add documentation once we decide that this is the final API we want to settle on

src/core/gosling.schema.ts
```diff
@@ -242,6 +242,18 @@ export interface GenomicPosition {
...
+    id: string;
+    /** The shape of the source track */
+    shape: BoundingBox | CircularTrackShape;
+}
+interface OnNewViewEventData {
+    /** Source visualization ID, i.e., `track.id` */
+    id: string;
+    /** The shape of the source track */
+    shape: BoundingBox;
+}
```
etowahadams: Currently onNewTrack and onNewView simply return the ID and the shape

sehilyi: I think we can also add `spec` as what @ThHarbig provided as an example (#926). We can reuse `TrackApiData` and `ViewApiData` for this.

@ThHarbig If you could share with us use cases, that would be helpful for determining what information we want to return.

sehilyi: Thinking more about this, I think we can just return an ID of a track or a view when it is mounted and then let users use another API function to get relevant information, such as the bounding box and spec:

```ts
const trackIdOfInterest = 'track-1';
gosRef.current.api.subscribe('onNewTrack', (type, eventData) ⇒ {
   if(eventData.id === trackIdOfInterest) {
      // 'track-1' has been mounted, so let's get the related data
      console.log(gosRef.current.api.getTrack(trackIdOfInterest); // get the bounding box and the spec
   }
});
```

I think this is cleaner as the information we return is not duplicated across API functions.

etowahadams: That makes a lot of sense to me!

etowahadams:
Another nuance to address: If a view does not have a user-defined ID in the schema, then an ID will be auto-generated. The auto-generation happens every time the spec is compiled. So for every view without a user-defined ID will trigger onNewView when the spec is re-compiled, since there is a new ID compared to the set of previous view IDs. Is this the expected behavior?

sehilyi:
It is not intended, and I suspect that this might be an issue of the editor and not the gosling.js. When I print something inside the `compiled` callback function in the `gosling-react-template` repo, which is a simpler example of using Gosling.js, I see the message printed only once.

```diff
				experimental={{ reactive: true }}
+				compiled={() => console.log('Compiled')}
			/>
```

https://github.com/gosling-lang/gosling-react-example/blob/a695b2ba532e35bc2f9e813a8aa1e385da8bb636/src/example/Simple.jsx#L30

I think as the editor becomes complex, we do not handle the React rendering cycle well, resulting in calling GoslingComponent multiple times with the same user spec.

sehilyi (commented):
Other than the event data type I commented on, this looks great to me. Since Theresa might not be available, we could merge this without her review for now and make improvements later when we can get input from her.

etowahadams:
Made the API only return the track/view ID! Going to merge now

PR #944, opened 2023-07-18
feat(api): consistent track/view IDs for JS API

### LINKED ISSUE #909: Gosling - Higlass track API improvements

sehilyi:
Fix #909
Toward #

Example Spec:
```
{
    views: [
        {
            tracks: [{ id: 's1' }, { id: 's2' }, { id: 's3' }]
        },
        {
            alignment: 'overlay',
            tracks: [{ id: 'o1' }, { id: 'o2' }, { id: 'o3' }]
        }
    ]
};
```

ID Table:
```
{
  "o1": "o1",
  "o2": "o1",
  "o3": "o1",
  "s1": "s1",
  "s2": "s2",
  "s3": "s3",
}
```

## Clarification: Boundaries of Gosling Tracks and Views 

- **Track**: A track is a visualization with a single mark type and dataset. If multiple tracks are overlaid using `alignment: 'overlay'`, the set of tracks is not a track. It is a view. For example, a gene annotation visualization that overlays multiple marks is a view instead of a track.

```ts
{ alignment: 'overlay', // ← a view with three tracks
  tracks: [
      { mark: 'rect', ... },  // ← a track
      { mark: 'rect', ... },  // ← a track
      { mark: 'point', ... }, // ← a track
  ]
}
```

- **View**: A view is a set of tracks. But, a view can contain only a single track as well. Unlike a track, there is much more flexibility regarding nesting views with a view.

```ts
{ id: 'view-root', // ← a view at the root level containing multiple views
  views: [
      { views: [...] },  // ← a view that contains multiple views
      { tracks: [...] },  // ← a view that contains multiple tracks as children
  ]
}
```

Currently, the JS API does not follow this exact concept. For example, the `trackId` we use in JS APIs is sometimes actually view IDs in the case of overlaying tracks. Other inner track IDs in each of the tracks are actually unused.

This PR will consistently use the aforementioned tracks and views concepts in the JS API.

## What are changed for users

- Users need to assign an ID to a track. If tracks are overlaid, users need to assign an ID to a track at the lowest level

<details>
<summary>History of comments</summary>

_**Update**_: I am thinking about whether it makes sense to call 'a set of tracks' that are overlaid 'a track' instead of a view. I think this may make the concept of track more intuitive to users. We also actually put the overlaid tracks into an element of `tracks`:

```ts
tracks: [
   { alignment: 'overlay', tracks: [...]} // ←
]
```

_**Update**_: The overlaid tracks can be set as an element of views as well. Might be better to just stick to the original concepts.

```ts
views: [
   { alignment: 'overlay', tracks: [...]} // ←
]
```
</details>

## Change List
 -

## Checklist


src/core/api.ts
```diff
@@ -45,8 +49,23 @@ export function createApi(
     hg: Readonly<HiGlassApi>,
     hgSpec: HiGlassSpec | undefined,
     tracksAndViews: readonly VisUnitApiData[],
-    theme: Required<CompleteThemeDeep>
+    theme: Required<CompleteThemeDeep>,
+    idTable: Readonly<IdTable>
 ): GoslingApi {
+    const idTableCopy = JSON.parse(JSON.stringify(idTable));
```
etowahadams:
```suggestion
    const idTableCopy = structuredClone(idTable);
```
consider using structuredClone() instead

src/core/compile.ts
```diff
@@ -6,9 +6,16 @@ import { getRelativeTrackInfo, type Size } from './utils/bounding-box';
...
 
 /** The callback function called everytime after the spec has been compiled */
-export type CompileCallback = (hg: HiGlassSpec, size: Size, gs: GoslingSpec, tracksAndViews: VisUnitApiData[]) => void;
+export type CompileCallback = (
+    hg: HiGlassSpec,
+    size: Size,
+    gs: GoslingSpec,
+    tracksAndViews: VisUnitApiData[],
+    idTable: IdTable
+) => void;
```
etowahadams, line 18: Thanks for adding a type to this!

src/gosling-track/gosling-track.ts
```diff
@@ -1002,11 +1010,13 @@ const factory: PluginTrackFactory<Tile, GoslingTrackOptions> = (HGC, context, op
...
-                        genomicPosition,
-                        data: capturedElements.map(d => d.value)
-                    });
+                    this.options.siblingIds.forEach(id =>
+                        publish('click', {
+                            id,
+                            genomicPosition,
+                            data: capturedElements.map(d => d.value)
+                        })
+                    );
```
etowahadams, line 1037: My understanding is that this publishes an event for each `siblingId`. Does this happen for every sibling track as well, or just the top overlaid track? i.e. when an overlaid track gets clicked on, does every overlaid track within that view get clicked on, or just the top-most overlaid track?

etowahadams (approved):
Looks good!

etowahadams:
Just to be clear:

This is allowed
```
views: [
  { alignment: 'overlay', // ← a view with three tracks
    tracks: [
        { mark: 'rect', ... },  // ← a track
        { mark: 'rect', ... },  // ← a track
        { mark: 'point', ... }, // ← a track
    ]
  }
]
```

Is below example also allowed? There is an [example](http://gosling-lang.org/docs/composition#inherit-property-in-nested-structure) in the docs like this.  This would appear to be a view within a list of tracks. 
```
tracks: [
  { alignment: 'overlay', // ← a view with three tracks
    tracks: [
        { mark: 'rect', ... },  // ← a track
        { mark: 'rect', ... },  // ← a track
        { mark: 'point', ... }, // ← a track
    ]
  }
]
```

src/gosling-track/gosling-track.ts
sehilyi: Yes, your understanding is correct. This just tries to ensure that we publish information for every track that can conceptually have an ID.

In many cases, users will be interested in a specific track and not all tracks, so I have been thinking that we can improve the API further through hierarchical pubsub topics (https://github.com/mroderick/PubSubJS#hierarchical-addressing), e.g., 

```ts
// src/core/pubsub.ts
export function publishTrackSpecific<Name extends EventName>(name: Name, trackId, string, data: _EventMap[Name]): void {
    PubSub.publish(`${name}.${trackId}`, data); // e.g., 'click.track-1'
}

export function subscribeTrackSpecific<Name extends EventName>(
    name: Name,
    trackId: string
    callback: (msg: string, data: _EventMap[Name]) => void
): void {
    PubSub.subscribe(`${name}.${trackId}`, callback);
}

// `api.subscribe('click', fn)` subscribes to all tracks
// `api.subscribeTrackSpecific('click', 'track-1', fn)` subscribes to `track-1` only
```

sehilyi:
```
Is below example also allowed? There is an [example](http://gosling-lang.org/docs/composition#inherit-property-in-nested-structure) in the docs like this. This would appear to be a view within a list of tracks.

tracks: [
  { alignment: 'overlay', // ← a view with three tracks
    tracks: [
        { mark: 'rect', ... },  // ← a track
        { mark: 'rect', ... },  // ← a track
        { mark: 'point', ... }, // ← a track
    ]
  }
]
```

Yes, it is possible, which makes the distinction between tracks and views complicated. In grammar, it is necessary because one may want to see the overlaid tracks along with other tracks. I wonder if we can make this clearer.

sehilyi:
Thanks for merging the master!

PR #946, opened 2023-07-20
feat(track): dummy-track

### LINKED ISSUE #924: Dummy/placeholder track/view
At the moment it is not (easily) possible to place other non-gosling-related views in-between gosling views. A use case for this would be placing metadata views or views with alt texts next to gosling views.

One idea I had is having a placeholder view/track that can be accessed using an ID

etowahadams:
Fix #924
Toward #

## Change List
 - Add new track type, DummyTrack 

Corresponding documentation PR https://github.com/gosling-lang/gosling-website/pull/74 
```
"tracks": [
    {
      "type": "dummy-track",
      "title": "Placeholder",
      "style": {"background": "#e6e6e6", "textFontSize": 8, "textFontWeight": "normal}
    }
],
```

Placeholder in circular layout: http://localhost:3000/?gist=etowahadams/a93bc681e03f61779f7028f4ae5a0920
Placeholder in square layout: http://localhost:3000/?gist=etowahadams/cf0389a9aa61cf98b0c183260b0a76c4

image

## Checklist
 - [x] Examples added or updated

### DIFF
--- a/src/core/compile.test.ts
+++ b/src/core/compile.test.ts
@@ -42,7 +42,7 @@ describe('gosling track.id => higlass view.uid', () => {
 });
 
 describe('Dummy track', () => {
-    it('can be compiled', () => {
+    it('compiles when layout linear', () => {
         const spec: GoslingSpec = {
             tracks: [
                 {
@@ -56,37 +56,61 @@ describe('Dummy track', () => {
                         textStrokeWidth: 0.2
                     }
                 }
-            ]
+            ],
+            layout: 'linear'
         };
         compile(
             spec,
             hgSpec => {
-                expect(hgSpec.views[0].tracks).toMatchInlineSnapshot(`
-                  {
-                    "bottom": [],
-                    "center": [],
-                    "gallery": [],
-                    "left": [],
-                    "right": [],
-                    "top": [
-                      {
+                expect(hgSpec.views[0].tracks.top).toMatchInlineSnapshot(`
+                  [
+                    {
+                      "height": 130,
+                      "options": {
+                        "background": "#000",
                         "height": 130,
-                        "options": {
-                          "background": "#000",
-                          "height": 130,
-                          "textFontSize": 10,
-                          "textStroke": "normal",
-                          "textStrokeWidth": 0.2,
-                          "title": "Placeholder",
-                          "width": 600,
-                        },
-                        "type": "dummy-track",
+                        "textFontSize": 10,
+                        "textStroke": "normal",
+                        "textStrokeWidth": 0.2,
+                        "title": "Placeholder",
                         "width": 600,
                       },
-                    ],
-                    "whole": [],
-                  }
+                      "type": "dummy-track",
+                      "width": 600,
+                    },
+                  ]
                 `);
+            },
+            [],
+            getTheme(),
+            {}
+        );
+    });
+    it('gets filtered out when layout circular', () => {
+        const spec: GoslingSpec = {
+            tracks: [
+                {
+                    type: 'dummy-track',
+                    id: 'dummy-1'
+                },
+                {
+                    id: 'track-id',
+                    data: {
+                        type: 'csv',
+                        url: ''
+                    },
+                    mark: 'rect',
+                    width: 100,
+                    height: 100
+                }
+            ],
+            layout: 'circular'
+        };
+        compile(
+            spec,
+            hgSpec => {
+                expect(hgSpec.views).toHaveLength(1);
+                expect(hgSpec.views[0].tracks).not.toBeUndefined();
             },
             [],
             getTheme(),
--- a/src/core/compile.test.ts
+++ b/src/core/compile.test.ts
@@ -40,3 +40,57 @@ describe('gosling track.id => higlass view.uid', () => {
         );
     });
 });
+
+describe('Dummy track', () => {
+    it('can be compiled', () => {
+        const spec: GoslingSpec = {
+            tracks: [
+                {
+                    type: 'dummy-track',
+                    id: 'my-dummy-track',
+                    title: 'Placeholder',
+                    style: {
+                        background: '#000',
+                        textFontSize: 10,
+                        textStroke: 'normal',
+                        textStrokeWidth: 0.2
+                    }
+                }
+            ]
+        };
+        compile(
+            spec,
+            hgSpec => {
+                expect(hgSpec.views[0].tracks).toMatchInlineSnapshot(`
+                  {
+                    "bottom": [],
+                    "center": [],
+                    "gallery": [],
+                    "left": [],
+                    "right": [],
+                    "top": [
+                      {
+                        "height": 130,
+                        "options": {
+                          "background": "#000",
+                          "height": 130,
+                          "textFontSize": 10,
+                          "textStroke": "normal",
+                          "textStrokeWidth": 0.2,
+                          "title": "Placeholder",
+                          "width": 600,
+                        },
+                        "type": "dummy-track",
+                        "width": 600,
+                      },
+                    ],
+                    "whole": [],
+                  }
+                `);
+            },
+            [],
+            getTheme(),
+            {}
+        );
+    });
+});
--- a/src/core/compile.test.ts
+++ b/src/core/compile.test.ts
@@ -74,9 +74,6 @@ describe('Dummy track', () => {
                         "options": {
                           "background": "#000",
                           "height": 130,
-                          "innerRadius": undefined,
-                          "layout": "linear",
-                          "outerRadius": undefined,
                           "textFontSize": 10,
                           "textStroke": "normal",
                           "textStrokeWidth": 0.2,
--- a/src/core/compile.test.ts
+++ b/src/core/compile.test.ts
@@ -74,6 +74,9 @@ describe('Dummy track', () => {
                         "options": {
                           "background": "#000",
                           "height": 130,
+                          "innerRadius": undefined,
+                          "layout": "linear",
+                          "outerRadius": undefined,
                           "textFontSize": 10,
                           "textStroke": "normal",
                           "textStrokeWidth": 0.2,
--- a/src/core/gosling.schema.ts
+++ b/src/core/gosling.schema.ts
@@ -213,11 +213,11 @@ export interface DataTrack extends CommonTrackDef {
  * A placeholder track. In contrast to other tracks, this track does not display any data. Instead it provides
  * empty space for third party tools to display their data on top of.
  */
-export interface DummyTrack extends Pick<CommonTrackDef, 'title' | 'width' | 'height' | 'id' | '_invalidTrack'> {
+export interface DummyTrack extends CommonTrackDef {
     /** Used to specify the dummy track */
     type: 'dummy-track';
     /** Defines how the track is styled */
-    style: DummyTrackStyle;
+    style?: DummyTrackStyle;
 }
 
 export interface DummyTrackStyle {
--- a/src/core/gosling.schema.ts
+++ b/src/core/gosling.schema.ts
@@ -213,13 +213,31 @@ export interface DataTrack extends CommonTrackDef {
  * A placeholder track. In contrast to other tracks, this track does not display any data. Instead it provides
  * empty space for third party tools to display their data on top of.
  */
-export interface DummyTrack extends CommonTrackDef {
+export interface DummyTrack
+    extends Pick<
+        CommonTrackDef,
+        | 'width'
+        | 'height'
+        | 'id'
+        | 'title'
+        | '_invalidTrack'
+        | 'overlayOnPreviousTrack'
+        | 'orientation'
+        | 'layout'
+        | 'static'
+    > {
     /** Used to specify the dummy track */
     type: 'dummy-track';
-    /** Dummy track layout must be linear */
-    layout?: 'linear';
     /** Defines how the track is styled */
     style?: DummyTrackStyle;
+    /** Dummy track layout must be linear */
+    layout?: 'linear';
+    /** Dummy track can only be static */
+    static?: true;
+    /** Dummy track cannot be zoomed in on */
+    zoomLimits?: [null, null];
+    /** No assemblies are associated with a dummy track */
+    assembly?: 'unknown';
 }
 
 export interface DummyTrackStyle {
--- a/src/core/gosling.schema.ts
+++ b/src/core/gosling.schema.ts
@@ -245,6 +245,8 @@ export interface DummyTrack
 export interface DummyTrackStyle {
     /** Background color of the track */
     background?: string;
+    /** Color of the outline of the track */
+    outline?: string;
     /** Specify the font size of the title */
     textFontSize?: number;
     /** Specify the font weight of the title. */
--- a/src/core/gosling.schema.ts
+++ b/src/core/gosling.schema.ts
@@ -229,15 +229,16 @@ export interface DummyTrack
     > {
     /** Used to specify the dummy track */
     type: 'dummy-track';
+    /** Text that gets shown on the DummyTrack */
+    title?: string;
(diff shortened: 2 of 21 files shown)


src/core/gosling.schema.ts
```diff
@@ -209,6 +209,25 @@ export interface CommonTrackDef extends CommonViewDef {
 export interface DataTrack extends CommonTrackDef {
     data: DataDeep;
 }
+/**
+ * A placeholder track. In contrast to other tracks, this track does not display any data. Instead it provides
+ * empty space for third party tools to display their data on top of.
+ */
+export interface DummyTrack extends CommonTrackDef {
```
etowahadams: ended up having to extend CommonTrackDef because too many of the other attributes were being used in other parts of the spec preprocessing

src/core/gosling.schema.ts
```diff
@@ -209,6 +209,29 @@ export interface CommonTrackDef extends CommonViewDef {
...
+    /** Background color of the track */
+    background?: string;
+    /** Specify the font size of the title */
+    textFontSize?: number;
+    /** Specify the font weight of the title. */
+    textFontWeight?: 'bold' | 'normal';
+    /** Specify the stroke color of title. */
+    textStroke?: string;
+    /** Specify the stroke width of the title. */
+    textStrokeWidth?: number;
```
etowahadams: I believe these are all of the styles we talked about. Let me know if there are any more you think would be useful!

src/core/utils/spec-preprocess.ts
```diff
@@ -258,6 +256,10 @@ export function traverseToFixSpecDownstream(spec: GoslingSpec | SingleView, pare
             // Override styles
             track.style = getStyleOverridden(spec.style, track.style);
             if (IsOverlaidTrack(track)) {
+                // Remove the dummy tracks from an overlay track
+                track.overlay = track.overlay.filter(overlayTrack => {
+                    return !('type' in overlayTrack && overlayTrack.type == 'dummy-track');
+                });
```
etowahadams, line 270: We don't allow dummy tracks to be overlaid on other tracks

src/missing-types.d.ts
```diff
@@ -565,6 +566,15 @@ declare module '@higlass/tracks' {
...
 
+    export class SVGTrack<Options> extends Track<Options> {
+        /* Properties */
+        gMain: d3Selection.Selection<d3Selection.Datum, d3Selection.PElement, d3Selection.PDatum>;
+        clipUid: string;
+        clipRect: d3Selection.Selection<d3Selection.Datum, d3Selection.PElement, d3Selection.PDatum>;
+        /* Constructor */
+        constructor(context: Context<unknown, Options>, options: Options);
+    }
+
```
etowahadams, line 577: Corresponds to [SVGTrack](https://github.com/higlass/higlass/blob/b2ee5940c519982dc53685153ff863d64443d0bb/app/scripts/SVGTrack.js#L6) in HiGlass

src/core/compile.test.ts
```diff
@@ -40,3 +40,57 @@ describe('gosling track.id => higlass view.uid', () => {
...
+                      },
+                    ],
+                    "whole": [],
+                  }
+                `);
+            },
+            [],
+            getTheme(),
+            {}
+        );
```
sehilyi: Thank you for adding a test!

src/core/gosling.schema.ts
sehilyi: We would need to clarify in the documentation that properties are actually unused.

src/dummy-track/dummy-track.ts
```diff
@@ -0,0 +1,57 @@
...
+                .append('text')
+                .attr('x', 2)
+                .attr('y', options.textFontSize)
+                .attr('color', 'red')
+                .style('font-size', `${options.textFontSize}px`)
+                .style('font-weight', options.textFontWeight)
+                .style('stroke', options.textStroke)
+                .style('stroke-width', options.textStrokeWidth)
+                .text(options.title);
+        }
```
sehilyi: Could we align the placeholder text to the center to make it more obvious? Something like,

```suggestion
                .attr('x', options.width / 2)
                .attr('y', (options.height + options.textFontSize) / 2)
                .attr('color', 'red')
                .style('text-anchor', 'middle')
                .style('font-size', `${options.textFontSize}px`)
                .style('font-weight', options.textFontWeight)
                .style('stroke', options.textStroke)
                .style('stroke-width', options.textStrokeWidth)
                .text(options.title);
        }
```

src/dummy-track/dummy-track.ts
```diff
@@ -0,0 +1,57 @@
...
+                .attr('x', 0)
+                .attr('y', 0)
+                .attr('width', options.width)
+                .attr('height', options.height);
+            // Title text
+            this.gMain
+                .append('text')
+                .attr('x', 2)
+                .attr('y', options.textFontSize)
+                .attr('color', 'red')
```
sehilyi: Color seems to be hard coded?

sehilyi (commented):
One thing to fix is that the dummy track does not work when a view with multiple tracks is circular:

Screenshot 2023-07-24 at 12 05 55

```ts
{  
  "layout": "circular",
  "views": [
    {
      "tracks": [
        {
          "id": "track-1",
          "data": {
            "type": "multivec",
            "row": "sample",
            "column": "position",
            "value": "peak",
            "categories": ["sample 1", "sample 2", "sample 3", "sample 4"],
            "binSize": 4
          },
          "mark": "rect",
          "x": {"field": "start", "type": "genomic", "axis": "top"},
          "row": {"field": "sample", "type": "nominal", "legend": true},
          "color": {"field": "peak", "type": "quantitative", "legend": true},
          "tooltip": [
            {"field": "start", "type": "genomic", "alt": "Start Position"},
            {"field": "end", "type": "genomic", "alt": "End Position"},
            {
              "field": "peak",
              "type": "quantitative",
              "alt": "Value",
              "format": ".2"
            },
            {"field": "sample", "type": "nominal", "alt": "Sample"}
          ],
          "width": 600,
          "height": 130
        },
          {
          "id": "du",
      "type": "dummy-track",
      "title": "Placeholder",
      "style": {"background": "#e6e6e6", "textFontSize": 18, "textFontWeight": "normal" },
      "width": 500,
      "height": 60
      }
      ]
    }
  ]
}
```

Ideally, we can support circular dummy tracks, but for most cases, we just need the linear dummy track, i.e., showing rectangles instead of arcs. So, for the quick fix, we could just remove dummy tracks in the spec processing when the layout is circular and clarify this in the documentation.

src/dummy-track/dummy-track.ts
etowahadams: Removed!

etowahadams:
I added support for circular dummy tracks! The example you gave now looks like this 

image

For reasons still unclear to me, the circular dummy tracks are break interactivity of the other tracks in the same view though---when there's a dummy track in a circular layout view, you can't zoom into the other tracks. I tried deleting all  SVG nodes in the DOM related to the circular dummy track, but it didn't restore interactivity. So it could be HiGlass schema related? Let me know if you have any ideas as to what I might look into

sehilyi:
It's great to see the circular dummy tracks. Let me take a look at the interactivity issue.

(1) One thing to mention is that the arrangement of views in circular layouts is quite complex where you could use half circles or 1/4 circles for a view by using a `serial` arrangement:

Screenshot 2023-07-25 at 10 31 06

```ts
{
  "layout": "circular",
  "arrangement": "serial",
  "views": [
    {...}, {...}
  ]
}
```

<details>

<summary>Full Spec</summary>

```ts
{
  "layout": "circular",
  "arrangement": "serial",
  "centerRadius": 0.8,
  "xDomain": {"chromosome": "chr1", "interval": [1, 3000500]},
  "views": [
    {
      "tracks": [
        {
          "data": {
            "type": "multivec",
            "row": "sample",
            "column": "position",
            "value": "peak",
            "categories": ["sample 1", "sample 2", "sample 3", "sample 4"],
            "binSize": 4
          },
          "mark": "rect",
          "x": {"field": "start", "type": "genomic", "axis": "top"},
          "row": {"field": "sample", "type": "nominal", "legend": true},
          "color": {"field": "peak", "type": "quantitative", "legend": true},
          "tooltip": [
            {"field": "start", "type": "genomic", "alt": "Start Position"},
            {"field": "end", "type": "genomic", "alt": "End Position"},
            {
              "field": "peak",
              "type": "quantitative",
              "alt": "Value",
              "format": ".2"
            },
            {"field": "sample", "type": "nominal", "alt": "Sample"}
          ],
          "width": 300,
          "height": 130
        }
      ]
    },
    {
      "tracks": [
        {
          "data": {
            "type": "multivec",
            "row": "sample",
            "column": "position",
            "value": "peak",
            "categories": ["sample 1", "sample 2", "sample 3", "sample 4"]
          },
          "mark": "bar",
          "x": {"field": "position", "type": "genomic", "axis": "top"},
          "y": {"field": "peak", "type": "quantitative"},
          "color": {"field": "sample", "type": "nominal", "legend": true},
          "tooltip": [
            {"field": "start", "type": "genomic", "alt": "Start Position"},
            {"field": "end", "type": "genomic", "alt": "End Position"},
            {
              "field": "peak",
              "type": "quantitative",
              "alt": "Value",
              "format": ".2"
            },
            {"field": "sample", "type": "nominal", "alt": "Sample"}
          ],
          "width": 300,
          "height": 130
        }
      ]
    }
  ]
}
```

</details>

This means, to fully support the circular visualization, dummy tracks will need to be able to show as arcs (e.g., having `startAngle` and `endAngle` defined internally). I used `d3arc` in `brush-track` to draw arcs, which you can refer to. 

https://github.com/gosling-lang/gosling.js/blob/ef5f58f43ee33ca2de068a547f21ed1767426437/src/gosling-brush/brush-track.ts#L46

But, I encourage you not to spend too much time implementing this. If you think this is not straightforward to implement, we can implement this in the future. The main use case of a dummy track (at least in the context of meta-vis and accessibility) is when using linear layouts.

(2) Thinking of use cases, I think it would be good to support `onNewTrack`. The use case would be like,

   1. Show genome-mapped data visualization using`<GoslingComponent/>` where spec defines a dummy track
   2. In the callback function of `onNewTrack`, retrieve the bounding box of the placeholder track and then display a metadata visualization (or accessibility interface) inside the bounding box of the dummy track.

sehilyi:
> For reasons still unclear to me, the circular dummy tracks are break interactivity of the other tracks in the same view though

It looks like this is dependent on HiGlass and would require updates to HiGlass. In the circular view, all tracks in Gosling are added to the same HiGlass **view**, and my assumption is that when an SVG track is added at last to the HiGlass view, it blocks mouse interactions. Maybe this is for another PR.

etowahadams:
I see that does make it more complex. How's this for a plan? 

1. Put off circular dummy tracks for a future PR. I'll write up a feature request for it with all the details/complexities. 
2. Add support for `onNewTrack` for the dummy track

sehilyi:
@etowahadams That sounds like a plan to me!

src/core/gosling.schema.ts
```diff
@@ -209,6 +209,50 @@ export interface CommonTrackDef extends CommonViewDef {
...
+    style?: DummyTrackStyle;
+    /** Dummy track layout must be linear */
+    layout?: 'linear';
+    /** Dummy track can only be static */
+    static?: true;
+    /** Dummy track cannot be zoomed in on */
+    zoomLimits?: [null, null];
+    /** No assemblies are associated with a dummy track */
+    assembly?: 'unknown';
+}
```
etowahadams, line 242: Rather than just extending `CommonTrackDef`, I tried making making the properties used explicit so users won't be confused. Some of them like `layout`, `static`, and `zoomLimits` I kept in there they're heavily used in `spec-preprocess.ts` and I would have to significantly refactor it to not have them. Let me know if you think there would be a better way to do this.

src/core/gosling.schema.ts
```diff
@@ -209,6 +209,51 @@ export interface CommonTrackDef extends CommonViewDef {
...
+    /** Defines how the track is styled */
+    style?: DummyTrackStyle;
+    /** Only linear layout are supported at this time */
+    layout?: 'linear';
+    static?: true;
+    /** Unused property for DummyTrack */
+    zoomLimits?: [null, null];
+    /** No assemblies can be associated with a dummy track */
+    assembly?: 'unknown';
+}
```
etowahadams, line 243: Rather than just extending CommonTrackDef, I tried making making the properties used explicit so users won't be confused. Some of them like layout, static, and zoomLimits I kept in there because they're heavily used in spec-preprocess.ts and I would have to significantly refactor it to not have them. Let me know if you think there would be a better way to do this.

etowahadams:
Added an example, removed the circular track support I started, added `onNewTrack` API support, added documentation (https://github.com/gosling-lang/gosling-website/pull/74) with a link to the example

src/missing-types.d.ts
```diff
@@ -565,6 +566,15 @@ declare module '@higlass/tracks' {
...
     }
 
+    export class SVGTrack<Options> extends Track<Options> {
+        /* Properties */
+        gMain: d3Selection.Selection<d3Selection.Datum, d3Selection.PElement, d3Selection.PDatum>;
+        clipUid: string;
+        clipRect: d3Selection.Selection<d3Selection.Datum, d3Selection.PElement, d3Selection.PDatum>;
+        /* Constructor */
+        constructor(context: Context<unknown, Options>, options: Options);
+    }
```
sehilyi, line 576: Thank you for adding this!

tsconfig.json
```diff
@@ -30,6 +30,7 @@
       "@gosling-track": ["./src/gosling-track/index.ts"],
       "@gosling-genomic-axis": ["./src/gosling-genomic-axis/index.ts"],
       "@gosling-brush": ["./src/gosling-brush/index.ts"],
+      "@gosling-dummy": ["./src/dummy-track/index.ts"],
```
sehilyi: Let's change the name. We will eventually use the `@gosling-lang` NPM organization to publish sub-packages, so the convention will be `@gosling-lang/package-name` when we support monorepo.

```suggestion
      "@gosling-lang/dummy-track": ["./src/dummy-track/index.ts"],
```

vite.config.js
```diff
@@ -61,6 +61,7 @@ const alias = {
     "@gosling-mouse-event": path.resolve(__dirname, "./src/gosling-mouse-event/index.ts"),
     "@gosling-genomic-axis": path.resolve(__dirname, "./src/gosling-genomic-axis/index.ts"),
     "@gosling-brush": path.resolve(__dirname, "./src/gosling-brush/index.ts"),
+    "@gosling-dummy": path.resolve(__dirname, "./src/dummy-track/index.ts"),
```
sehilyi:
```suggestion
    "@gosling–lang/dummy-track": path.resolve(__dirname, "./src/dummy-track/index.ts"),
```

src/core/utils/spec-preprocess.ts
sehilyi: It looks like a dummy track in a circular view hide a regular track, e.g., 

```ts
{
      "layout":  "circular",
      "tracks": [
        {
          "type": "dummy-track",
          "title": "here",
          "style": { "background": "grey" },
          "width": 100
        },
        { ... } // ← regular track that is hidden in a rendered visualization
   ]
}
```

I think it would be good to prevent users from doing unsupported behavior. I think we can simply remove the dummy track in a circular view, similar to what you did for overlaid tracks, somewhere in this function.

etowahadams: Done! Dummy track in circular layout now removed

etowahadams:
- Added an additional `outline` style
- Dummy track now removed from circular layouts 
I think this should be everything now!

sehilyi (approved):

PR #957, opened 2023-08-02
fix(data-fetcher): gmod/gff stream issue

### LINKED ISSUE #956: Broken GFF data because of vite issue
Another story of "works locally, but not when deployed"

https://gosling.js.org/?example=doc_gff

I noticed the GFF data fetcher example wasn't working in the deployed version but it was working locally. There's this error when `yarn build-editor` is run 

```
node_modules/lodash-es/_asciiToArray.js'emitter' is imported by emitter?commonjs-external, but could not be resolved – treating it as an external dependency

No name was provided for external module 'emitter' in output.globals – guessing 'require$$0'
```

This results in the following error in the editor for a track with gff data:   
image

Appears to be the same issue discussed here https://github.com/vitejs/vite/issues/6586

Things I'm going to look into 
- Try bumping vite version
- Create an [alias](https://stackoverflow.com/questions/73650567/vite-build-fails-emitter-is-imported-by-emittercommonjs-external-but-could) for the emitter 

cc: @sehilyi

etowahadams:
Fix #956 
Toward #

## Change List
- Alias for `stream-browserify`

gmod/gff uses `stream-browserify`, but defines it using the `browser` [field in package.json](https://github.com/GMOD/gff-js/blob/6373b3afb432fc4524f99f69b1116d70934f4a21/package.json#L9C1-L11). Apparently, [vite has trouble](https://github.com/vitejs/vite/issues/7576#issue-1190601191) using packages that are defined that way. 

HiGlass has `stream` as a dependency, so vite uses `stream` instead of `stream-browserify` for gmod/gff. This caused the gff fetcher to not work in gosling. 

I just added an alias for `stream-browserify` so gmod/gff uses the correct stream. This doesn't appear to impact HiGlass. 

before: 
image

after: 
image 

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,6 +81,7 @@
         "quick-lru": "^6.1.1",
         "rbush": "^3.0.1",
         "react-grid-layout": "^1.2.5",
+        "stream-browserify": "^3.0.0",
         "threads": "^1.6.4",
         "uuid": "^8.3.2"
     },
--- a/vite.config.js
+++ b/vite.config.js
@@ -64,7 +64,8 @@ const alias = {
     "@gosling-lang/dummy-track": path.resolve(__dirname, "./src/dummy-track/index.ts"),
     "@data-fetchers": path.resolve(__dirname, "./src/data-fetchers/index.ts"),
     zlib: path.resolve(__dirname, './src/alias/zlib.ts'),
-    uuid: path.resolve(__dirname, './node_modules/uuid/dist/esm-browser/index.js')
+    uuid: path.resolve(__dirname, './node_modules/uuid/dist/esm-browser/index.js'),
+    stream: path.resolve(__dirname, './node_modules/stream-browserify')
 };
 
 const skipExt = new Set(['@gmod/bbi', 'uuid']);


etowahadams:
I think the lesson (for both here and with https://github.com/gosling-lang/gosling.js/pull/938) is that if we add new packages, then we should not expect them to work in the built version, and we should always review the warnings that are produced by `yarn build` and `yarn build-editor`, which aren't caught in CI. I wonder if we can add a test which checks for new warnings when building. 

Should we release a patch too?

sehilyi (approved):
Thank you for the fix. I will make a patch after merging this.

Regarding the test, I wonder if we could set up tests that compare the actual images of visualizations that the previous and updated `<GoslingComponent/>` displayed. It looks like [puppeteer](https://pptr.dev/), which is a library that helps to capture screenshots, has the ability to add `<script>` using a local file (https://pptr.dev/api/puppeteer.page.addscripttag), so I imagine that we can bundle the gosling.js and then use in puppeteer to create Gosling image files based on a updated Gosling.js.

An example of puppeteer with Gosling: https://github.com/hms-dbmi/chromoscope/blob/master/src/script/gosling-screenshot.js

etowahadams:
I agree that would be good. I'll make an issue for it!

PR #962, opened 2023-08-17
test(core): add visual regression testing 

### LINKED ISSUE #958: Automated testing for rendered visualization differences
**Background**: 
Sometimes, Gosling code changes will break a visualization, but will not cause any tests to fail.  For example, https://github.com/gosling-lang/gosling.js/pull/938 https://github.com/gosling-lang/gosling.js/pull/957 were patches that had to be made after a release because automatic tests did not alert us of anything wrong. 

**Potential solution**:
@sehilyi: To avoid future instances of this, we could set up tests that compare the actual images of visualizations that the previous and updated <GoslingComponent/> displayed. It looks like [puppeteer](https://pptr.dev/), which is a library that helps to capture screenshots, has the ability to add <script> using a local file (https://pptr.dev/api/puppeteer.page.addscripttag), so I imagine that we can bundle the gosling.js and then use in puppeteer to create Gosling image files based on a updated Gosling.js.

An example of puppeteer with Gosling: https://github.com/hms-dbmi/chromoscope/blob/master/src/script/gosling-screenshot.js

etowahadams:
Fix #958
Toward #

Adds a `yarn screenshot` command. This command will use `dist/gosling.js` in a headless browser (using puppeteer) and take a screenshot of the visualization that is included in the editor. It will save these screenshots in `img/visual-regression/new-screenshots`. If there is a difference compared to the screenshots in `img/visual-regression/screenshots`it will show the differences in an image in `img/visual-regression/diffs`. 

Interestingly, the Gosling genomic axis is non-deterministic for many specs. Parts of the axis will look different between renders. This makes automated visual regression testing unsuitable. I think this best used as a check when testing locally to make sure that many visualizations are not impacted.  

## Checklist


vite.config.js
```diff
@@ -121,9 +128,27 @@ const testing = defineConfig({
...
+        globals: true,
+        environment: 'jsdom',
+        threads: false,
+        environmentOptions: {
+            jsdom: {
+                resources: 'usable'
+            }
+        }
+    }
+});
```
etowahadams, line 146: New config to be used only when `yarn screenshot` is called so that only tests in `img/visual-regression` will be run

etowahadams:
Ready for review. I finally felt inspired to try new things after I was unable to get workers to work in puppeteer. Turns out workers don't work if there is no URL. The trick is to navigate to a page first, then to change the content of the page. I spent way too long figuring this out 😭 

I decided not to make this visual regression test automated for now because many gosling visualizations are non deterministic. In particular, the axis seems to have some non-deterministic behavior when zoomed out. Here is an example of the diff between two renderings of the doc_point spec. In red are the regions which do not match between old and new renders. About 20 of the specs had this behavior. This might be a bug that we should fix.  

Still, I think this should be a useful tool to understand the extend of visual changes. 

image

img/visual-regression/visual-regression.test.ts
```diff
@@ -0,0 +1,116 @@
...
+	<script src="${baseUrl}/pixi.js@${pixijsVersion}/dist/browser/pixi.min.js"></script>
+	<script src="${baseUrl}/higlass@${higlassVersion}/dist/hglib.js"></script>
+    <script type="text/javascript">${gosling}</script>
+<body>
+	<div id="vis"></div>
+	<script>
+		gosling.embed(document.getElementById("vis"), JSON.parse(\`${spec}\`))
+	</script>
+</body>
+</html>`;
```
etowahadams, line 36: This will have to change when we update Gosling to only have a ESM build

etowahadams:
Any updates on this? I think this will be quite useful when implementing performance related improvements.

sehilyi:
Thank you for the reminder! And, sorry, I lost track of this.

Regarding the change of the axis, what do you think of making the visibility of ticks deterministic? We can remove [the use of random values](https://github.com/gosling-lang/gosling.js/blob/f17e69f8ea9d83e9c574e864b7dfca94a49fa328/src/tracks/gosling-genomic-axis/axis-track.ts#L101) for determining the visibility, and it might be as simple as removing the following line:

https://github.com/gosling-lang/gosling.js/blob/f17e69f8ea9d83e9c574e864b7dfca94a49fa328/src/tracks/gosling-genomic-axis/axis-track.ts#L672

CONTRIBUTING.md
```diff
@@ -61,6 +61,14 @@ yarn build-editor
 yarn preview
 ```
 
+## Visual Regression
+
+To test if your changes make a difference visually across a large number of specifications, you can use can use the `yarn screenshot` command.
```
sehilyi:
```suggestion
To test if your changes make a difference visually across a large number of specifications, you can use the `yarn screenshot` command. 
```

CONTRIBUTING.md
```diff
@@ -61,6 +61,14 @@ yarn build-editor
 yarn preview
 ```
 
+## Visual Regression
+
+To test if your changes make a difference visually across a large number of specifications, you can use can use the `yarn screenshot` command. 
+This command will use `dist/gosling.js` in a headless browser (using puppeteer) and take a screenshot of the visualization that is included in the editor.
```
sehilyi, line 72: You could briefly mention that one needs to run `yarn build` before `yarn screenshot` to test the updated Gosling.js.

CONTRIBUTING.md
```diff
@@ -61,6 +61,14 @@ yarn build-editor
 yarn preview
 ```
 
+## Visual Regression
+
+To test if your changes make a difference visually across a large number of specifications, you can use can use the `yarn screenshot` command. 
+This command will use `dist/gosling.js` in a headless browser (using puppeteer) and take a screenshot of the visualization that is included in the editor. 
+It will save these screenshots in `img/visual-regression/new-screenshots`. If there is a difference compared to the screenshots in `img/visual-regression/screenshots`
```
sehilyi: So, for now, would the usage pattern be like (occasionally) running the `yarn build` and `yarn screenshot` and then moving all files under `img/visual-regression/new-screenshots` to `img/visual-regression/screenshots`?

sehilyi (commented):
This looks like a very useful feature! Thank you for making these chnages.

One thing I notice is when I run `yarn`, `yarn build`, and `yarn screenshot`, I am getting the following error complaining that the `new-screenshots/*` are not found. Just in case it is related to wait times, I tested with larger values for related parameters, e.g., `idelTime`, `delay()`, but it shows the same issue. There must be something that I am missing. Do you have any clue about this issue?

Screenshot 2023-10-17 at 15 00 47

sehilyi:
**Update**: It looks like I needed to create empty `new-screenshots` and `diffs` folders first. I thinks this can be added to the `yarn screenshot` script.

CONTRIBUTING.md
etowahadams: Will add!

CONTRIBUTING.md
etowahadams: That's how I currently see it being used in its current state, yes. I can add this as usage pattern in the documentation. I can also add a command which moves the screenshots from `img/visual-regression/new-screenshots` to `img/visual-regression/screenshots`.

PR #965, opened 2023-08-25
feat(data-fetcher,core): Add support for getting data from URLs that require certain request initializations 

### LINKED ISSUE #963: Support for using data from authenticated URLs
We would like Gosling to be able to use data from private URLs. To do this, we need add a new option to the Gosling schema. 

**Background and motivation**
Currently, users provide Gosling with a URL to their data. Data fetchers get the data from the URL. However, if the URL points to a server which requires certain request options, for example a HTTP Authorization request header, then Gosling will not be able to retrieve that data. 

One use case is retrieving files private files from Google Drive. To retrieve private files from Google drive, web applications must obtain an OAuth 2.0 authorization token. The web application can then [send that authorization token](https://developers.google.com/identity/protocols/oauth2#4.-send-the-access-token-to-an-api.) to a Google API to do things which are permitted by the authorization token's scope. 

**Proposed Gosling Schema changes**
I propose adding a an optional property called `fetchOptions` to each data type that takes a URL (csv, gff, vcf, etc. but not json). The `fetchOptions` property will take an object which has the [RequestInit](https://fetch.spec.whatwg.org/#requestinit) type.  Alternatively, it could take a more limited version of this type. 

```
interface RequestInit {
    /** A BodyInit object or null to set request's body. */
    body?: BodyInit | null;
    /** A string indicating how the request will interact with the browser's cache to set request's cache. */
    cache?: RequestCache;
    /** A string indicating whether credentials will be sent with the request always, never, or only when sent to a same-origin URL. Sets request's credentials. */
    credentials?: RequestCredentials;
    /** A Headers object, an object literal, or an array of two-item arrays to set request's headers. */
    headers?: HeadersInit;
    /** A cryptographic hash of the resource to be fetched by request. Sets request's integrity. */
    integrity?: string;
    /** A boolean to set request's keepalive. */
    keepalive?: boolean;
    /** A string to set request's method. */
    method?: string;
    /** A string to indicate whether the request will use CORS, or will be restricted to same-origin URLs. Sets request's mode. */
    mode?: RequestMode;
    /** A string indicating whether request follows redirects, results in an error upon encountering a redirect, or returns the redirect (in an opaque fashion). Sets request's redirect. */
    redirect?: RequestRedirect;
    /** A string whose value is a same-origin URL, "about:client", or the empty string, to set request's referrer. */
    referrer?: string;
    /** A referrer policy to set request's referrerPolicy. */
    referrerPolicy?: ReferrerPolicy;
    /** An AbortSignal to set request's signal. */
    signal?: AbortSignal | null;
    /** Can only be null. Used to disassociate request from any Window. */
    window?: null;
}
``` 
**Example usage**
Getting a file with an authorization header 
```
data: {
    url: 'https://<google drive url>.csv',
    type: 'csv',
    fetchOptions: {'headers': {'Authorization': 'Bearer [YOUR_ACCESS_TOKEN]'}}
    chromosomeField: 'Chromosome',
    genomicFields: ['chromStart', 'chromEnd']
},
```

cc: @sehilyi @manzt

etowahadams:
Fix #963
Toward #

## Change List
- Adds a `urlToFetchOptions` parameter to the compiler so that fetch options specific to a url can be passed to datafetchers. 
- Now `RemoteFile` is used instead of `fetch`
- DataFetchers receive a `urlFetchOptions` and/or `indexUrlFetchOptions` to call fetch with 

Will add documentation once we settle on this specific API 

Example:
```javascript
const urlToFetchOptions = { 'https://my-csv-url.com': { headers: { Authentication: 'Bearer 1234' } } };

const spec: GoslingSpec = {
            tracks: [
                {
                    id: 'track-id',
                    data: {
                        type: 'csv',
                        url: 'https://my-csv-url.com'
                    },
                    mark: 'rect',
                    width: 100,
                    height: 100
                }
            ]
        };

// then we compile it with the urlToFetchOptions 
compile(
            spec,
            () => {},
            [],
            getTheme(),
            {},
            urlToFetchOptions
        );

// then the following gets passed to the datafetcher 
{
  "assembly": "hg38",
  "indexUrlFetchOptions": {},
  "type": "csv",
  "url": "https://my-csv-url.com",
  "urlFetchOptions": {
    "headers": {
      "Authentication": "Bearer 1234",
    },
  },
  "x": undefined,
  "x1": undefined,
  "x1e": undefined,
  "xe": undefined,
}
```

## Checklist


src/data-fetchers/csv/csv-data-fetcher.test.ts
etowahadams, line 1: These tests no longer work because because of a vitest error. 
Before we were using fetch in the csv datafetcher which can be patched in the following way when running tests:

https://github.com/gosling-lang/gosling.js/blob/master/src/data-fetchers/csv/csv-data-fetcher.test.ts#L3-L5

However, if another module calls fetch, then this method for patching fetch doesn't work. In our case, we are now using `RemoteFile` which calls fetch internally.

I would prefer not to get rid of these tests, but the only thing I can think of is to pass in a `fetch` function into RemoteFile if it is in a node environment. This is not the most elegant solution, so if others have thoughts or have encountered a similar issue I'd love to hear.

src/compiler/gosling-to-higlass.ts
```diff
@@ -152,10 +154,19 @@ export function goslingToHiGlass(
...
+                ('url' in firstResolvedSpec.data && urlToFetchOptions?.[firstResolvedSpec.data.url]) || {};
+            const indexUrlFetchOptions =
+                ('indexUrl' in firstResolvedSpec.data && urlToFetchOptions?.[firstResolvedSpec.data.indexUrl]) || {};
+
+            // This object will be passed to the data fetchers
             hgTrack.data = {
                 ...firstResolvedSpec.data,
                 ...xFields,
+                urlFetchOptions,
+                indexUrlFetchOptions,
```
etowahadams, line 169: This is where the `urlToFetchOptions` map is used to pass the fetch options to the data fetchers

src/higlass-schema/higlass.schema.ts
```diff
@@ -100,21 +100,38 @@ export interface EnumTrack {
...
+    body?: any;
+    method?: string;
+    mode?: string;
+    credentials?: string;
+    cache?: string;
+    referrer?: string;
+    referrerPolicy?: string;
+    [key: string]: any;
+}
+
```
etowahadams: Should I add the [whole](https://fetch.spec.whatwg.org/#requestinit) RequestInit type here? Or just a subset?

src/data-fetchers/bed/bed-worker.ts
```diff
@@ -55,12 +59,20 @@ export class BedFile {
...
+        indexUrl: string,
+        uid: string,
+        urlFetchOptions?: RequestInit,
+        indexUrlFetchOptions?: RequestInit
+    ) {
         const tbi = new TabixIndexedFile({
-            filehandle: new RemoteFile(url),
-            tbiFilehandle: new RemoteFile(indexUrl)
+            filehandle: new RemoteFile(url, { overrides: urlFetchOptions }),
+            tbiFilehandle: new RemoteFile(indexUrl, { overrides: indexUrlFetchOptions })
```
etowahadams, line 75: All the datafetchers pass the `urlFetchOptions`/`indexUrlFetchOptions` into `RemoteFile` as an [override](https://github.com/GMOD/generic-filehandle/blob/master/src/remoteFile.ts#L85-L98)

etowahadams:
I forgot to follow up this with this PR. Just added some comments and questions. Ready for review.

src/higlass-schema/higlass.schema.ts
sehilyi: I think the subset that we actually use should be fine

src/data-fetchers/csv/csv-data-fetcher.test.ts
```diff
@@ -1,31 +1,40 @@
...
+ */
+vi.mock('generic-filehandle', () => {
+    const str = 'c1,s1,e1,c2,s2,e2\n1,486,76975,15,100263879,100338121\n1,342608,393885,15,100218755,100268630';
+    const RemoteFile = vi.fn();
+    RemoteFile.prototype.readFile = vi.fn().mockResolvedValue(str);
+    RemoteFile.prototype.fetch = vi.fn();
+    return {
+        RemoteFile
+    };
+});
```
sehilyi, line 17: This is great!

src/data-fetchers/csv/csv-data-fetcher.test.ts
```diff
@@ -42,36 +51,53 @@ describe('CSV data fetcher', () => {
         new Promise<void>(resolve => {
             fetcher.fetchTilesDebounced(
                 (loadedTile: LoadedTiles) => {
-                    expect(Object.keys(loadedTile['0.0'].tabularData[0])).toMatchInlineSnapshot(`
+                    expect(loadedTile['0.0'].tabularData).toMatchInlineSnapshot(`
```
sehilyi, line 54: Thanks for updating this to test more broadly

src/data-fetchers/csv/csv-data-fetcher.ts
```diff
@@ -72,14 +77,14 @@ export class CsvDataFetcherClass {
...
+        const { chromosomeField, genomicFields, headerNames, longToWideId, genomicFieldsToConvert } = this.dataConfig;
 
         const separator = this.dataConfig.separator ?? ',';
 
         try {
-            const response = await fetch(url);
-            const text = await (response.ok ? response.text() : Promise.reject(response.status));
+            const buffer = await this.#file.readFile();
+            const text = buffer.toString();
+            console.warn(text);
```
sehilyi:
```suggestion
```

src/core/gosling-embed.ts
```diff
@@ -20,6 +20,7 @@ export type GoslingEmbedOptions = Omit<HiGlassComponentWrapperProps['options'],
     id?: string;
     className?: string;
     theme?: Theme;
+    urlToFetchOptions?: UrlToFetchOptions;
```
sehilyi, line 24: Are you planning to update `GoslingCompProps` of `GoslingComponent` by adding `urlToFetchOptions` as well?

etowahadams: Oops I forgot about that! Yes just added now.

sehilyi:
@etowahadams Let me know when we are read for a minor update!

etowahadams:
I just tested with 
```
yarn build
yarn build-editor
yarn preview
```
I noticed that the xDomain parameter seems to be ignored when switching between specs in the editor. 
Open editor - this one looks as expected! 
image

Switch to band encoding - the xDomain parameter is ignored
image

What it is supposed to look like: 
image

I think that https://github.com/gosling-lang/gosling.js/pull/971 could be why? I just tested https://gosling.js.org and it has the same behavior. So perhaps we should fix this bug and then do a minor release? Otherwise everything else looks fine.

sehilyi:
Thank you for noticing this. Let me try to fix it first.

etowahadams:
Is this ready to merge now then?

sehilyi:
It looks like this was already merged. I noticed this after making a patch, so I ended up bumping a minor version as well https://github.com/gosling-lang/gosling.js/releases/tag/v0.11.0

etowahadams:
Oh whoops I didn't realize I had already merged!

PR #974, opened 2023-09-26
fix(core): preserve the zoom status only when needed

### LINKED ISSUE #973: Zooms are preserved between Editor examples
I just tested with 
```
yarn build
yarn build-editor
yarn preview
```
I noticed that the xDomain parameter seems to be ignored when switching between specs in the editor. 
Open editor - this one looks as expected! 
image

Switch to band encoding - the xDomain parameter is ignored
image

What it is supposed to look like: 
image

I think that https://github.com/gosling-lang/gosling.js/pull/971 could be why? I just tested https://gosling.js.org and it has the same behavior. So perhaps we should fix this bug and then do a minor release? Otherwise everything else looks fine.

_Originally posted by @etowahadams in https://github.com/gosling-lang/gosling.js/issues/965#issuecomment-1735592865_

sehilyi:
Fix #973

## Change List
 - Do not preserve zoom status across examples. This is fixed by adding an additional if statement. Also, added tests for this.

## Checklist


src/compiler/spec-preprocess.ts
```diff
@@ -565,3 +566,27 @@ export function overrideDataTemplates(spec: GoslingSpec) {
...
+        const viewUid = view.uid!;
+        const isNewView = !prevSpec.views.find(v => v.uid === viewUid);
+        if (isNewView) {
+            // if this view is linked with another view, we need to preverse the current zooming status of this view from the linked view
+            // Otherwise, all the views that is linked with this view will be reset to the original zooming position
+            const { locksByViewUid } = newSpec.zoomLocks;
+            const lockUid = locksByViewUid[viewUid];
+            const linkedViewUid = Object.entries(locksByViewUid).find(([_, uid]) => _ && uid === lockUid)?.[0];
+            // Only if the linked view existed in the previous spec, we copy the zooming status
+            const linkedViewExistedPrev = !!prevSpec.views.find(v => v.uid === linkedViewUid);
```
sehilyi: The main code I added to fix the issue. This basically checks whether if this view is linked with a view that existed previously.

src/core/gosling-component.tsx
```diff
@@ -13,6 +13,7 @@ import { isEqual } from 'lodash-es';
 import * as uuid from 'uuid';
 import { publish } from '../api/pubsub';
 import type { IdTable } from '../api/track-and-view-ids';
+import { preverseZoomStatus } from '../compiler/spec-preprocess';
```
sehilyi: Moved under \compiler – a better place than this?

etowahadams: Given that this function is only used in `gosling-component.tsx` I think it should be kept in /core? I see you logic though, that it is making manipulations on the spec.

etowahadams (approved):
Just tested with the built editor by adding and removing tracks and switching between specs. Looks good to me! Thanks for taking care of this.

src/core/gosling-component.tsx
sehilyi: Agreed. Would be better to put a file closer to where it is used! Added a util file for this function.

PR #976, opened 2023-09-28
feat(core): replace `uuid` dependency with `crypto.randomUUID()`

manzt:
Fix #
Toward #

## Change List
 - Removes `uuid` and `@types/uuid` as direct dependencies
 - Adds `src/core/utils/uuid.ts`, which exports a `uuid` function

## Checklist


editor/Editor.tsx
```diff
@@ -723,7 +723,7 @@ function Editor(props: RouteComponentProps) {
                         }
                         return (
                             <div
-                                key={v4()}
+                                key={uuid()}
```
manzt, line 726: Just a side observation. I'm guessing the uuid is used here to please eslint-react, but I don't think key-ing with a UUID is desirable. Ideally it would be some piece of data from the array of objects that is deterministic between renders.

sehilyi: Thank you for finding this! This should have used the id of the view or track instead. I will update this after testing it.

etowahadams (approved):

etowahadams:
Looks good! Wondering if in the future we should also replace instances of `HGC.libraries.slugid.nice();` with this new uuid function too. Some of the datafetchers use this.

PR #977, opened 2023-09-28
chore(core): update build  

etowahadams:
Fix #
Toward #

## Change List
 - Removes /embed entrypoint 
 - Removes umd build 
 - Adds "exports" field to package.json 
 - Adds [knip](https://github.com/webpro/knip)
 - Removes unused dependencies 
 - Removes unused files 
 - Removes some unused methods
 - Make all CI/CD run on Node 20
 - Updates gosling.js HTML template to use module imports 
 - Upgrade vite and vitest
 - replace @vitejs/plugin-react-refresh with @vitejs/plugin-react

## Checklist


etowahadams:
The main thing remaining is to change `getHtmlTemplate` so that it uses esm.sh or something similar instead of the current unpkg. Will do that tomorrow. 

Also also I'm wondering if should get rid of knip since it only supports Node 16+. Or try downgrading it assuming it used to support Node 14. I cut most of the dependencies knip marked as unused. There's also a bunch of exports (`export const ...`) that don't get used outside of the file they're defined in. I might deleting the `export` from those later.

etowahadams:
I've been experimenting with using esm.sh. Because HiGlass is UMD and not ESM, esm.sh can't get find any of the exports from HiGlass that Gosling uses. 
```
Uncaught SyntaxError: The requested module '/v132/higlass@1.12.4/X-ZC9waXhpLmpzQDYuNS4xMCxyZWFjdC1kb21AMTcuMC4yLHJlYWN0QDE3LjAuMg/es2022/higlass.mjs' does not provide an export named 'HiGlassComponent' (at gosling.mjs:3:2125)
```
@manzt Do you think it would be worth updating HiGlass to give it a ESM build? I see that what happens now is that it creates a higlass.umd.js which gets transpiled using babel (with [plugin transform classes](https://babeljs.io/docs/babel-plugin-transform-classes)). 

The only alternative I can think of is to use importmaps. Its less clean but would probably get the job done. 

Here's the HTML I've been playing with:
```
<!DOCTYPE html>
<html>
  <head>
    <link rel="stylesheet" href="https://esm.sh/higlass@1.12/dist/hglib.css">
  </head>
  <body>
    <div id="gosling-container"></div>
  </body>
  <script type="module">
    // you can import submodules from esm.sh
    import { embed } from "https://esm.sh/gosling.js@0.11.0?deps=react@17,react-dom@17,pixi.js@6,higlass@1.12";
    embed(document.getElementById('gosling-container'), {
  "title": "Basic Marks: bar",
  "subtitle": "Tutorial Examples",
  "tracks": [
    {
      "layout": "linear",
      "width": 800,
      "height": 180,
      "data": {
        "url": "https://resgen.io/api/v1/tileset_info/?d=UvVPeLHuRDiYA3qwFlm7xQ",
        "type": "multivec",
        "row": "sample",
        "column": "position",
        "value": "peak",
        "categories": ["sample 1"],
        "binSize": 5
      },
      "mark": "bar",
      "x": {"field": "start", "type": "genomic", "axis": "bottom"},
      "y": {"field": "peak", "type": "quantitative", "axis": "right"},
      "size": {"value": 5}
    }
  ]
})
    </script>
</html>
```

manzt:
Excited to see progress on this! Will try to take closer look this weekend. But just wanted to chime in.

> Also also I'm wondering if should get rid of knip since it only supports Node 16+.

If gosling really just runs in the browser, I don't really see a motivation to support legacy versions of Node.js for development purposes. Node 14 and 16 are both EOL, meaning they are no longer being developed or will receive security fixes https://endoflife.date/nodejs). I'd look into using Node 20 and maybe consider just testing in that environment. Could reduce CI times.

> @manzt Do you think it would be worth updating HiGlass to give it a ESM build? I see that what happens now is that it creates a higlass.umd.js which gets transpiled using babel (with [plugin transform classes](https://babeljs.io/docs/babel-plugin-transform-classes)).

Yeah, I think that would be nice downstream. HiGlass will need to keep UMD for some time, due to its plugin system and other projects that rely on using it through script tags (many more higlass developers). 

However, one thing I've been very interested in there is breaking up higlass into smaller packages that are pure ESM. We would then have `@higlass/*` which something like gosling could depend on, and then just have a top level `higlass` that imports these things and publishes to npm.

etowahadams:
> I'd look into using Node 20 and maybe consider just testing in that environment. Could reduce CI times.

Good idea, I'll try it out!

> However, one thing I've been very interested in there is breaking up higlass into smaller packages that are pure ESM

Agreed and I would expect this to be relevant to our [previous discussion](https://hackmd.io/evjUHM33S_q2ucIUu4ZznQ) too.

etowahadams:
edit: using the`?bundle` option with esm.sh works! 

## esm.sh
I attempted with importmaps and the esm.sh CDN but without success. 

It seems that esm.sh doesn't have the pixi v6.x so it keeps using pixi v5.3.12. I noticed this because the error relates to not being able to import this es6-promise-polyfill package. Not sure why it can't import this package. 
```
/* esm.sh - esbuild bundle(@pixi/polyfill@5.3.12) es2022 production */
import {Polyfill as f} from "/v132/es6-promise-polyfill@1.2.0/es2022/es6-promise-polyfill.mjs";
```
Here is the html:
```
<!DOCTYPE html>
<html>
  <head>
    <link rel="stylesheet" href="https://unpkg.com/higlass@1.12/dist/hglib.css">

    <script type="importmap">
    {
    "imports": {
        "react": "https://esm.sh/react@17",
        "react-dom": "https://esm.sh/react-dom@17",
        "pixi.js": "https://esm.sh/pixi.js@6",
        "higlass": "https://esm.sh/higlass@1.12",
        "gosling.js": "https://esm.sh/gosling.js@0.11.0"
    }
    }
    </script>
<body>
    <div id="gosling-container"></div>
  
    <script type="module">
      import { embed } from 'gosling.js';
    embed(document.getElementById('gosling-container'), {
  .... same spec as above }
    </script>
</body>
</html>
```

## skypack.dev
I also tried using another CDN called skypack.dev, without success, although with a different error. 
```
Uncaught SyntaxError: The requested module 'gosling.js' does not provide an export named 'embed'
```
This error is a bit surprising because  `dist/gosling.es.js` does have an export called embed
```
export { GoslingComponent, GoslingSchema, GoslingTemplates, theme_schema as ThemeSchema, compile, embed, init, name, validateGoslingSpec, version };
```

## Next steps 

I'm wondering what the best way to go about troubleshooting this. Perhaps there is something obvious I'm missing, or perhaps there's some deeper understanding of how these CDNs work that I'm missing. My hunch is that there's something wrong with the higlass or gosling build that is causing these difficulties? More to investigate.

editor/html-template.ts
```diff
@@ -8,17 +8,27 @@ export const getHtmlTemplate = (spec: string) => `
...
+    <link rel="stylesheet" href="https://esm.sh/higlass@${higlassVersion}/dist/hglib.css">
+
+    <script type="importmap">
+    {
+    "imports": {
+        "react": "https://esm.sh/react@${reactVersion}?bundle",
+        "react-dom": "https://esm.sh/react-dom@${reactVersion}?bundle",
+        "pixi.js": "https://esm.sh/pixi.js@${pixiVersion}?bundle",
+        "higlass": "https://esm.sh/higlass@${higlassVersion}?bundle",
+        "gosling.js": "https://esm.sh/gosling.js@${goslingVersion}?bundle"
```
manzt: With `?bundle` I'm actually not sure gosling.js has any imports to resolve. I would be curious about just using:

```javascriprt
        import { embed } from 'https://esm.sh/gosling.js?bundle';
```

below. The main issue I issue I see with this is that I'm not sure how to control the version of HiGlass so that it aligns with the css.

etowahadams: esm.sh says 
> In bundle mode, all dependencies are bundled into a single JS file except the peer dependencies.

So actually I don't need to import Higlass! Just react, react-dom, and pixi.js

etowahadams: Actually you're right that does work! I'm surprised though. I'll have to read more about this bundle option. 
```
import { embed } from "https://esm.sh/gosling.js@0.11.0?bundle"
embed(document.getElementById('gosling-container'), {...}
```

etowahadams:
Node 20 has the Web Crypto API built in (globalThis.crypto). But Node 18 doesn't. 

Because of this, using changing crypto with `vi.stubGlobal` or `globalThis` or `window` works in Node 18 but not Node 20 (else you get a `TypeError: Cannot set property crypto of #<Object> which has only a getter`). 

What are our options? I'm pretty sure that @manzt 's PR https://github.com/gosling-lang/gosling.js/pull/976 would fix this since it includes a fallback for `crypto.randomUUID`. So perhaps we update the test Node versions after that PR? Or just only test with Node 20 which is probably fine since Gosling is mainly to be run in a browser?

manzt:
> Or just only test with Node 20 which is probably fine since Gosling is mainly to be run in a browser?

Ya, I'm in favor of just using 20. Especially since it better reflects the environment Gosling actually runs in. We just need to communicate the change/requirement to the gosling team. We could also look into adding an [`.nvmrc`](https://github.com/nvm-sh/nvm#nvmrc), so users of `nvm` and `fnm` automatically get the right version of node when the cd into the gosling.js.

etowahadams:
That makes sense to me. I'll add a note about testing using Node 20 to the contributing doc and will add a .nvmrc. We can make an announcement to the Gosling channel when this PR is finalized. 
cc: @sehilyi

package.json
```diff
@@ -8,21 +8,24 @@
         "url": "https://github.com/gosling-lang/gosling.js"
     },
     "homepage": "https://gosling-lang.github.io/gosling.js/",
-    "main": "dist/gosling.js",
+    "main": "dist/gosling.es.js",
```
etowahadams, line 11: "main" applies to both ES module and CJS module entry points. There is no `dist/gosling.js` created anymore so we point this to `gosling.es.js`

vite.config.js
```diff
@@ -1,11 +1,15 @@
 import { defineConfig } from 'vite';
 import reactRefresh from '@vitejs/plugin-react-refresh';
 import * as esbuild from 'esbuild';
-import path from 'path';
-import pkg from './package.json';
+import path, { dirname } from 'path';
+import { fileURLToPath } from 'url';
+import pkg from './package.json' assert { type: 'json' }; // must do the assert to solve https://nodejs.org/api/errors.html#err_import_assertion_type_missing
+
+const _dirname = fileURLToPath(dirname(import.meta.url));
```
etowahadams: There is no `__dirname` when `type: module` ([see here](https://nodejs.org/api/esm.html#no-__filename-or-__dirname)), so I replace this. However, node (?) still replaces `const __dirname` with the actual dirname so I changed the variable name to `_dirname`.

```
✘ [ERROR] Expected identifier but found "\"/Users/etowah/projects/gosling-lang/gosling.js\""

    vite.config.js:8:6:
      8 │ const "/Users/etowah/projects/gosling-lang/gosling.js" = fileURLToPath(dirname("file:///Users/etowah/projects/gosling-lang/gosling.js/vite.config.js"));
```

package.json
```diff
@@ -8,21 +8,24 @@
...
     "module": "dist/gosling.es.js",
     "types": "dist/src/index.d.ts",
     "files": [
         "dist"
     ],
+    "type": "module",
```
etowahadams, line 17: My understanding of this is that with the exception of some configs, we want all files to be treated as ESM so we use this option.

scripts/setup-vitest.js
```diff
@@ -25,14 +25,10 @@ apis.forEach(api => {
 });
 
 beforeAll(() => {
-    // jsdom doesn't come with a WebCrypto implementation (required for uuid)
-    global.crypto = {
-        getRandomValues: function (buffer) {
-            return randomFillSync(buffer);
-        }
-    };
```
etowahadams, line 33: No longer needed anymore since Web API crypto is now built in

vite.config.mjs
```diff
@@ -1,11 +1,15 @@
 import { defineConfig } from 'vite';
 import reactRefresh from '@vitejs/plugin-react-refresh';
 import * as esbuild from 'esbuild';
-import path from 'path';
-import pkg from './package.json';
+import path, { dirname } from 'path';
+import { fileURLToPath } from 'url';
+import pkg from './package.json' assert { type: 'json' }; // must do the assert to solve https://nodejs.org/api/errors.html#err_import_assertion_type_missing
+
+const __dirname = fileURLToPath(dirname(import.meta.url));
```
etowahadams: `__dirname` has been [replaced](https://nodejs.org/api/esm.html#no-__filename-or-__dirname) by import.meta.url

vite.config.mjs
etowahadams, line 1: Had to rename from `.js` to `.mjs` since vitest was ignoring `package.json` and kept using this file as if it were CJS

etowahadams:
Everything seems good in the production build I just tested. Anything else that would be good to add to this PR? Otherwise this is ready for review!

etowahadams:
Removed mention of `nvm` based on comments today

editor/html-template.ts
```diff
@@ -1,24 +1,19 @@
...
+    <link rel="stylesheet" href="https://esm.sh/higlass@${higlassVersion}/dist/hglib.css">
+    
 </head>
 <body>
-    <div id="gosling-container"/>
-    <script>
-        gosling.embed(document.getElementById('gosling-container'), ${spec})
+    <div id="gosling-container"></div>
+    <script type="module">
+        import { embed } from 'https://esm.sh/gosling.js@${goslingVersion}?bundle';
```
manzt: The only question I have here is whether the version of higlass and gosling.js will stay in sync over time. Unfortunately the names used in the styles are specific to the higlass version, so I don't know if there could be a case where higlass makes a patch (e.g., v1.11.8) but we have styles from 1.11.7 in our template and the two end up out of sync. I wonder if it's possible to do something like:

```javascript
import { embed } from 'https://esm.sh/gosling.js@${goslingVersion}?bundle&deps=higlass@${higlassVersion}';
```

but I'm not sure if the `deps` is handled when `bundle` is used.

embed/index.ts
```diff
@@ -1,33 +0,0 @@
...
-        (
-          (isString(args[0]) && !isURL(args[0])) ||
-          isElement(args[0]) ||
-          args.length === 3
-        )
-    ) {
-        return goslingEmbed(args[0], args[1], args[2]);
-    }
-    return container(args[0], args[1]);
-};
```
manzt, line 33: Only thing to note here is that we should make the "core" `embed` function handle usage in Observable, but that is ok for another PR I think. Basically overload the "core" embed do what `embed` here is doing, creating an element  depending on the number of arguments and returning the `wrapper`. We have this code as vega-embed as reference.

knip.config.json
```diff
@@ -0,0 +1,4 @@
+{
```
manzt, line 1: ❤️

package.json
manzt: Yup! Making things work nice (like Vite) might require upgrading those deps.

scripts/setup-vitest.js
manzt: yay

src/data-fetchers/gff/gff-worker.ts
```diff
@@ -26,7 +26,7 @@ export interface EmptyTile {
 /**
  * A class to represent a GFF file. It takes care of setting up gmod/tabix.
  */
-export class GffFile {
+class GffFile {
```
manzt, line 29: I'm guessing this is just for consistency?

vite.config.mjs
manzt: I see. Ya, this might be fixed with upgrading Vite (not sure).

manzt (approved):
Awesome. Thanks for all your work on this Etowah, looking great. Just one thought about esm.sh, otherwise LGTM!

manzt:
Ah, one thing I forgot. We have branch protection rules that require "Test (14.x)" and "Test (16.x)" to pass to merge into `master`. That's why "Some checks haven't completed yet". You removed these workflows in this PR, so that's why it's still waiting for them to complete. We should be ok to remove these rules in Settings (since they will never fulfill) and maybe add one for the node test workflow.

package.json
etowahadams: That makes sense! I did try upgrading Vite before, but it we're using a plugin ([@vitejs/plugin-react-refresh](https://www.npmjs.com/package/@vitejs/plugin-react-refresh)) which requires this version of vite. However, I didn't realize until now that the package is now deprecated in favor of [@vitejs/plugin-react](https://www.npmjs.com/package/@vitejs/plugin-react). 

I'll switch to @vitejs/plugin-react and upgrade vite if that sounds good!

editor/html-template.ts
sehilyi: Is it possible to use different react and react-dom versions as well that user wishes to use?

package.json
```diff
@@ -34,27 +37,21 @@
         "schema": "node scripts/generate-schemas.mjs",
         "predeploy": "yarn build-editor; echo \"gosling.js.org\" >> build/CNAME",
         "deploy": "gh-pages -d build",
-        "version": "conventional-changelog -p angular -i CHANGELOG.md -s && git add CHANGELOG.md"
+        "version": "conventional-changelog -p angular -i CHANGELOG.md -s && git add CHANGELOG.md",
+        "knip": "knip --config knip.config.json"
```
sehilyi, line 41: This is a nice addition! I see many unused components as an output :)

package.json
```diff
@@ -34,27 +37,21 @@
...
         "@gmod/bed": "^2.1.2",
         "@gmod/gff": "^1.3.0",
         "@gmod/tabix": "^1.5.6",
         "@gmod/vcf": "^5.0.10",
-        "@types/bezier-js": "^4.1.0",
-        "@types/d3": "^7.0.0",
-        "@types/lodash": "^4.14.151",
-        "@types/node": "^18.6.2",
-        "@types/rbush": "^3.0.0",
-        "@types/uuid": "^8.3.1",
```
sehilyi, line 57: Thank you!

src/missing-types.d.ts
```diff
@@ -26,8 +26,6 @@ declare module '@higlass/libraries' {
     export * as d3Geo from 'd3-geo';
     export * as d3Scale from 'd3-scale';
     export * as d3Selection from 'd3-selection';
-    // export * as d3Queue from 'd3-queue';
-    // export * as d3Request from 'd3-request';
```
sehilyi, line 30: Is this intended? Could be removed if not needed.

src/tracks/gosling-track/gosling-track-model.ts
```diff
@@ -115,10 +115,6 @@ export class GoslingTrackModel {
         return this.id;
     }
 
-    public getRenderingId(): string {
-        return this.spec()._renderingId ?? this.getId();
-    }
-
     public originalSpec(): SingleTrack {
```
sehilyi, line 118: I assume the changes in this file are that unused functions are removed.

sehilyi (approved):

sehilyi:
@manzt @etowahadams Removed the "Test (14.x)" and "Test (16.x)" branch protection rules. I will add one for 20.x after this is merged.

editor/html-template.ts
manzt: Same question. Ideally you could bundle _and_ override dependency versions, but I'm not sure for ESM.sh if that is current possible.

embed/index.ts
etowahadams: Makes sense, added an issue for follow-up PR! https://github.com/gosling-lang/gosling.js/issues/981

src/data-fetchers/gff/gff-worker.ts
etowahadams: Yep! I started addressing some of the unused exports that knip identified.

src/tracks/gosling-track/gosling-track-model.ts
etowahadams: Exactly! Knip identified these.

package.json
etowahadams: Feels good to get rid of these!

src/missing-types.d.ts
etowahadams: The exports for`d3Queue` and `d3Request` were commented out so I just removed them. Perhaps they're unused in the parts of HiGlass that Gosling uses?

vite.config.mjs
etowahadams: That did the trick! Upgraded Vite to the latest version

package.json
etowahadams: Trevor suggested the tool! Seems quite useful for this type of cleanup.

editor/html-template.ts
etowahadams: Just tested. Unfortunately, it seems to ignore the version of the dependencies. I tried using a lower version of HiGlass and I still got `HiGlass v1.12.4: http://higlass.io/` in the console. What I'll try next is to use the pre-bundled peer dependencies in an import map. 

```javascript
import { embed } from 'https://esm.sh/gosling.js@0.11.0?bundle&deps=higlass@$1.11.7';
```

etowahadams: I didn't have any luck using unbundled gosling.js with bundled peer dependencies either using an importmap. 

The problem seems to be that HiGlass has `"pixi.js": "^5.0.3"` as a peer dependency. One of the pixi subpackages  called `@pixi/polyfill` relies on a package called `es6-promise-polyfill` which doesn't have the export that @pixi/polyfill expects. Thus I get a `The requested module '/v133/es6-promise-polyfill@1.2.0/es2022/es6-promise-polyfill.mjs' does not provide an export named 'Polyfill'` error when trying to import HiGlass. 

Looking at that version of es6-promise-pollyfill it does look like there is export called [Polyfill](https://github.com/lahmatiy/es6-promise-polyfill/blob/v1.2.0/promise.js#L29-L50) but I don't think it gets converted in a way that is accessible in the [ESM version](https://esm.sh/v133/es6-promise-polyfill@1.2.0/es2022/es6-promise-polyfill.mjs) that esm.sh generates. 

How hard it would it be to upgrade the pixi version that HiGlass uses to v6?

etowahadams: Added a PR to HiGlass to upgrade Pixi https://github.com/gosling-lang/gosling.js/pull/977

manzt:
[Released](https://github.com/higlass/higlass/releases/tag/v1.13.0) HiGlass v1.13

etowahadams:
Amazing thanks @manzt! Will try it out tomorrow.

etowahadams:
The polyfill error is resolved! Now encountering a new error due to HiGlass not being ESM. 

```jsx
// [OLD] This still uses higlass@1.12.4 and we get the polyfill import error
import { embed } from 'https://esm.sh/gosling.js@0.11.0';

// [NEW] This uses higlass@1.13
// No more polyfill error (yay!). However, we get a new error
// higlass.mjs does not provide an export named 'HiGlassComponent
import { embed } from 'https://esm.sh/gosling.js@0.11.0?deps=higlass@1.13'
```

Why is this failing? higlass.js is a UMD module. Esm.sh converts this into a ES module which has a single export: 

```jsx
// https://esm.sh/v133/higlass@1.13.0/es2022/higlass.mjs 
export {oNe as default};
```

So in `gosling.mjs`, this fails:  

```jsx
import { HiGlassComponent as ps } from "/v133/higlass@1.13.0/es2022/higlass.mjs";
// ERROR: higlass.mjs does not provide an export named 'HiGlassComponent 
```

## Potential solutions

### Add feature to esm.sh

One solution which would require adding a feature to esm.sh. Currently esm.sh allows you to [specify named](https://esm.sh/#specify-cjs-exports) exports from CJS file: 

```jsx
import { HiGlassComponent } from "https://esm.sh/higlass@1.13?cjs-exports=HiGlassComponent";
```

This gets turned into the following in `higlass.mjs`, allowing `HiGlassComponent` to be imported.  

```jsx
export * from "/v133/higlass@1.13.0/X-dHMvSGlHbGFzc0NvbXBvbmVudA/es2022/higlass.mjs";
import __cjs_exports$ from "/v133/higlass@1.13.0/X-dHMvSGlHbGFzc0NvbXBvbmVudA/es2022/higlass.mjs";
export const { HiGlassComponent } = __cjs_exports$;
```

However, you can’t combine this feature with the `?deps` option

```jsx
// This doesn't work :(
import { embed } from 'https://esm.sh/gosling.js@0.11.0?deps=higlass@1.13?cjs-exports=HiGlassComponent'
```

So we would have to submit a PR to esm.sh to add this feature. 

### ESM build for Higlass

Once again we return to this. I’m ready to make this a priority if others are onboard with me doing so. 

### Add back the UMD build

We could also just bring back the UMD build as a short term fix to merge this PR. This is tempting. 

Thoughts? @manzt @sehilyi

etowahadams:
Just waiting on https://github.com/higlass/higlass/pull/1171 and we should be good to go!

etowahadams:
This should be everything. Thanks for all the work on HiGlass Trevor!

PR #978, opened 2023-10-07
feat(api): publish traversed spec to gosling api

thomcsmits:
Fix #
Toward #

## Change List
 - Added the spec as changed by overrideDataTemplates, replaceTrackTemplates and most importantly traverseToFixSpecDownstream to the Gosling api. This is used in the AltGoslingComponent.

## Checklist
 - [X] Ensure the PR works with all demos on the online editor
 - [X] [Documentation](https://github.com/gosling-lang/gosling-website) updated (e.g., added API functions)

### DIFF
--- a/src/compiler/compile.ts
+++ b/src/compiler/compile.ts
@@ -73,13 +73,13 @@ export function compile(
     // used for alt-gosling
     try {
         if (PubSub) {
-            PubSub.publish('spec-traversed', {
+            PubSub.publish('specTraversed', {
                 id: specCopy.id,
-                data:
-                    specCopy
+                data: specCopy
             });
         }
     } catch (e) {
+        //
     }
 
     // Make HiGlass models for individual tracks
--- a/src/compiler/compile.ts
+++ b/src/compiler/compile.ts
@@ -8,6 +8,7 @@ import type { UrlToFetchOptions } from 'src/core/gosling-component';
 import { renderHiGlass as createHiGlassModels } from './create-higlass-models';
 import { manageResponsiveSpecs } from './responsive';
 import type { IdTable } from '../api/track-and-view-ids';
+import { publish } from '../api/pubsub';
 
 /** The callback function called everytime after the spec has been compiled */
 export type CompileCallback = (
@@ -71,16 +72,10 @@ export function compile(
 
     // publish the fixed spec
     // used for alt-gosling
-    try {
-        if (PubSub) {
-            PubSub.publish('specTraversed', {
-                id: specCopy.id,
-                data: specCopy
-            });
-        }
-    } catch (e) {
-        //
-    }
+    publish('specProcessed', {
+        id: specCopy.id,
+        spec: specCopy
+    });
 
     // Make HiGlass models for individual tracks
     createHiGlassModels(specCopy, trackInfos, callback, theme, urlToFetchOptions);
--- a/src/compiler/compile.ts
+++ b/src/compiler/compile.ts
@@ -69,6 +69,19 @@ export function compile(
         trackInfos = getRelativeTrackInfo(specCopy, theme).trackInfos;
     }
 
+    // publish the fixed spec
+    // used for alt-gosling
+    try {
+        if (PubSub) {
+            PubSub.publish('spec-traversed', {
+                id: specCopy.id,
+                data:
+                    specCopy
+            });
+        }
+    } catch (e) {
+    }
+
     // Make HiGlass models for individual tracks
     createHiGlassModels(specCopy, trackInfos, callback, theme, urlToFetchOptions);
 }
--- a/src/gosling-schema/gosling.schema.ts
+++ b/src/gosling-schema/gosling.schema.ts
@@ -284,6 +284,13 @@ interface CommonEventData {
     data: Datum[];
 }
 
+interface SpecEventData {
+    /** Source visualization ID, i.e., `track.id` */
+    id: string;
+    /** Gosling spec */
+    spec: GoslingSpec;
+}
+
 export interface GenomicPosition {
     chromosome: string;
     position: number;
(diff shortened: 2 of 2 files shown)


sehilyi (commented):
Thank you for making this change! 

One comment: Could we make this a 'standard' Gosling API function? This will require a small change, which are

- Adding `specTraversed` as an API event type: https://github.com/gosling-lang/gosling.js/blob/71a2a28b0739f452389990603c7bff7911536ffd/src/gosling-schema/gosling.schema.ts#L370-L380
- Replace the code to publish the event, like how we do for other API events, e.g., https://github.com/gosling-lang/gosling.js/blob/71a2a28b0739f452389990603c7bff7911536ffd/src/tracks/gosling-track/gosling-track.ts#L478-L481 

This way, users can use the `gosRef.current.api.subscribe('specTraversed', (...args) ⇒ console.log(args));` pattern.

src/compiler/compile.ts
```diff
@@ -69,6 +69,19 @@ export function compile(
         trackInfos = getRelativeTrackInfo(specCopy, theme).trackInfos;
     }
 
+    // publish the fixed spec
+    // used for alt-gosling
+    try {
+        if (PubSub) {
+            PubSub.publish('specTraversed', {
+                id: specCopy.id,
+                data: specCopy
```
sehilyi: Maybe more intuitive and consistent with how we use "data" and "spec" in other API functions

```suggestion
                spec: specCopy
```

src/compiler/compile.ts
```diff
@@ -69,6 +69,19 @@ export function compile(
         trackInfos = getRelativeTrackInfo(specCopy, theme).trackInfos;
     }
 
+    // publish the fixed spec
+    // used for alt-gosling
+    try {
+        if (PubSub) {
+            PubSub.publish('specTraversed', {
```
sehilyi: Regarding the name of the event, how about `specProcessed` instead? I think users might not understand what the traverse means. And, we can explain what is the different between the input spec and the processed spec briefly in the documentation.

thomcsmits: That sounds like a good idea! I indeed was fidgetting around with a good name, specProcessed is more intuitive

src/compiler/compile.ts
thomcsmits: Ah I see, I had already written this piece before the [standardization](https://github.com/gosling-lang/gosling.js/commit/1c2b7669660856cae98f88ba38d9326b788c75e1)! Thanks!

sehilyi (approved):
Thanks for making updates! They looks good to me.

src/compiler/compile.ts
manzt: Sorry, bikeshedding, maybe `specResolved`?

sehilyi: I like `specResolved`, which I think gives a better impression that the spec is not converted into another format, but rather the spec is slightly modified to resolve things!

thomcsmits: Wouldn't resolved give the idea that the spec is fully changed into the 'final form' that it gets for gosling? Because that isn't the case here, it is just partially processed. I'm not sure what the convention is, but I'm afraid resolved gives off the wrong idea?

thomcsmits: Happy to change it though

sehilyi: Ah, I had the opposite idea! At this point, I don't have a strong preference between the two. Let's leave it as is and merge this when you think is ready.

manzt: > Because that isn't the case here, it is just partially processed.

Sorry to dwell on this. What would we name an event that emitted a "fully processed" spec? Perhaps naming the event something more specific would be useful. How has the input been modified at this point in the program?

src/compiler/compile.ts
```diff
@@ -69,6 +69,19 @@ export function compile(
...
+                id: specCopy.id,
+                data: specCopy
+            });
+        }
+    } catch (e) {
+        //
+    }
+
```
manzt: Just noticed this `try` /`catch` without any error handling. Is it expected that this could fail with some frequency and we wouldn't want to propagate that to developer?

thomcsmits: I had already taken it out to mimic https://github.com/gosling-lang/gosling.js/blob/71a2a28b0739f452389990603c7bff7911536ffd/src/tracks/gosling-track/gosling-track.ts#L478-L481 (see latest changes)

I had originally used this similar to the deprecated rawData publishing https://github.com/gosling-lang/gosling.js/blob/f17e69f8ea9d83e9c574e864b7dfca94a49fa328/src/tracks/gosling-track/gosling-track.ts#L921

src/compiler/compile.ts
sehilyi: > How has the input been modified at this point in the program?

The input spec is modified to have default values filled in, and some of the users' incorrectly-specified properties are fixed. Some parts of the spec are not touched by the program (e.g., when values are already specified by users, so no need to add default values), which I think is what @thomcsmits was referring to by `"it is just partially processed"`.

thomcsmits: Hm, we don't emit a fully processed spec ("fully processed" as in the final state that it becomes in Gosling), but that might happen in the future

thomcsmits: I don't have truely strong opinions on this, so I'm happy to switch it to specResolved still

thomcsmits: https://github.com/gosling-lang/gosling.js/pull/985 (as this PR is already merged)

src/compiler/compile.ts
manzt: Ah got it! Sorry I missed the `outdated` tag

src/compiler/compile.ts
manzt: I don't really have a strong opinion about "resolved" vs "processed", but think this could be a good exercise to be explicit about the state of the gosling spec when it is emitted. "Resolved" and "processed" convey that completion of some transformations/resolution, which is not the case. I have left a longer comment in #985

PR #985, opened 2023-10-19
feat(api): rename specProcessed to specResolved

thomcsmits:
Fix # Change api name based on https://github.com/gosling-lang/gosling.js/pull/978#discussion_r1359666778
Toward #

## Change List
 - Change name of api subscribe event specProcessed to specResolved

## Checklist
 - [X] [Documentation](https://github.com/gosling-lang/gosling-website) updated (e.g., added API functions)

### DIFF
--- a/src/compiler/compile.ts
+++ b/src/compiler/compile.ts
@@ -72,7 +72,7 @@ export function compile(
 
     // publish the fixed spec
     // used for alt-gosling
-    publish('specProcessed', {
+    publish('specResolved', {
         id: specCopy.id,
         spec: specCopy
     });
--- a/src/gosling-schema/gosling.schema.ts
+++ b/src/gosling-schema/gosling.schema.ts
@@ -379,7 +379,7 @@ export type _EventMap = {
     click: PointMouseEventData;
     rangeSelect: RangeMouseEventData;
     rawData: CommonEventData;
-    specProcessed: SpecEventData;
+    specResolved: SpecEventData;
     trackMouseOver: TrackApiData;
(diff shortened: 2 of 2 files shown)


manzt:
> The input spec is modified to have default values filled in, and some of the users' incorrectly-specified properties are fixed. Some parts of the spec are not touched by the program (e.g., when values are already specified by users, so no need to add default values), which I think is what @thomcsmits was referring to by "it is just partially processed".

Hmm hard for me to visualize the "state" of the spec at this moment. Similar to our discussion in the developer meeting, perhaps it would be useful to think of the spec in stages of transformations (e.g., a different type for each stage) and then name the event based on the stage. 

My only comment about both `specProcessed` and `specResolved` is that these both are ambiguous and don't capture the "partial" nature of the processing. Imagine a pipeline of processing steps in the compiler:

```
inputSpec ---(pass 1)--> specStage1 --(pass 2)--> specStage2 --(pass 3)--> finalSpec
```

Let's say right now we emit the `specStage2` version of the spec. What would we call events publishing the `finalSpec` or `specStage1`? My suggestion was to try to come up with a more specific name than either `specProcessed` or `specResolved` that appropriately captures the state of the data being sent. 

I don't think we need to bikeshed about processed vs resolved, both are totally fine for me. Since the only current subscriber for this event is alt-gosling, maybe we could use something like `alt-gosling:specProcessed` in lieu of codifying the names/types for the steps of the Gosling compiler.

sehilyi:
I am down with `alt-gosling:specProcessed` given that alt-gosling is the only use case for this at the moment.

Issue #823, opened 2022-10-20
Improve performance of json-type by removing deep-clones of spec/data


puehringer:
When inlining long arrays (>10k objects) as `values` in a spec via the `type: json`, zooming and panning becomes painfully slow. We identified the key reason for this to be the deep-cloning/structuredClone of the spec which happens in several places. See this (draft) PR for the places where we removed the deep-cloning, without any breaking tests/demos: https://github.com/datavisyn/gosling.js/pull/1

We know that the main point of using higlass is to avoid plotting that many points, however the performance improvements are so drastic that refactoring towards less deep-cloning operations should still be considered. 

It should also be noted that the `sampleLength` property **does not** have any influence on this, because the bottleneck really is the cloning of the entire values array, causing gosling to be slow even with `sampleLength: 100`. 

## Slow demo (v0.9.25) 
[slow_demo](https://user-images.githubusercontent.com/51900829/196889746-bc7e282a-8dc0-4600-b09a-f4489a9ca7f8.webm) 

(ignore the title, it's the slow version)

## Fast demo (without the deep-cloning, using the built version of https://github.com/datavisyn/gosling.js/pull/1)

[fast_demo](https://user-images.githubusercontent.com/51900829/196888932-4dc99602-665c-4b74-b592-6be88f6b28a8.webm)

## Open questions: 
- [ ] see the ones in https://github.com/datavisyn/gosling.js/pull/1

sehilyi:
Thank you @puehringer for exploring this! The performance improvement is really interesting to see. 

The only reason I was using the deep copy frequently was to safely keep the original (or intermediate) specs. For example, `this.specComplete` copies the original spec so that any values added to this later (e.g., missing values like `domain`) are not added to the original spec.

https://github.com/gosling-lang/gosling.js/blob/7edb6c0bc31e2245de091f342e8ffb4b07dd4427/src/core/gosling-track-model.ts#L84

https://github.com/gosling-lang/gosling.js/blob/7edb6c0bc31e2245de091f342e8ffb4b07dd4427/src/core/gosling-track-model.ts#L97

Similarly, whenever the Gosling API exposes the spec to users, I am deep cloning before sending it to users so that whenever users edit the spec, it does not affect to the spec used inside the GoslingComponent.

```ts
publish(eventType, {
    id: context.viewUid,
    spec: structuredClone(this.options.spec), // instead of spec: `spec: this.options.spec,`
    shape: { x, y, width, height }
});

...
[...]

PR #988, opened 2023-11-02
feat(track): performance improvements 

etowahadams:
Fix #
Toward #

## Change List
 - Adds playwright
 - Add playwright test which measures how long it takes to zoom in an example spec (from https://github.com/gosling-lang/gosling.js/issues/823#issuecomment-1782524800) 
 - Persist resolved spec as a property of `GoslingTrackClass` so that it doesn't have to be regenerated every time 
 - Minimize number of deep clones of the spec or the data. 

## Checklist


etowahadams:
Zooming works (after much trial and error). However it is still more choppy than I would like. Not sure that there's an obvious solution.

etowahadams:
For reasons still unclear to me, it seems that occasionally the first zoom/pan event has a lower frame rate compared to subsequent interactions.

etowahadams:
Last thing to do here is just to add the same visual regression testing from https://github.com/gosling-lang/gosling.js/pull/962

etowahadams:
test

etowahadams:
I spent quite a bit of time getting automatic snapshot testing working in CI, but without luck. Maybe it needs to be merged before it can be triggered from a comment? Will have to do additional testing. Going to reduce the scope of this to just adding the performance test and the performance related changes.

src/tracks/gosling-track/gosling-track-model.ts
```diff
@@ -77,11 +76,10 @@ export class GoslingTrackModel {
 
         this.theme = theme ?? getTheme();
 
-        this.dataOriginal = JSON.parse(JSON.stringify(data));
-        this.dataAggregated = JSON.parse(JSON.stringify(data));
+        this.dataAggregated = data; // this will be updated after validity of the spec is checked
 
-        this.specOriginal = JSON.parse(JSON.stringify(spec));
-        this.specComplete = JSON.parse(JSON.stringify(spec));
+        this.specOriginal = spec;
```
etowahadams, line 81: specOrigional never gets modified so we don't need to make a clone of it

src/tracks/gosling-track/gosling-track.ts
```diff
@@ -801,12 +802,24 @@ const factory: PluginTrackFactory<Tile, GoslingTrackOptions> = (HGC, context, op
...
+                const copy = structuredClone(this.options.spec);
+                const tracks = resolveSuperposedTracks(copy).filter(t => t.mark !== 'brush');
+                // We will never need to access the values field in the data spec. It can be quite large which can degrade performance so we remove it.
+                tracks.forEach(track => {
+                    if ('values' in track.data) {
+                        track.data.values = [];
+                    }
+                });
+                this.resolvedTracks = tracks;
+            }
```
etowahadams: Rather than creating a resolved spec every time, we make it once and reuse it.

etowahadams: I tried to come up with ways in creating a new resolved spec at every render is needed but couldn't think of one. Sehi do you know if there are any?

src/tracks/gosling-track/gosling-track-model.ts
sehilyi: Nice catch!

src/tracks/gosling-track/gosling-track.ts
```diff
@@ -801,12 +802,24 @@ const factory: PluginTrackFactory<Tile, GoslingTrackOptions> = (HGC, context, op
...
+        getResolvedTracks() {
+            if (!this.resolvedTracks) {
+                const copy = structuredClone(this.options.spec);
+                const tracks = resolveSuperposedTracks(copy).filter(t => t.mark !== 'brush');
+                // We will never need to access the values field in the data spec. It can be quite large which can degrade performance so we remove it.
+                tracks.forEach(track => {
+                    if ('values' in track.data) {
+                        track.data.values = [];
+                    }
+                });
```
sehilyi, line 817: ❤️

src/tracks/gosling-track/gosling-track.ts
sehilyi: It is possible that the track spec can be edited by users even after the track has been rendered, so we would need to force updating `resolvedTracks` when this happens. `override rerender(newOptions: GoslingTrackOptions)` is the function that is called when a user edits a track spec, so we can update `resolvedTracks` inside the function.

I don't think this will decrease the performance though since this function is called only when the user manually edits the spec.

sehilyi (commented):
I reviewed the code changes mainly related to the renderer, i.e., changes in gosling-track.ts and gosling-track-model.ts, and left one comment to force updating `resolvedTracks` when a user manually edits the spec. I will need to find more time to look at other updates (playwright), but please feel free to merge this after addressing the comment.

src/tracks/gosling-track/gosling-track.ts
etowahadams: That's exactly what I was looking for, thank you! Just changed.

PR #1001, opened 2023-11-29
feat(core): Create `gosling.js/utils` entrypoint for exporting utility funcs

manzt:
Allows import of some whitelisted functions via separate entrypoint (much simpler than #1000):

```javascript
import { getRelativeGenomicPosition } from 'gosling.js/utils';
```

...at the cost of some build complexity. If you look at the output, you'll see that `dist/utils.ts` contains imports for `d3-array` and `generic-filehandle` even though the exported functions don't use these imports at all.

```javascript
import { c, b, d, a, p, s } from "./exported-utils-068d7b17.js";
import "d3-array";
import "generic-filehandle";
export {
  c as computeChromSizes,
  b as getChromInterval,
  d as getChromTotalSize,
  a as getRelativeGenomicPosition,
  p as parseGenomicPosition,
  s as sanitizeChrName
};
//# sourceMappingURL=exported-utils.js.map
```

This is because those dependencies do not have `sideEffect: false` in the package.json, which means Vite doesn't know if importing the modules have global side effects. Should be fine in this context, but just so folks know what's going on.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,10 @@
         ".": {
             "types": "./dist/src/index.d.ts",
             "import": "./dist/gosling.js"
+        },
+        "./utils": {
+            "types": "./dist/src/exported-utils.d.ts",
+            "import": "./dist/utils.js"
         }
     },
     "scripts": {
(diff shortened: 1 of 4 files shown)


manzt:
Requested syntax in #983

sehilyi:
This is nice, thank you! 😊

PR #1018, opened 2023-12-20
perf(core): stretch certain marks instead of redraw

### LINKED ISSUE #1006: Performance: Stretch and translate instead of redraw
### Goal
Improve the performance of Gosling by reducing the number of times that shapes are redrawn. Instead of redrawing, we can stretch and translate the shapes when appropriate. 

### Pixi Basics

[Reference tutorial](https://github.com/kittykatattack/learningPixi)

A `PIXI.Container` is like an empty box where you display and group display objects. These include `PIXI.Sprites` which renders a `PIXI.Texture` (an image), and `PIXI.Graphics` which are primitive shapes and lines. 

For example, all of the marks (except text) in Gosling are `PIXI.Graphics` objects and text is a `PIXI.Sprite`. 

Laying out the terms here to avoid confusion: 

**Texture:**
- Created from various sources like images, canvases, videos, and even other Textures
- Cannot be directly displayed, but needs to be used with Sprites or Graphics.

**Sprite:**
- A simple display object that renders a single Texture.

**Graphics:**
- Low level object that allows you to draw shapes and lines (`drawRect`, `drawCircle`)
- Graphics objects can have other Graphics objects as children. Child Graphics objects will be rendered on top

**Container:**
- A display object that acts as a parent for other display objects, allowing you to group multiple Sprites, Graphics, or even other Containers together.

### Pixi in HiGlass

Where does the `PIXI.Container` used in HiGlass get instantiated? And how do `PIXI.Graphics` objects that Gosling draws things to (such as `Tile.graphics`) to get added to that container? 

- The main `PIXI.Container` used in HiGlass is created in the `HiGlassComponent`. All tracks that use Pixi will eventually add their `PIXI.Graphics` objects to other `PIXI.Graphics` objects which are inside of this `PIXI.Container`.
- Every `Tile` (the basic unit of data in HiGlass/Gosling) has a `graphics` property  (`Tile.graphics`) which is set to a `PIXI.Graphics` object. Gosling adds shapes and lines to this graphics object. `TiledPixiTrack`, which `GoslingTrack` extends, adds the `Tile.graphics` objects to `PixiTrack.pMain` which is a child of the `PIXI.Container` created in `HiGlassComponent`

```jsx
// In HiGlassComponent
this.pixiRoot = new GLOBALS.PIXI.Container();
this.pixiStage = new GLOBALS.PIXI.Container();
this.pixiRoot.addChild(this.pixiStage);

// In TrackRenderer
this.pStage = new GLOBALS.PIXI.Graphics();
this.currentProps.pixiStage.addChild(this.pStage);
const context = {scene: this.pStage}
new PixiTrack(context, options)

// PixiTrack.js
const { scene } = context 
this.scene = scene // therefore is pStage, a new PIXI.Graphics() object
this.pBase = new GLOBALS.PIXI.Graphics();
this.pMasked = new GLOBALS.PIXI.Graphics();
this.pMain = new GLOBALS.PIXI.Graphics();
this.scene.addChild(this.pBase);
this.pBase.addChild(this.pMasked);
this.pMasked.addChild(this.pMain);

// The hierarchy in summary: 
pixiRoot = new GLOBALS.PIXI.Container();
pixiStage = new GLOBALS.PIXI.Container(); // gets added to pixiRoot
pStage = new GLOBALS.PIXI.Graphics(); // gets added to pixiStage
pBase = new GLOBALS.PIXI.Graphics(); // gets added to pStage
pMasked = new GLOBALS.PIXI.Graphics(); // gets added to pBase
pMain = new GLOBALS.PIXI.Graphics(); // gets added to pMasked
// Then Tile.graphics will get added to pMain 

```

```jsx
export type Tile = {
  tileId: string;
  remoteId: string;
  drawnAtScale: Scale;
  graphics: PIXI.Graphics; // Every Tile has a graphics property 
  tileData: TileData;
};
```

What is the relationship between `GoslingTrack` and `PixiTrack`? 

- `GoslingTrack` extends `BarTrack` extends `HorizontalLine1DPixiTrack` extends `HorizontalTiled1DPixiTrack` extends `Tiled1DPixiTrack` extends `TiledPixiTrack` extends `PixiTrack` extends `Track`. `Track` is the base class for most HiGlass tracks.

### Stretching and translating instead of redrawing

When a user zooms in, we want the marks that are drawn to simply be stretched and translated, rather than redrawn completely. 

Currently when a user zooms or pans, all of the shapes and lines which are drawn to the `PIXI.Graphics` object (`Tile.graphics`) are cleared (`Tile.graphics.clear()`) and everything is redrawn. 

In other HiGlass tracks, this is avoided by instead stretching or translating the shapes. For example, there is a function called `stretchRects` that is used by the gene annotation and Bed viewer track that stretches the drawn shapes.  

```jsx
/** When zooming, we want to do fast scaling rather than re-rendering
 * but if we do this too much, shapes (particularly arrowheads) get
 * distorted. This function keeps track of how stretched the track is
 * and redraws it if it becomes too distorted (tileK < 0.5 or tileK > 2) */
function stretchRects(track, graphicsAccessors) {
}
```

We would like to do something similar in Gosling. 

When `draw()` gets called in GoslingTrack, we want to stretch and translate the marks when appropriate to avoid redrawing them over and over again. 

For inspiration:

- [HorizontalGeneAnnotationsTrack](https://github.com/higlass/higlass/blob/13b868b8ea17fef567180b1978a85014944da18c/app/scripts/HorizontalGeneAnnotationsTrack.js#L510)
- [BedLikeTrack](https://github.com/higlass/higlass/blob/13b868b8ea17fef567180b1978a85014944da18c/app/scripts/BedLikeTrack.js#L43)
    - Functions like `movedY` and `zoomedY`

etowahadams:
Fix #1006
Toward #

## Change List
 - Stretches tiles instead of redrawing them if they they have stretchable marks (`'bar', 'line', 'rect', 'area') and is not a circular or 2D track. 
 - Adds an experimental `stretchTiles` option. If true, all tiles will be stretched. If false, no tiles will be stretched. The threshold the tile can be stretched to can be tuned with `stretchTilesThreshold` option. 

## Checklist


etowahadams:
Stretching individual shapes within a graphics object isn't allowed; you can only stretch the whole graphics object apparently. Perhaps another experiment to try would be to have all of the `drawCircles` etc be drawn to a Graphics object that is a child of `tile.graphics`, not to `tile.graphics` directly.

src/tracks/gosling-track/gosling-track.ts
```diff
@@ -1472,6 +1501,31 @@ const factory: PluginTrackFactory<Tile, GoslingTrackOptions> = (HGC, context, op
...
+        }
+
+        /**
+         * Used in drawTile()
+         * Checks if the tile Graphic is too stretched. If so, it returns true.
+         * @param stretchFactor The factor by which the tile is stretched
+         * @returns True if the tile is too stretched, false otherwise
+         */
+        #isTooStretched(stretchFactor: number) {
+            return stretchFactor > 2 || stretchFactor < 0.5;
```
sehilyi: Could we add another experimental option for controlling the threshold? I would like to experiment with this using some examples.

```suggestion
            const threshold = this.options.spec.experimental?.stretchGraphicsThreshold ?? 2;
            return stretchFactor > threshold || stretchFactor < 1 / threshold;
```

src/tracks/gosling-track/gosling-track.ts
```diff
@@ -1,4 +1,4 @@
-import type * as PIXI from 'pixi.js';
+import * as PIXI from 'pixi.js';
```
sehilyi: Is the removal of `type` intended here?

src/tracks/gosling-track/gosling-track.ts
```diff
@@ -333,14 +333,43 @@ const factory: PluginTrackFactory<Tile, GoslingTrackOptions> = (HGC, context, op
...
+                // This is the first time this tile is being drawn
+                tile.drawnAtScale = this._xScale.copy();
+            }
+
+            /**
+             * For certain types of marks and layouts (linear), we can stretch the graphics to avoid redrawing
+             * This is much more performant than redrawing everything at every frame
+             */
+            const [graphicsXScale, graphicsXPos] = this.getXScaleAndOffset(tile.drawnAtScale);
+            if (this.#hasStretchableGraphics() && !this.#isTooStretched(graphicsXScale) && graphicsXScale !== 1) {
```
sehilyi: Could you remind me what it means that `graphicsXScale` is equal to `1`?

src/tracks/gosling-track/gosling-track.ts
```diff
@@ -1472,6 +1501,31 @@ const factory: PluginTrackFactory<Tile, GoslingTrackOptions> = (HGC, context, op
...
+         * Checks if the track has marks which are stretchable. Stretching
+         * is not supported for circular layouts or 2D tracks
+         */
+        #hasStretchableGraphics() {
+            const stretchableMarks = ['bar', 'line', 'rect', 'area'];
+            return (
+                this.options.spec.experimental?.stretchGraphics ||
+                (!Is2DTrack(this.getResolvedTracks()[0]) &&
+                    this.options.spec.layout !== 'circular' &&
+                    stretchableMarks.includes(this.options.spec.mark || ''))
```
sehilyi: Could we still enable testing the original rendering with `stretchGraphics: false`?

```suggestion
                (typeof this.options.spec.experimental?.stretchGraphics  == 'undefined' &&
                    !Is2DTrack(this.getResolvedTracks()[0]) &&
                    this.options.spec.layout !== 'circular' &&
                    stretchableMarks.includes(this.options.spec.mark || ''))
```

sehilyi: It is possible that a track that superposes multiple sub-tracks can have different mark types defined by the sub-tracks (e.g., rect, text, and triangles in ideograms). So, in some cases, it may end up using the new rendering method even though there are un-stretchable marks being used by sub-tracks. e.g., 

```ts
alignment: 'overlay',
mark: 'rect',
tracks: [
   { ... },
   { mark: 'text', ...}
]
```

<a href="http://localhost:5173/?full=false&spec=(%0A*'alignmentE'overlay'QdataE(8'urlE'https%3A%2F%2Fraw.githubusercontent.com%2Fsehilyi%2Fgemini-datasets%2Fmaster%2Fdata%2Fcytogenetic_band.csv'%2C8'Vcsv'%2C8'chromosomeFieldE'Chr.'%2C8'genomicFieldsz8*'ISCN_start'H'ISCN_stop'HYart'HYop'8%5D%0A*)QmarkE'rect'Qtracksz8(8*J%22UqcolorE(8~ZE'Density'H*'Vnominal'H*'domainz'G25G50G75G100'%5DH*'rangez'whiteG%23D9D9D9G%23979797G%23636363Gblack'%5D8*Xext'HJ%22UqtextE(ZE'BandGVnominal')H'colorj'black')H'visibilityz8~(8~*'operationE'less-than'H~'measureE'width'H~'thresholdE'%7Cxe-x%7C'H~'transitionPaddingE10H~'targetE'mark'8~)8*%5D8)%2C8(8*JKgvarMA0A0F2'XriangleRight'HJKacen-1MB40101'XriangleLeft'HJKacen-2MB40101')8)%0A*%5DQxE(ZEYart%24xeE(ZEYop%24strokej'gray')QstrokeWidthj0.5)QwidthE1000QheightE20%0A)*%20%208%0A~E!%20G'%2C%20'H%2C8*J'dataTransformz8~(K'VfilterGfieldE'StainGoneOfz'M'%5D)8*%5DH'colorj'%23Q%2C%0A*'UE'Stain'H~'oneOfz'acen-1Gacen-2'%5DVtypeE'X)8)%2C8(8*'markE'tY'Basepair_stZ'fieldjE('valueEqH~'notEtrue8~)8*%5DH'zE%5B~**%228~*'Vfilter'H~Z%24GVgenomic')Q%01%24%22~zqjZYXVUQMKJHGE8*_">Example Editor</a>

Screenshot 2023-12-23 at 2 57 38 PM


So, we will need to check all the marks used in the sub-tracks as well, e.g.,

```ts
this.getResolvedTracks().reduce((acc, spec) => acc && stretchableMarks.includes(spec.mark), true))
```

sehilyi (commented):
Awesome, I think this is a great improvement! I left several comments, but otherwise the changes look good to me. Feel free to merge this to `main` after addressing them. I created an [example](https://github.com/gosling-lang/gosling.js/pull/1023) on Editor to compare the original/new rendering options – please merge it to this PR if it looks good to you.

I will also test this with tracks in Chromoscope since I think this approach of stretching graphics works well with several tracks in Chromoscope.

src/tracks/gosling-track/gosling-track.ts
etowahadams: That's a good idea!

src/tracks/gosling-track/gosling-track.ts
etowahadams: Ah this is because I was experimenting with some other Pixi stuff before. I'll add it back. Good catch!

src/tracks/gosling-track/gosling-track.ts
etowahadams: On first render, the graphicsXScale is equal to 1. We don't to stretch because nothing has been drawn yet. I just added a variable to better reflect this. Hopefully it is more clear now!

src/tracks/gosling-track/gosling-track.ts
etowahadams: I see that makes sense! I'll add that to `hasStretchableGraphics`

etowahadams:
These should resolve all the comments, thanks for the review. Will test the changes once the SSL cert is renewed to make sure everything looks good, and then will merge

etowahadams:
Caught an additional thing---we don't want stretching to occur by default then mouse interactions are enabled.

sehilyi:
Thanks for merging. I will bump the minor version of `gosling.js`

PR #1024, opened 2023-12-23
fix(core): publish events on first render

### LINKED ISSUE #1020: Events published during compiling are not captured due to rendering cycle
**Problem**
When using the gosling ref to subscribe to events of the API in a component wrapping the GoslingComponent, events published during compiling are not captured. 

**Minimal example**
[Minimal example](https://github.com/thomcsmits/alt-gosling-minimal/tree/thomcsmits/re-create-error) using [the same UseEffect as in the editor](https://github.com/gosling-lang/gosling.js/blob/ca315b16d8b977a92c5a0ed8a99c534d602c766f/editor/Editor.tsx#L313)

**Observed behavior**
Nothing is logged

**Expected behavior**
{specProcessed, object} is logged

**Explanation:** 
The useRef hook works by assigning the ref to a DOM node. This happens when the component is rendered, e.g. assigned to the root. Thus, the UseEffect hook (with the ref in it's dependency array) is called after the first render.
With Gosling the compiling happens before the component is rendered. Therefore, the events are already published before the ref is assigned (the subscribe function is called too late).

**Possible solution?** 
An initial render of an empty canvas, assigning the DOM node and then updating the component after it has compiled.

etowahadams:
Fix #1020
Toward #

## Change List
 - Compile a blank spec in the initial render so that the Gosling ref can get associated with the DOM. 

This enables subscription events to work properly. 

https://github.com/gosling-lang/gosling.js/assets/14843470/3d57b4ff-1919-426f-bdbf-83dcc06ae1ad

## Checklist

### DIFF
--- a/src/core/gosling-component.tsx
+++ b/src/core/gosling-component.tsx
@@ -100,7 +100,7 @@ export const GoslingComponent = forwardRef<GoslingRef, GoslingCompProps>((props,
     // TODO: add a `force` parameter since changing `linkingId` might not update vis
     const compile = useCallback(
         (altSpec?: gosling.GoslingSpec) => {
-                const spec = altSpec ?? props.spec;
+            const spec = altSpec ?? props.spec;
             if (spec) {
                 const valid = gosling.validateGoslingSpec(spec);
 
--- a/src/core/gosling-component.tsx
+++ b/src/core/gosling-component.tsx
@@ -174,7 +174,7 @@ export const GoslingComponent = forwardRef<GoslingRef, GoslingCompProps>((props,
                 wrapperSize.current.width !== newSize.width
             ) {
                 wrapperSize.current = newSize;
-                compile(props.spec);
+                compile();
             }
         });
 
@@ -188,7 +188,7 @@ export const GoslingComponent = forwardRef<GoslingRef, GoslingCompProps>((props,
                 wrapperParentSize.current.width !== newSize.width
             ) {
                 wrapperParentSize.current = newSize;
-                compile(props.spec);
+                compile();
             }
         });
 
@@ -208,7 +208,7 @@ export const GoslingComponent = forwardRef<GoslingRef, GoslingCompProps>((props,
             });
             setIsInitialRender(false);
         } else {
-            compile(props.spec);
+            compile();
         }
     }, [props.spec, theme, isInitialRender]);
 
--- a/src/core/gosling-component.tsx
+++ b/src/core/gosling-component.tsx
@@ -99,7 +99,8 @@ export const GoslingComponent = forwardRef<GoslingRef, GoslingCompProps>((props,
 
     // TODO: add a `force` parameter since changing `linkingId` might not update vis
     const compile = useCallback(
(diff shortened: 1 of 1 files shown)


etowahadams:
Here is what I used to test the subscription events (basically the example that Thomas provided, thanks Thomas!) 

```
import React, { useEffect, useRef, useState } from 'react';
import { GoslingComponent, type GoslingRef, type GoslingSpec } from 'gosling.js';

const goslingSpec = {
    title: 'Basic Marks: bar',
    subtitle: 'Tutorial Examples',
    tracks: [
        {
            layout: 'linear',
            width: 800,
            height: 180,
            data: {
                url: 'https://resgen.io/api/v1/tileset_info/?d=UvVPeLHuRDiYA3qwFlm7xQ',
                type: 'multivec',
                row: 'sample',
                column: 'position',
                value: 'peak',
                categories: ['sample 1'],
                binSize: 5
            },
            mark: 'bar',
            x: { field: 'start', type: 'genomic', axis: 'bottom' },
            xe: { field: 'end', type: 'genomic' },
            y: { field: 'peak', type: 'quantitative', axis: 'right' },
            size: { value: 5 }
        }
    ]
} as GoslingSpec;

function App() {
    const gosRef = useRef<GoslingRef>(null);

    useEffect(() => {
        if (gosRef.current) {
            const currentRef = gosRef.current;
            // specProcessed
            currentRef.api.subscribe('onNewView', (_, data) => {
                console.warn('onNewView', data);
            });
            currentRef.api.subscribe('onNewTrack', (_, data) => {
                console.warn('onNewTrack', data);
            });
            currentRef.api.subscribe('specProcessed', (_, data) => {
                console.warn('specProcessed', data);
            });
        }

    }, [gosRef.current]);

    return (
        <div>
            <GoslingComponent ref={gosRef} spec={goslingSpec} />
        </div>
    );
}

export default App;

```

src/core/gosling-component.tsx
```diff
@@ -96,63 +98,66 @@ export const GoslingComponent = forwardRef<GoslingRef, GoslingCompProps>((props,
     );
 
     // TODO: add a `force` parameter since changing `linkingId` might not update vis
-    const compile = useCallback(() => {
-        if (props.spec) {
-            const valid = gosling.validateGoslingSpec(props.spec);
+    const compile = useCallback(
+        (spec: gosling.GoslingSpec | undefined) => {
```
sehilyi: I see most of the time just the `props.spec` is used here. I think this would be more understandable if we just use the `props.spec` unless an alternative spec is passed.

```suggestion
        (altSpec?: gosling.GoslingSpec) => {
                const spec = altSpec ?? props.spec;
```

(used `?` instead of `| undefined` as we used it more frequently)

and then

```diff
- compile(props.spec);
+ compile();
```

sehilyi (approved):
Ah, I see how this is fixed – Great!

src/core/gosling-component.tsx
etowahadams: Just changed, thanks for the suggestion!

PR #1038, opened 2024-02-07
feat(core): export gosling svg logo

sehilyi:
Can we include the SVG element of the Gosling logo image in the gosling.js library? This would make it easier for people to add a Gosling logo to their applications (e.g., "Powered by" labels). Helpful for some of our applications in the lab as well since we do not need to copy and paste the SVG code.

```ts
import react from 'React';
import * gosling from 'gosling.js';

return (
   ...
   {gosling.getGoslingLogoSVG(20, 20)}
)
```

## Change List
 -

## Checklist


src/logo/index.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+
+export const getGoslingLogoSvg = (width = 20, height = 20) => (
```
manzt, line 3: This function returns JSX, so it requires React to use it (i.e., it is a React component). Rather than `getGoslingLogoSvg` why not export `GoslingLogo` component:

```tsx
type GoslingLogoProps = {
  width: number;
  height: number;
}

export function GoslingLogo({ width = 20, height = 20 }: GoslingLogoProps) {
  return <svg>...</svg>
}
```

I'd also include some JSDoc comments for `GoslingLogoProps` to explain what width and height are (pixels, fractional units, etc).

manzt (changes requested):
Some minor comments. I'd export this function (and use it) as a React Component.

manzt:
Another option would be to have a button in the Gosling editor /website to copy the icon in different ways (svg "raw", or JSX):

image

https://lucide.dev/icons/align-horizontal-distribute-end

I personally like this option because you can copy the SVG directly into your project and style it however you like (with or without react). With the React component, you define how styling needs to happen.

src/logo/index.tsx
manzt: If width and height should always be then same, then I'd go for just `size`.

manzt:
```jsx
import react from 'React';
import { GoslingLogo } from 'gosling.js';

function App() {
  return <GoslingLogo size={20} />
}
```

sehilyi:
Ah, I like the idea of adding a button to copy the logo! I am now leaning toward adding that to the editor or probably our website.

PR #1044, opened 2024-02-22
perf(track): Persist GoslingTrackModels

### LINKED ISSUE #1043: Synchronous navigation of tracks seems broken
I found the issue in the GIVE example on the Editor (https://gosling.js.org/?example=GIVE) and also in the Gosling Designer that uses the most recent Gosling version.

I can reproduce this by zooming in (e.g., seeing a few hundred BPs), panning, and then zooming out quickly.

https://github.com/gosling-lang/gosling.js/assets/9922882/5785e7fd-9dc4-4a0c-9b41-59390c593340

I see the following error message:

Screenshot 2024-02-22 at 8 55 24 AM

Copied texts:
```
index-fdb53256.js:1668 Uncaught RangeError: Maximum call stack size exceeded
    at new MJt (index-fdb53256.js:1668:1241331)
    at index-fdb53256.js:1668:1264625
    at Array.forEach (<anonymous>)
    at _e.transformDataAndCreateModels (index-fdb53256.js:1668:1263708)
    at index-fdb53256.js:1668:1259473
    at Array.flatMap (<anonymous>)
    at _e.processAllTiles (index-fdb53256.js:1668:1259456)
    at Ie (index-fdb53256.js:1668:1256496)
    at _e.draw (index-fdb53256.js:1668:1256723)
    at _e.zoomed (index-fdb53256.js:1668:1258755)
```

The maximum call stack error is coming from the following code:

https://github.com/gosling-lang/gosling.js/blob/9ecd4d30d1ae01a7ceb4edf4effe85a3eed2b619/src/tracks/gosling-track/gosling-track-model.ts#L82

etowahadams:
Fix #1043
Toward #

## Change List
 - Persist GoslingTrackModels between draw calls 
 - Switch from `structuredClone()` to `JSON.parse(JSON.stringify())`

https://github.com/gosling-lang/gosling.js/assets/14843470/7fd82e40-3fac-4814-9228-52c722011bb3

## Checklist

### DIFF
--- a/src/tracks/gosling-track/gosling-track.ts
+++ b/src/tracks/gosling-track/gosling-track.ts
@@ -174,6 +174,8 @@ const factory: PluginTrackFactory<Tile, GoslingTrackOptions> = (HGC, context, op
         #loadingText = new HGC.libraries.PIXI.Text('', loadingTextStyle);
         prevVisibleAndFetchedTiles?: Tile[];
         resolvedTracks: SingleTrack[] | undefined;
+        // This is used to persist processed tile data across draw() calls.
+        #processedTileMap: WeakMap<Tile, boolean> = new WeakMap();
 
         /* *
          *
@@ -416,6 +418,8 @@ const factory: PluginTrackFactory<Tile, GoslingTrackOptions> = (HGC, context, op
             this.getResolvedTracks(true); // force update
             this.clearMouseEventData();
             this.textsBeingUsed = 0;
+            // Without this, tracks with the same ID between specs will not be redrawn
+            this.#processedTileMap = new WeakMap();
 
             this.processAllTiles(true);
             this.draw();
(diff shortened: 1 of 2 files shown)


etowahadams:
Persisting GoslingTrackModels is something I originally tested in #1039. This offers substantial performance improvements since now the GoslingTrackModels don't have to be regenerated with every draw() call. 

It does seem that `JSON.parse(JSON.stringify())` is about 40% faster than `structuredClone()` in Firefox so I switched it back to that https://www.measurethat.net/Benchmarks/ShowResult/506888

src/tracks/gosling-track/gosling-track.ts
```diff
@@ -555,6 +559,11 @@ const factory: PluginTrackFactory<Tile, GoslingTrackOptions> = (HGC, context, op
             this.tileSize = this.tilesetInfo?.tile_size ?? 1024;
 
             const tiles = this.visibleAndFetchedTiles();
+            // If we have already processed all tiles, we don't need to do anything
+            // this.#processedTileMap contains all of data needed to draw
+            if (tiles.every(tile => this.#processedTileMap.get(tile) !== undefined)) {
+                return;
+            }
```
sehilyi, line 566: This is nice! I assume this also increases the performance to some extent.

sehilyi (approved):
Looks great! Thanks a lot for such a quick fix!

sehilyi:
`JSON.parse(JSON.stringify())` being faster than `structuredClone()` is surprising. I thought it was the opposite.

etowahadams:
Its worth noting that this reduces, but doesn't completely get rid of the freezing that sometimes occurs when zooming and panning. As far as I can tell, the reason for this is because of the HiGlass renderer, where an infinite loop can happen. The renderer detects that the scales have changed before the drawing has finished so it tries redrawing everything according to the new scales. But after it redraws, it sees that the scales have been updated again so it draws everything again. 

image

PR #1057, opened 2024-05-06
perf(track): reduce webgl memory usage

etowahadams:
Fix #
Toward #

Interacting with a visualization causes webgl memory to skyrocket. I use `about:memory` in Firefox to check memory usage. 
image
This was because `Graphics` objects were not being destroyed between draw calls, causing an accumulation of Graphics objects. With this fix, webgl textures memory only gets up to 10-15 MB. 

## Change List
 - Destroy the Graphics objects in `pBorder` before redrawing

## Checklist

### DIFF
--- a/src/tracks/gosling-track/gosling-track.ts
+++ b/src/tracks/gosling-track/gosling-track.ts
@@ -1,4 +1,4 @@
-import type * as PIXI from 'pixi.js';
+import * as PIXI from 'pixi.js';
 import { isEqual, sampleSize, uniqBy } from 'lodash-es';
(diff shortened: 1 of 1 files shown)


src/tracks/gosling-track/gosling-track.ts
```diff
@@ -1,4 +1,4 @@
-import type * as PIXI from 'pixi.js';
+import * as PIXI from 'pixi.js';
```
sehilyi: Maybe this was not the intended change?

etowahadams: Good catch!

PR #1074, opened 2024-06-07
feat(core): remove DataTrack from schema and compiler

### LINKED ISSUE #1065: Remove `DataTrack` or fully support it
For the context, where `DataTrack` is defined:

https://github.com/gosling-lang/gosling.js/blob/c3169dfa27df37c5ed7e954488d46f6d6eed075b/src/gosling-schema/gosling.schema.ts#L157

https://github.com/gosling-lang/gosling.js/blob/c3169dfa27df37c5ed7e954488d46f6d6eed075b/src/gosling-schema/gosling.schema.ts#L206-L211

The motivation of introducing Data Track was to enable users to quickly visualize data without thinking about visual encoding. Data Track is a partial type of a regular gosling track (`BasicSingleTrack`), which means users can specify the data type and its URL, and Gosling chose visual encoding based on the data type (e.g., bar charts for bigwig files).

Given the higher priority of supporting other important features than supporting/maintaining this, we made this feature deprecated at some point. However, we did not entirely remove this from the grammar.

We need to decide whether we want to remove this entirely from the grammar or support this better.

etowahadams:
Fix #1065
Toward #

## Change List
 - Removes DataTrack from schema 
 - Removes DataTrack related processing from compiler 

## Checklist


src/compiler/compile.ts
```diff
@@ -33,9 +33,6 @@ export function compile(
     // Make sure to keep the original spec as-is
     const specCopy = JSON.parse(JSON.stringify(spec));
 
-    // Override default visual encoding (i.e., `DataTrack` => `BasicSingleTrack`)
-    overrideDataTemplates(specCopy);
-
```
etowahadams, line 38: This function adds default visual encodings to DataTrack. Since there is no more DataTrack, we can remove this function

sehilyi (approved):
This looks good after merging https://github.com/gosling-lang/gosling.js/pull/1079 to this PR.

PR #1077, opened 2024-06-10
refactor(core): change _processedSpec to experimental_processedSpec

### LINKED ISSUE #1068: Revisit the callback function of `compiled` for naming/appropriateness
https://github.com/gosling-lang/gosling.js/blob/c3169dfa27df37c5ed7e954488d46f6d6eed075b/src/core/gosling-component.tsx#L29

This is added to be used in AltGosling (https://github.com/gosling-lang/altgosling/blob/8fbd4fffac1eca5259de7d8aab4689638a321d90/src/AltGoslingComponent.tsx#L315).

I think we need to review the name of the object ("_additionalData" and  "_processedSpec") and also find out if this is the best way to support such internal-usage data.

etowahadams:
Fix #1068
Toward #

## Change List
 - Change `_processedSpec` in compile callback function to `experimental_processedSpec`

## Checklist

### DIFF
--- a/src/core/gosling-component.tsx
+++ b/src/core/gosling-component.tsx
@@ -26,7 +26,7 @@ export interface UrlToFetchOptions {
 type CompiledCallbackFn = (
     goslingSpec: gosling.GoslingSpec,
     higlassSpec: gosling.HiGlassSpec,
-    _additionalData: { _processedSpec: gosling.GoslingSpec }
+    experimental: { experimental_processedSpec: gosling.GoslingSpec }
 ) => void;
 
 export interface GoslingCompProps {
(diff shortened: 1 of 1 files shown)


etowahadams:
In our discussion last week it seems like we settled on a new naming convention:
- For internally used names: Use underscores (for example`_header`), 
- For externally used names: prefix with `experimental`, or wrap with `experimental` 

Following this convention, I renamed `_specProcessed` to `experimental_specProcessed`. This is meant to be used externally, but it is unstable

src/core/gosling-component.tsx
```diff
@@ -26,7 +26,7 @@ export interface UrlToFetchOptions {
 type CompiledCallbackFn = (
     goslingSpec: gosling.GoslingSpec,
     higlassSpec: gosling.HiGlassSpec,
-    _additionalData: { _processedSpec: gosling.GoslingSpec }
+    experimental: { experimental_processedSpec: gosling.GoslingSpec }
```
sehilyi: `experimental` looks duplicated to me. Since this object (processed spec) is already wrapped with `experimental`, I think we can just use underscore to denote that it is both experimental and internal.

```suggestion
    experimental: { _processedSpec: gosling.GoslingSpec }
```

sehilyi: We can clarify this further using JS Doc that we discussed during the last meeting.

```suggestion
    /** @deprecated This is an experimental object that is not intended for production usage. */
    experimental: { _processedSpec: gosling.GoslingSpec }
```

sehilyi:
cc'ing @thomcsmits for heads up

PR #1094, opened 2025-01-09
feat: React-less and HiGlass-less Gosling.js

sehilyi:
This is a PR of @etowahadams's fork that tries to get rid of HiGlass and React dependencies. Not yet ready to merge.

- Etowah's comment of things left 
> - More extensive testing of the new renderer. It seems to work well for the examples in the editor, but I'm sure we'll find more bugs. We have ways to comprehensively test certain aspects of the spec but not all.
> - React API: I [implemented](https://github.com/etowahadams/gosling.js/blob/main/demo/GoslingComponent.tsx) a basic GoslingComponent but it doesn't have all of the features of the current component. I don't think this will be too difficult though.
- Etowah's note on potential improvements [explanation.md](https://github.com/etowahadams/gosling.js/blob/6b62a32955f16d2f911ce92fc8f107eca1c450e0/explanation.md)

### Fixing TS/build errors
- [x] `pnpm build` gives a bunch of TypeScript errors (Found 43 errors in 13 files as of [8b3cea](https://github.com/gosling-lang/gosling.js/pull/1094/commits/8b3ceacc0603b162bf1825c654dcef0368f63383))
  - [ ] I used several `@ts-expect-error` tags around d3 types.
  - [ ] There seems to be many HiGlass-related types that need to be cleaned (e.g., some types are still left `any`, such as `pubSub`, and there are methods/variables defined in HiGlass but not used in Gosling)
- [x] `pnpm test` gives 7 fails in 2 files → Fixed in [#1097](https://github.com/gosling-lang/gosling.js/pull/1097)
- [x] `pnpm run format` gives 74 problems (29 errors, 45 warnings) → Fixed in [#1099](https://github.com/gosling-lang/gosling.js/pull/1099)
  - [x] Include `/demo` to eslint scope and fix lint errors
### Implement original features
- [ ] GoslingComponents' original features need to be fully implemented (https://github.com/gosling-lang/gosling.js/pull/1094#discussion_r1913656995).
  - [ ] `idTable` that we supported for zoom API functions needs to be newly implemented.
- [x] This PR does not support `urlToFetchOptions` ([original PR](https://github.com/gosling-lang/gosling.js/pull/965)) anymore, which should be newly implemented. (to be fixed in https://github.com/gosling-lang/gosling.js/pull/1105)
- [x] BAM file won't load (`bam-data-fetcher.ts:55 Uncaught (in promise) ReferenceError: Buffer is not defined`). Probably, this should work in the preview (`pnpm preview`).
- [ ] TrackTemplates does not work, resulting in the empty canvas (`Track Templates: Genes, Sequence, and Ideograms` example in the Editor)
- [ ] Showing the mouse position does not seem to work anymore?
### Fix discrepancy
- [x] Zoom speed has been changed when using mouse wheels (much faster when tested in this PR) → Fixed in https://github.com/gosling-lang/gosling.js/pull/1094/commits/c58c7d9fb59055a5f649eaf46908fe7e2585d479

### Packaging
- [ ] Make two packages? (gosling core and component). How to manage that, e.g., separate repo vs monorepo?
- [ ] Move track-def, renderer, and linking to `src`. Currently they are under `demo`


editor/Editor.tsx
```diff
@@ -1211,19 +1212,7 @@ function Editor(props: RouteComponentProps) {
...
-                                            padding={60}
-                                            margin={0}
-                                            border={'none'}
-                                            id={'goslig-component-root'}
-                                            className={'goslig-component'}
-                                            compiled={(_, h) => {
-                                                setHg(h);
-                                            }}
-                                        />
+                                        <GoslingComponent spec={goslingSpec} width={1000} height={2000} />
```
sehilyi, line 1193: Additional implementation needed here for removed options.

package.json
```diff
@@ -116,7 +125,6 @@
     "c8": "^7.11.2",
     "conventional-changelog-cli": "^2.1.1",
     "d3-drag": "^2.0.0",
```
sehilyi: d3-drag is included in the normal dependency, so remove from the dev dependencies?

src/index.ts
```diff
@@ -8,9 +8,9 @@ export { GoslingTemplates } from './core/utils/template';
...
 
-export { init } from './core/init';
+// export { init } from './core/init';
 export { compile } from './compiler/compile';
 export { validateGoslingSpec } from '@gosling-lang/gosling-schema';
-export { GoslingComponent } from './core/gosling-component';
+// export { GoslingComponent } from './core/gosling-component';
 export type { GoslingRef } from './core/gosling-component';
-export { embed } from './core/gosling-embed';
+// export { embed } from './core/gosling-embed';
```
sehilyi: Should be reverted back

src/missing-types.d.ts
```diff
@@ -484,6 +481,10 @@ declare module '@higlass/tracks' {
         abstract mouseMoveZoomHandler(absX?: number, abxY?: number): void;
     }
 
+    export abstract class HeatmapTiledPixiTrack<Options> extends Tiled1DPixiTrack<TileData, Options> {
+        // TODO: fill this out
```
sehilyi, line 486: TS types to be added here

dvdkouril:
I'll try to tackle at least some of the typescript errors we get at `pnpm build`. 
Starting point: `Found 120 errors in 25 files.`

Edit: some low-hanging fruit fixes bring this to `Found 44 errors in 14 files.`, but most of the remaining errors might need some more thinking...

src/compiler/compile.ts
```diff
@@ -2,33 +2,34 @@ import type { GoslingSpec, TemplateTrackDef, VisUnitApiData } from '@gosling-lan
...
-    callback: CompileCallback,
     templates: TemplateTrackDef[],
     theme: Required<CompleteThemeDeep>,
     containerStatus: {
         containerSize?: { width: number; height: number };
         containerParentSize?: { width: number; height: number };
     },
     urlToFetchOptions?: UrlToFetchOptions
-) {
+): CompileResult {
```
sehilyi: This change makes sense (i.e., returning desired objects instead of passing over the callback function) but is not consistently reflected on other parts of code (e.g., several test fails come from this). I will stick to this change and address other parts.

dvdkouril: Great! I was slightly confused about the changes to `compile` function and didn't want to touch it in case the callback approach is needed for some use case.

demo/track-def/types.ts
```diff
@@ -0,0 +1,63 @@
...
+};
+
+export type ProcessedCircularTrack = ProcessedTrackBase & {
+    id: string;
+    layout: 'circular';
+    startAngle: number;
+    endAngle: number;
+    outerRadius: number;
+    innerRadius: number;
+};
```
sehilyi: I like this change. I will see how I could address the TS errors that are occurred by this change.

dvdkouril:
I just realized one thing:
in `demo/GoslingComponent.tsx`:
```typescript
function renderGosling(/* ... */) {
    const compileResult = compile(gs, [], getTheme('light'), { containerSize: { width: 0, height: 0 } });
    const { trackInfos, gs: processedSpec, theme } = compileResult;
    /* ... */
}
```
the `compile` function calls this function:
```typescript
// Make HiGlass models for individual tracks
const compileResult = createHiGlassModels(specCopy, trackInfos, theme, urlToFetchOptions);
```
But the higlass part of `compileResult` is ignored above! I was pretty confused today, cause I found myself in the part of the code that creates the custom Higlass tracks (views?) and was wondering how is that getting rid of the Higlass dependency.

The [explanation in this diagram](https://github.com/etowahadams/gosling.js/blob/main/explanation.md) was super helpful to realize this. I hope I'm not mistaken.

So I guess besides just documenting my confusion from today, the insight for the PR itself is that there's a lot more fat that needs trimming in the code!

sehilyi:
Ah, that's good to know. I didn't realize that those HiGlass-related parts are largely unused. I can try to see how I can cut the fat, and it may make the compiling process simpler.

sehilyi:
closing since we now have another dedicated branch, i.e., `v2`

PR #1099, opened 2025-01-27
chore: upgrade lint/prettier for new TS version

sehilyi:
Upgrading eslint and prettier packages since we need to use new TS versions. This addresses an error message received when running `pnpm format`:

```
=============

WARNING: You are currently running a version of TypeScript which is not officially supported by @typescript-eslint/typescript-estree.

You may find that it works just fine, or you may not.

SUPPORTED TYPESCRIPT VERSIONS: >=3.3.1 <5.2.0

YOUR TYPESCRIPT VERSION: 5.7.3

Please only submit bug reports when using the officially supported version.

=============
```

- [x] I get 52 problems (52 errors, 2 warnings) after running `pnpm format`


dvdkouril:
I probably should have made the typescript version bump in a separate PR, too...Sorry.

sehilyi:
@dvdkouril No worries! We had to upgrade TS anyway.

I am working on separate PRs since it I think makes it little more easier for me to keep track of what I changed for which issues.

.eslintignore
sehilyi, line 1: This file is not supported by the latest eslint.

sehilyi:
CI Test is failing mainly due to build errors, which we still need to fix in a separate PR.

PR #1100, opened 2025-01-29
Restructuring the compile process

sehilyi:
Still working on editing types to reflect on the newly introduced type, i.e., ProcessedTrack. This may require huge changes to the codebase.


sehilyi:
This is going little far than I expected.

sehilyi:
I will work on restructuring the compile process in a separate PR.

PR #1102, opened 2025-02-03
fix(core): typescript errors when building

sehilyi:
Screenshot 2025-02-04 at 9 51 07 AM

Several tasks that I want to address in a follow-up PR.

- [ ] I used several `@ts-expect-error` tags around d3 types.
- [ ] There seems to be many HiGlass-related types that need to be cleaned (e.g., some types are still left `any`, such as `pubSub`, and there are methods/variables defined in HiGlass but not used in Gosling)


sehilyi:
@dvdkouril I will merge this to the `higlassless` branch and then probably work on [cutting fat related to unused HiGlass](https://github.com/gosling-lang/gosling.js/pull/1094#issuecomment-2638113189). I think it may reduce the complexity of working on the `higlassless` branch since we don't have to edit code that we will be removed at the end.

PR #1104, opened 2025-02-07
fix: add valid esm.sh path for higlass-text module

### LINKED ISSUE #1103: Unable to export html: higlass-text@%5E0.1.1:1 Failed to load resource: the server responded with a status of 404 (Not Found)
I am trying to export an example from [Gosling.js](https://gosling.js.org/), but when I try to open the saved HTML file, I get the following error in the developer console:

gosling.js@0.17.0:14 
       GET https://esm.sh/higlass-text@%5E0.1.1?target=es2022 net::ERR_ABORTED 404 (Not Found)

When I try to open [this URL](https://esm.sh/higlass-text@%5E0.1.1?target=es2022) directly, it shows that the module is not found.

dvdkouril:
Temporary fix for the HTML files produced by HTML export in Gosling editor not working due to an esm.sh URL that no longer finds the correct (ESM) module. Probably would be better to fix on higlass-text level, but in the meantime, this PR at least makes it so that the produced HTML files continue working.

Some more context: https://github.com/gosling-lang/gosling.js/issues/1103#issuecomment-2643447163

Fix #1103
Toward #

## Change List
 -

## Checklist

### DIFF
--- a/editor/html-template.ts
+++ b/editor/html-template.ts
@@ -19,7 +19,8 @@ export const getHtmlTemplate = (
           "react-dom": "https://esm.sh/react-dom@${reactVersion}",
           "pixi": "https://esm.sh/pixi.js@${pixiVersion}",
           "higlass": "https://esm.sh/higlass@${higlassVersion}?external=react,react-dom,pixi",
-          "gosling.js": "https://esm.sh/gosling.js@${goslingVersion}?external=react,react-dom,pixi,higlass"
+          "higlass-text": "https://esm.sh/higlass-text/es/index.js",
+          "gosling.js": "https://esm.sh/gosling.js@${goslingVersion}?external=react,react-dom,pixi,higlass,higlass-text"
         }
       }
     </script>


manzt (approved):
This is a good temporary fix. I think we should shoot to fix the upstream package with Etowah's PR. I don't see any push back there, so I say we should merge. I'll defer to Alex first, but can also merge myself if we don't hear back.

dvdkouril:
Another thing is that the package doesn't seem to be used in the new Gosling version.

manzt:
Is it not?

https://github.com/gosling-lang/gosling.js/blob/e66dc34543c502f3547e37f358f00998ea8b21d3/src/core/init.ts#L4

dvdkouril:
Hm, but the `init()` function isn't actually used anywhere, is it? I'm probably missing something, but my impression was that the goal of the new version was to not have to do `higlassRegister` but render the tracks with pixi directly in Gosling.

Anyway, that's a conversation for elsewhere. Thanks for the review!

manzt:
Sorry, I'm a bit out of the loop with latest development, but I think it is?

```sh
~/github/gosling-lang on main
❯ rg 'init\(\)'
gosling.js/src/core/higlass-component-wrapper.tsx
14:gosling.init();

gosling.js/src/core/init.ts
16:export function init() {
```

https://github.com/gosling-lang/gosling.js/blob/e66dc34543c502f3547e37f358f00998ea8b21d3/src/core/higlass-component-wrapper.tsx#L11-L14

dvdkouril:
Ah, I see, sorry.

sehilyi:
Yes, `init()` is currently used but is removed in our PR for HiGlass-less Gosling (https://github.com/gosling-lang/gosling.js/pull/1094).

I will merge this. Thank you for the PR and the review!

PR #1107, opened 2025-02-11
chore(core): sync changes to the `spatial-layout` branch

sehilyi:
Note that the base branch of this PR is `dvdkouril/spatial-layout`.

This PR merges recent changes from the `higlassless` branch to `dvdkouril/spatial-layout`.


sehilyi:
I think it would be good to constantly sync the two branches so that their differences do not become too complex/large.

dvdkouril:
Yeah, I agree that it'd be great to keep these two branches in sync!

I'm not going through all the changes here but I think it's safe to merge.

dvdkouril (approved):

PR #1109, opened 2025-02-12
fix(core): remove build errors

sehilyi:
This PR reduces the number of build errors from `16` to `0`. I mainly added several common track-level properties (e.g., `layout`) in `ChromospaceTrack` to walk around the errors. Not the ultimate solution, but it enables building the project.

I think reducing the complexity of the Gosling compiler will make it easier for us to define only the properties needed in `ChromospaceTrack`. I tried to simplify the compiler in https://github.com/gosling-lang/gosling.js/pull/1100 and will work on it again.


dvdkouril (approved):
I see, thanks! I'm still orienting myself and figuring out where the "switch" to spatial rendering should happen.

sehilyi:
I saw there is an if statement to determine whether to render a chromospace visualization ([code](https://github.com/gosling-lang/gosling.js/blob/e803b94b7788d30c1fd41255bd1913ed1b5f8bd9/demo/renderer/main.ts#L109-L145)). Are you referring to something different?

dvdkouril:
Yes! And there's also the part where you actually make it `TrackType.Spatial` and that's something I'm still figuring out/working on.

PR #1114, opened 2025-02-20
refactor: make normalization steps as the first step in the compiler

sehilyi:


sehilyi:
Will create another PR after making changes from `v2`

PR #1126, opened 2025-03-08
chore(core): Upgrade vitest/vite and patch @gmod/bam

manzt:
These deps are quite old and upgrading substantially improves bundle size and test time.

This PR also uses [`pnpm patch <pkg>`](https://pnpm.io/cli/patch) to apply a
patch to `@gmod/bam` to import `buffer`. Ideally, I think we could just upgrade
@gmod/bam to the latest version but that should be in a separate PR.

## Checklist

### DIFF
--- a/patches/@gmod__bam@1.1.18.patch
+++ b/patches/@gmod__bam@1.1.18.patch
@@ -1,0 +1,12 @@
+diff --git a/CHANGELOG.md b/CHANGELOG.md
+deleted file mode 100644
+index 7df9b5857ac626b6c8806ede2c0c93a3658d7d2a..0000000000000000000000000000000000000000
+diff --git a/esm/bamFile.js b/esm/bamFile.js
+index 4fa55dd7bd83eb97314f857af61ea0f149f7396b..0aa9579ea442e6301bd0b2191c6d59f450799247 100644
+--- a/esm/bamFile.js
++++ b/esm/bamFile.js
+@@ -1,3 +1,4 @@
++import { Buffer } from "buffer";
+ import crc32 from 'buffer-crc32';
+ import { unzip, unzipChunkSlice } from '@gmod/bgzf-filehandle';
+ import entries from 'object.entries-ponyfill';
--- a/vite.config.js
+++ b/vite.config.js
@@ -1,68 +1,10 @@
 import { defineConfig } from 'vite';
 import react from '@vitejs/plugin-react';
-import * as esbuild from 'esbuild';
 import path, { dirname } from 'path';
 import { fileURLToPath } from 'url';
 import pkg from './package.json' assert { type: 'json' }; // must do the assert to solve https://nodejs.org/api/errors.html#err_import_assertion_type_missing
 
 const __dirname = fileURLToPath(dirname(import.meta.url));
-
-/**
- * Bundles vite worker modules during development into single scripts.
- * Also injects node globals for gmod libraries.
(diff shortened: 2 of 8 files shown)


patches/@gmod__bam@1.1.18.patch
manzt, line 1: patches are automatically applied when running `pnpm install`.

vite.config.js
```diff
@@ -1,68 +1,29 @@
...
-  const objURL = blob && (window.URL || window.webkitURL).createObjectURL(blob);
-  try {
-    return objURL ? new Worker(objURL) : new Worker("data:application/javascript;base64," + encodedJs, {type: "module"});
-  } finally {
-    objURL && (window.URL || window.webkitURL).revokeObjectURL(objURL);
-  }
-}`;
-        }
-    }
-};
```
manzt, line 65: I'm suspect that this has not been working since upgrading vite previously. A better solution is to apply patches to dependencies specifically with pnpm.

manzt:
I think this may still need some testing to ensure the workerized loaders are working for dev and build. I was able to confirm things working for BAM.

vite.config.js
```diff
@@ -1,68 +1,29 @@
...
-
-// audit libraries to see if they use `Buffer` in a way that needs to be shimmed
+function determineExternalDependencies(pkg, options) {
+    // Packages that we have modified cannot be external because package consumers will not
+    // get our modifications. Ideally, we can avoid modifying these packages ourselves.
+    const patchedPackages = Object.keys(pkg.pnpm.patchedDependencies).map(
+        specifier => specifier.split('@').slice(0, specifier.startsWith('@') ? 2 : 1).join('@')
+    );
+    return [
+        ...Object.keys(pkg.dependencies),
```
sehilyi, line 20: Is it intended that `pkg.dependencies` is included here? I thought any packages specified as `external` are the ones that users need to install on their own.

sehilyi (approved):
Thank you for this patch! Tested locally for BAM, VCF, and GFF and can confirm that they are working both for dev and production.

vite.config.js
manzt: Yes. When someone installs `gosling.js` (e.g., `npm install gosling.js`), the command also resolves and installs _all_ packages listed in gosling's `dependencies`. Keeping a package's dependencies "external" means that the final build will include the original specifier, eg an import like:

```js
import { debounce } from "lodash";
```

stays as

```js
import { debounce } from "lodash";
```

in `gosling.mjs`.

This means that a consumer of the package (e.g., bundler) will resolve the identifier to the package installed from gosling's `dependencies`. Preserving the original import gives the consumer several benefits:

- Deduplication. An app using gosling cans share a dependency with gosling and the package will only end up in the bundle once. If that dep was included in gosling (ie not external) it would end up in the output twice.
- Upgrade dependencies. If a dependency in gosling has a patch fix or security fix and is external, end users are free to upgrade deps to the next compatible semver and get the fix. If not, Gosling needs to upgrade and make a new release.
- Module resolution. Similar to how we modify packages, consumers of gosling could make modifications or aliases for imports if they so desire. If a package is bundled, you can't do that.

Hopefully that makes sense. Ideally for a package, all dependencies should be external. In our case, we can't do that for Gosling since we modify some of the dependencies and those modifications aren't published to npm. Thus a package consumer wouldn't get our changes for those dependencies.

sehilyi: Ah, this is very helpful, and it makes a lot of sense. Thank you for the write-up!

PR #1127, opened 2025-03-17
chore(editor): spatial examples are the default demo in the editor

sehilyi:
Created a dedicated example group ('Spatial Layout'). Examples belonging to this group will be shown at the top of the example gallery. The default example of the editor can be set by the `forceShow` property.

```ts
// example list
SPATIAL_LAYOUT: {
    group: 'Spatial Layout',
    name: 'Integrating Spatial Models of Chromatin',
    spec: JsonExampleSpecs.EX_SPEC_SPATIAL,
    image: THUMBNAILS.SPATIAL_DRAFT,
    forceShow: true // default example
},

```

Screenshot 2025-03-17 at 10 14 56 AM


sehilyi:
The lint/test fails are from the `spatial` branch. These can be addressed in a separate PR.

PR #1138, opened 2025-04-02
feat: add capabilities for `spatial` layout type

sehilyi:
This PR implements the ability to specify a view to use a `spatial` layout. This allows the user to supply a "model" file that associates genomic regions with xyz coordinates in space.

Prototypical version of the editor with these capabilities is deployed at https://3d.gosling-lang.org

image

## basic use
```json
{
  "views": [
    {
      "layout": {
        "type": "spatial",
        "model": {
          "type": "csv",
          "url": "https://.../yeast_model.csv",
          "xyz": ["x", "y", "z"],
          "chromosome": "chr",
          "position": "coord"
        }
      },
      "tracks": [
        { /* ... */ }
      ]
    }
  ]
}
```

## TODO (MVP)
_When these are resolved, we should be able to merge_
- [x] switch to using uchimata (+ use the new unified API)
    - [ ] fix: filtering only works on first refresh, not when switching to subsequent examples in the editor UI
- [x] fix types (`pnpm build-types` should run without issues)
- [ ] Better solution for the hard-coded `data.sampleLength = 30000` (in `demo/renderer/main.ts`, line ~163)

## TODO (extras)
_Document here but likely work on in separate PRs after merging this one_
- [ ] simple linking of views via synchronized camera

## Change List
 - adds two new dependencies: [uchimata](https://github.com/hms-dbmi/uchimata), to facilitate rendering of 3D data; and [flechette](https://github.com/uwdata/flechette) for working with Arrow IPC format.

## Checklist


src/gosling-schema/gosling.schema.ts
```diff
@@ -80,12 +80,22 @@ export type ResponsiveSpecOfMultipleViews = {
     }[];
 };
 
-export type Layout = 'linear' | 'circular';
+export type Layout = 'linear' | 'circular' | 'spatial' | LayoutSpatial;
```
dvdkouril, line 83: I changed this from previous suggestion to have:
```javascript
export type LayoutDeep =
    | { type: 'linear' | 'circular' }
    | { type: 'spatial';
        model: {
            type: 'csv';
            url: string;
            xyz: [string, string, string];
            chromosome: string;
            position: string;
        };
    };
```
Main reason is that there's no reasonable situation in which you'd want to define a linear or circular layout with `layout: { type: 'linear' }` instead of doing simply `layout: 'linear'`. This makes the spatial layout object special (nothing bad with that, IMHO), and also simplifies the method that pushes this view-level layout specification down to tracks (in `_normalize-3d-spec.ts`).

dvdkouril:
@sehilyi Since you opened this PR, I can't assign you as reviewer. Could you take a look if there's anything critical that would prevent from merging this so that we can start using it in GD? 

It's not perfect, but the CI checks are green, and we can iron out the other issues while working on the GD UI integration, since this will still end up in the alpha version of Gosling.

sehilyi:
Thanks for letting me know - I somehow did not notice that the PR is ready to review. I will review this now and let you know here.

editor/example/json-spec/3d-yeast-model.ts
```diff
@@ -0,0 +1,127 @@
...
+                    height: trackWidth
+                }
+            ]
+        },
+        {
+            arrangement: 'vertical',
+            views: [
+                {
+                    layout: 'linear',
+                    tracks: [{ ...chrTrack }, { title: 'Linear Layout', ...base }]
```
sehilyi: It looks like the `locus` is now only used for the `spatial` layout, which I think is okay (or probably better). This change should fix the example.

```suggestion
                    tracks: [{ ...chrTrack }, { title: 'Linear Layout', ...base, locus: undefined, x: { field: 'position', type: 'genomic' } }]
```

editor/example/json-spec/3d-yeast-model.ts
```diff
@@ -0,0 +1,127 @@
...
+                },
+                {
+                    layout: 'circular',
+                    centerRadius: 0.5,
+                    spacing: 1,
+                    tracks: [{ ...chrTrack }, { title: 'Cicular Layout', ...base }]
```
sehilyi:
```suggestion
                    tracks: [{ ...chrTrack }, { title: 'Cicular Layout', ...base, locus: undefined, x: { field: 'position', type: 'genomic' } }]
```

src/gosling-schema/gosling.schema.ts
```diff
@@ -436,6 +451,17 @@ interface SingleTrackBase extends CommonTrackDef {
...
+    //  The reason to have this in the SingleTrackBase instead of making a different (SpatialTrack) is that a spatial track
+    //  should still be a "single track". Not specializing makes it easier for flowing between the different procedures of spec transformations.
+    spatial?: {
+        x: string;
+        y: string;
+        z: string;
+        chr: string;
+        coord: string;
+    };
+
```
sehilyi, line 464: It makes sense to have this here, especially given the tight timeline.

But, given that we now have more layout options, and some of them have specialized properties (circular layouts also have some properties, like `innerRadius`), I think we should consider having more scoped types **in the future**. For example, several circular-specific properties are defined in `CommonTrackDef` that will also need to be moved out, in addition to `locus` and `spatial`, e.g.,

```ts
type SingleTrackBase = SingleTrackBaseLinear | SingleTrackBaseCircular | SingleTrackBaseSpatial;
```

sehilyi (commented):
@dvdkouril I tested locally, and I can confirm that this works in VEC. Thanks for making this change - I'm excited to include this in our alpha version.

Please feel free to merge this when you feel ready. Also, after you merged it, can you try releasing a new alpha version based on the description I wrote on [v2/CONTRIBUTING.md](https://github.com/gosling-lang/gosling.js/blob/v2/CONTRIBUTING.md#alpha-or-beta-versions)?

src/gosling-schema/gosling.schema.ts
dvdkouril: Yea, I definitely took the easy way right now. Will be happy to revisit the types when there's some more time.

PR #1136, opened 2025-03-28
feat: grammar extension for 3D layout

sehilyi:
teaser

```ts
{
  "title": "Spatial Layout: Grammar Extension",
  "assembly": [
    ["chrI", 230218],
    ["chrII", 813184],
    ["chrIII", 316620],
    ["chrIV", 1531933],
    ["chrV", 576874],
    ["chrVI", 270161],
    ["chrVII", 1090940],
    ["chrVIII", 562643],
    ["chrIX", 439888],
    ["chrX", 745751],
    ["chrXI", 666816],
    ["chrXII", 1078177],
    ["chrXIII", 924431],
    ["chrXIV", 784333],
    ["chrXV", 1091291],
    ["chrXVI", 948066]
  ],
  "views": [
    {
      "layout": {
        "type": "spatial",
        "model": {
          "type": "csv",
          "url": "https://gosling-lang.s3.us-east-1.amazonaws.com/data/3d/yeast.csv",
          "xyz": ["x", "y", "z"],
          "chromosome": "chr",
          "position": "coord"
        }
      },
      "tracks": [
        {
          "mark": "point",
          "data": {
            "type": "bigwig",
            "url": "https://gosling-lang.s3.us-east-1.amazonaws.com/data/3d/GSM5241662_y1483_plus.bw"
          },
          "locus": {"field": "position"},
          "color": {"field": "value", "type": "quantitative"},
          "size": {"field": "value", "type": "quantitative"},
          "width": 500,
          "height": 500
        }
      ]
    }
  ]
}
```

PR #1139, opened 2025-04-02
chore: move the 3d examples to the top; use the new grammar in example specs

sehilyi:
Somehow, the 3d examples are moved to the bottom of the gallery recently. I moved this examples back to the top in this PR. I also edited example specs so that they use the new grammar introduced in https://github.com/gosling-lang/gosling.js/pull/1136.

# Before

Screenshot 2025-04-02 at 2 45 31 PM

# After

Screenshot 2025-04-02 at 2 48 02 PM


src/core/utils/chrom-size.ts
```diff
@@ -187,6 +187,25 @@ export const CHROM_SIZE_MM9 = {
...
+    chrIX: 439888,
+    chrX: 745751,
+    chrXI: 666816,
+    chrXII: 1078177,
+    chrXIII: 924431,
+    chrXIV: 784333,
+    chrXV: 1091291,
+    chrXVI: 948066
+};
+
```
sehilyi, line 208: The Yeast assembly is added to Gosling, just so that we do not need to define the chromosome names and sizes in example specs. Now, we can use `assembly: 'sacCer3'`.

editor/example/index.ts
```diff
@@ -443,17 +447,10 @@ export const editorExampleObj: {
         spec: JsonExampleSpecs.EX_SPEC_SPATIAL_MULTIPLE_TRACKS,
         image: THUMBNAILS.SPATIAL_DRAFT
     },
-    SPATIAL_LAYOUT_GRAMMAR_EXTENSION: {
-        group: 'Experimental',
-        name: 'Spatial Layout: Grammar Extension',
-        spec: JsonExampleSpecs.EX_SPEC_SPATIAL_LAYOUT_GRAMMAR_EXTENSION,
-        image: THUMBNAILS.SPATIAL_DRAFT,
-        forceShow: true
-    },
```
sehilyi, line 452: Removed this example I previously created since it is very similar to another one (`SPATIAL_LAYOUT_3D_YEAST_MODEL`).

editor/example/json-spec/spatial-layout.ts
```diff
@@ -133,20 +123,22 @@ export const EX_SPEC_SPATIAL_MULTIPLE_TRACKS: GoslingSpec = {
...
                         type: 'csv',
                         url: 'https://raw.githubusercontent.com/dvdkouril/chromospace-sample-data/main/gosling-3d/yeast_model.csv'
                     },
-                    spatial: {
-                        x: 'x',
-                        y: 'y',
-                        z: 'z',
-                        chr: 'chr',
-                        coord: 'coord'
-                    },
```
sehilyi, line 149: The changes I made in examples are basically removing `"spatial"` and adding `"layout: { type: 'spatial', model: { ...}, ... }`, reflecting the grammar extension. But, using the previous approach, i.e., "spatial", should also work as well (i.e., Gosling will render the same thing).

PR #1140, opened 2025-04-02
feat(editor): add a 3D yeast model example displayed in three layout options

sehilyi:
The 3D yeast model and genomic data displayed in three layouts. The same bigwig file is used in all views. The colors in the second 3D view represent chromosomes.

## Screenshot

Screenshot 2025-04-02 at 4 17 02 PM

## Data Source
The bigwig file was taken from the following source:

- Paper: https://wiki.yeastgenome.org/images/f/f0/Ng_2019_PMID_31612944.pdf
- Data Repo: https://www.ncbi.nlm.nih.gov/geo/query/acc.cgi?acc=GSM5241662

The data description found from the paper:

> For WT cells grown in glucose media (ypd), the amount of transcripts covering each
position on the plus strand is represented in this track.


src/tracks/spatial-track/spatial-track.ts
```diff
@@ -161,7 +161,7 @@ function handleColorField(color?: ChannelValue | Color | string, arrowIpc: Uint8
         if (color.type === 'nominal') {
             console.warn('not implemented!');
             const values = fetchValuesFromColumn(color.field, arrowIpc) as string[]; //~TODO: forcing to string[] not good
-            const colScale = randomColors(50); //~ just some big number
+            const colScale = color.range ?? randomColors(50); //~ just some big number
```
sehilyi, line 164: I added this to respect user-defined colors (`color: { ..., range: ['black' ,...] }`), if exist. Is this okay?

dvdkouril: Yes! I did exactly this in a separate branch right before the deadline: https://github.com/gosling-lang/gosling.js/blob/01d2a5f6f3dec1e0641cafd11b9ff2698bd14819/src/tracks/spatial-track/spatial-track.ts#L158

dvdkouril:
Is CORS for the .bw file set for certain origins only? 
image

sehilyi:
It should work now (I changed the AWS CORS rules). I realized my Chrome walks around the CORS policy.

PR #1143, opened 2025-04-03
fix: parsed genomic coordinates shouldn't be NaN even if assembly is not specified

sehilyi:
With this patch, the join transform should work with CSV files.

## Screenshots
Screenshot 2025-04-03 at 5 22 27 PM

Screenshot 2025-04-03 at 5 22 35 PM

### DIFF
--- a/editor/example/json-spec/spatial-layout.ts
+++ b/editor/example/json-spec/spatial-layout.ts
@@ -17,9 +17,11 @@ export const EX_SPEC_SPATIAL_MINIMAL: GoslingSpec = {
             },
             tracks: [
                 {
+                    color: { field: 'coord', type: 'quantitative' },
                     data: {
                         type: 'csv',
-                        url: 'https://gist.githubusercontent.com/sehilyi/29d1cfed56da3ed37370f31a508c8024/raw/9ab733bff25c4c539d86451dfb9d03d4e241d1ad/yeast_model.values.csv'
+                        url: 'https://gist.githubusercontent.com/sehilyi/29d1cfed56da3ed37370f31a508c8024/raw/9ab733bff25c4c539d86451dfb9d03d4e241d1ad/yeast_model.values.csv',
+                        genomicFields: ['position']
                     },
                     width: 500,
                     height: 500
--- a/src/compiler/_normalize-3d-spec.ts
+++ b/src/compiler/_normalize-3d-spec.ts
@@ -18,13 +18,23 @@ export function _fixTrackToWalkaround(t: Track) {
             chr,
             coord
         };
-        const dataTransform: JoinTransform = {
+        const dataTransform: Omit<JoinTransform, 'to'> = {
             type: 'join',
-            from: { url, chromosomeField: chr, genomicField: coord },
-            to: { startField: 'start', endField: 'end' }
+            from: { url, chromosomeField: chr, genomicField: coord }
         };
+        // TODO: need to support other data types as well
+        let to: JoinTransform['to'];
+        if (t.data?.type === 'bigwig') {
+            to = { startField: 'start', endField: 'end' };
+        } else if (t.data?.type === 'csv') {
+            // TODO: Not the reliable way to select two data fields corresponding to start and end positions
+            const startField = t.data.genomicFields?.[0] ?? 'unknown';
+            const endField = t.data.genomicFields?.[1] ?? undefined;
+            to = { startField, endField };
+        }
+
         // @ts-expect-error
-        t.dataTransform = [dataTransform];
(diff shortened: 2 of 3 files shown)


editor/example/json-spec/spatial-layout.ts
```diff
@@ -17,9 +17,11 @@ export const EX_SPEC_SPATIAL_MINIMAL: GoslingSpec = {
             },
             tracks: [
                 {
+                    color: { field: 'coord', type: 'quantitative' },
                     data: {
                         type: 'csv',
-                        url: 'https://gist.githubusercontent.com/sehilyi/29d1cfed56da3ed37370f31a508c8024/raw/9ab733bff25c4c539d86451dfb9d03d4e241d1ad/yeast_model.values.csv'
+                        url: 'https://gist.githubusercontent.com/sehilyi/29d1cfed56da3ed37370f31a508c8024/raw/9ab733bff25c4c539d86451dfb9d03d4e241d1ad/yeast_model.values.csv',
+                        genomicFields: ['position']
```
sehilyi, line 24: The data field defined here is used in the join transform (i.e., check if values of the 'position' field and the 'coord' field are equal).

src/core/utils/data-transform.ts
```diff
@@ -108,23 +107,25 @@ export async function joinData(
         for (const d of data) {
             const chrName = d[from.chromosomeField];
             const chromPosition = d[from.genomicField];
-            d[from.genomicField] = chromSizes.interval[chrName]?.[0] + +chromPosition;
+            d[from.genomicField] = (chromSizes.interval[chrName]?.[0] ?? 0) + +chromPosition;
```
sehilyi, line 110: This was the main reason why the `coord` was `NaN` before. In the minimal example, genome `assembly` is not defined. In this case, Gosling uses the default assembly, which is a human genome (hg38). However, since the chromosome names defined in the 3D genome model (A, B, C, ...) do not match the chromosome names for the human genome (chr1, chr2, chr3, ...), `chromSizes.interval[chrName]?.[0]` is `NaN`.

By adding `?? 0`, the parsed genomic positions (`coord`) are not `NaN` anymore.

src/core/utils/data-transform.ts
```diff
@@ -108,23 +107,25 @@ export async function joinData(
...
-            const start = +t[to.startField] <= +f[from.genomicField];
-            const end = to.endField ? +t[to.endField] >= +f[from.genomicField] : true;
-            return start && end;
+            if (to.endField) {
+                const start = +t[to.startField] <= +f[from.genomicField];
+                const end = +t[to.endField] >= +f[from.genomicField];
+                return start && end;
+            } else {
+                return +t[to.startField] == +f[from.genomicField];
+            }
```
sehilyi, line 126: Fixing an incorrect join operation when `to.endField` is undefined.

src/compiler/_normalize-3d-spec.ts
```diff
@@ -18,13 +18,23 @@ export function _fixTrackToWalkaround(t: Track) {
...
+        const dataTransform: Omit<JoinTransform, 'to'> = {
             type: 'join',
-            from: { url, chromosomeField: chr, genomicField: coord },
-            to: { startField: 'start', endField: 'end' }
+            from: { url, chromosomeField: chr, genomicField: coord }
         };
+        // TODO: need to support other data types as well
+        let to: JoinTransform['to'];
+        if (t.data?.type === 'bigwig') {
+            to = { startField: 'start', endField: 'end' };
```
sehilyi, line 28: bigwig data comes with `start` and `end` data fields. We can just use those fields for join.

src/compiler/_normalize-3d-spec.ts
```diff
@@ -18,13 +18,23 @@ export function _fixTrackToWalkaround(t: Track) {
...
+        } else if (t.data?.type === 'csv') {
+            // TODO: Not the reliable way to select two data fields corresponding to start and end positions
+            const startField = t.data.genomicFields?.[0] ?? 'unknown';
+            const endField = t.data.genomicFields?.[1] ?? undefined;
+            to = { startField, endField };
+        }
```
sehilyi, line 34: Users need to specify what are the genomic fields in the csv file (e.g., `genomicFields`: ['position']). With this change, we are using the first two fields in the join operation.

If only one genomic field is specified in the csv data (e.g., `genomicFields: ['position']`, the `join` operation compares the values of `position` and `coord` to connect the 3D genome model and the CSV data.

If two genomic fields are specified (e.g., `genomicFields: ['start', 'end']`, the `join` operation checks whether `coord` is within the range of 'start' and 'end'.

PR #1147, opened 2025-04-04
feat(editor): example of composing multiple tracks within one spatial view

dvdkouril:
WIP.
Touches the 'join' implementation slightly and is a bit hardcoded at the moment

image

## Checklist


editor/example/json-spec/3d-yeast-model.ts
```diff
@@ -95,12 +95,13 @@ export const EX_SPEC_3D_YEAST_MODEL: GoslingSpec = {
     layout: { ...spatialLayoutDef },
     views: [
         {
-            tracks: [{ ...base, height: trackWidth }]
+            tracks: [{ ...base, mark: "sphere", height: trackWidth }]
```
dvdkouril, line 98: Had to add the explicit mark declaration to override what comes from spreading `base`: there mark is set to `"rect"` which does not have a meaning in a spatial view and therefore doesn't get rendered as anything.

src/compiler/_normalize-3d-spec.ts
```diff
@@ -34,7 +34,7 @@ export function _fixTrackToWalkaround(t: Track) {
         }
 
         // @ts-expect-error
-        t.dataTransform = [{ ...dataTransform, to }];
+        t.dataTransform = [{ ...dataTransform, to }, ...(t.dataTransform ?? [])];
```
dvdkouril, line 37: This change makes sure that the implicit `join` transform doesn't override other transforms defined by the user.

Issue #1161, opened 2025-06-27
Re-implement Gosling API functions in `v2`


sehilyi:
API functions are entirely unimplemented in the `v2` branch. An example function is `zoomTo`, which navigates to a specific genomic range from the JS side. It looks like implementing the missing API functions seems to be the main thing left for releasing `v2`.

The complete list of API functions defined in `v1` can be found in [createApi()](https://github.com/gosling-lang/gosling.js/blob/b7f7f0a065d99c66aee2b87db71e220e18d534ca/src/api/api.ts#L48).

https://github.com/gosling-lang/gosling.js/blob/b7f7f0a065d99c66aee2b87db71e220e18d534ca/src/api/api.ts#L48

The API functions need to be implemented in this new `createAPIV2()` function:

https://github.com/gosling-lang/gosling.js/blob/4133f37205b6d37158bc0e14db34d9a9e711ba1e/src/api/api.ts#L53-L59

### Subscribing Events
- [x] subscribe
- [x] unsubscribe

### Exporting Images
- [x] getCanvas
- [x] exportPng
- [x] exportPdf

### Zoom (#1165)
- [ ] zoomTo
- [ ] zoomToExtent
- [ ] zoomToGene
- [ ] suggestGene

### Retrieving Track/View Information
- [x] getTrackIds
- [x] getTracksAndViews
- [x] getTracks
- [x] getTrack
- [x] getViews
- [x] getView

PR #1168, opened 2025-07-14
feat: re-implement `(un)subscribe` APIs

sehilyi:
Toward #1161 

## Change List
 - Re-implement `(un)subscribe` APIs (The user API remains the same).

## Note
- Since we do not need to use HiGlass' view IDs, we related code can be largely removed and cleaned up, such as https://github.com/gosling-lang/gosling.js/blob/v2/src/api/track-and-view-ids.ts
- After implementing all APIs, it would be good to add tests so that all APIs work as expected.

### DIFF
--- a/demo/gosling-component.tsx
+++ b/demo/gosling-component.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect, type RefObject } from 'react';
+import React, { useState, useEffect, type RefObject, useImperativeHandle } from 'react';
 import { PixiManager } from '@pixi-manager';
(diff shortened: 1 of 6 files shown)


sehilyi:
The `rawData` listener still does not work since `siblingIds` object is somehow always empty:

https://github.com/gosling-lang/gosling.js/blob/72e1194a60af671890428509b74cb7a7e5b61075/src/tracks/gosling-track/gosling-track.ts#L557

This is because we entirely removed the `gosling-to-higlass` conversion logic:

https://github.com/gosling-lang/gosling.js/blob/main/src/compiler/gosling-to-higlass.ts#L106

PR #1170, opened 2025-07-15
feat: support tracks- and views-related APIs

sehilyi:
Toward #1161 

- ~Something to still fix: The track/view objects stored in the `ref` is not up-to-date~

### DIFF
--- a/editor/Editor.tsx
+++ b/editor/Editor.tsx
@@ -14,7 +14,7 @@ import 'allotment/dist/style.css';
 import { debounce, isEqual } from 'lodash-es';
 import stripJsonComments from 'strip-json-comments';
 import JSONCrush from 'jsoncrush';
-import type { Datum } from '@gosling-lang/gosling-schema';
+import type { Datum, VisUnitApiData } from '@gosling-lang/gosling-schema';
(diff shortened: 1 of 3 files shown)


sehilyi:
The bounding boxes of tracks/views obtained through the API seem to be a few pixels off
Screenshot 2025-07-16 at 11 08 57 AM

editor/Editor.tsx
```diff
@@ -1191,14 +1194,23 @@ function Editor(props: RouteComponentProps) {
...
                                                     : 'calc(100% - 50px)',
                                             background: isResponsive ? 'white' : 'none'
                                         }}
                                     >
-                                        <GoslingComponent ref={gosRef} spec={goslingSpec} theme={'light'} />
+                                        <GoslingComponent
+                                            ref={gosRef}
+                                            spec={goslingSpec}
+                                            theme={'light'}
+                                            visualized={() => {
```
sehilyi, line 1209: We need to review if this callback function makes sense. Gosling used to have "compiled()" callback function that did the similar thing.

sehilyi: To track this: #1171

PR #1193, opened 2025-10-07
fix: cursor does not jump as user edits a code, enabling to use the online editor again

sehilyi:
Fix #
Toward #

## Change List
 -

## Checklist


sehilyi:
Hmm, this change makes that the cursor does not jump to the start position instantly, but after I wait for a few secs, it still happens.

PR #1196, opened 2025-10-12
fix: fix the editor issue where a cursor jumps to the top every time a code is updated

### LINKED ISSUE #1195: cursor jumps to the top as I edit a code
It seems like this bug was introduced in https://github.com/gosling-lang/gosling.js/commit/72e1194a60af671890428509b74cb7a7e5b61075 when React version was upgraded from 18 to 19. react-resize-detector was also upgraded accordingly. This bug does not exist in the previous commit.

cc @dvdkouril

sehilyi:
Fix #1195

### DIFF
--- a/editor/EditorPanel.tsx
+++ b/editor/EditorPanel.tsx
@@ -136,40 +136,36 @@ function EditorPanel(props: {
         if (props.onChange) props.onChange(newCode, language);
     }
 
-    const MonacoEditorWithResizer = () => {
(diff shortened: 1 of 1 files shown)


sehilyi:
@dvdkouril Apologies for taking so long to fix this issue. Your clarification that this bug exists in the later v2 helped me track down the source of the issue. This PR should fix the editor.

dvdkouril:
That's awesome, great to see that you found a solution!!

PR #1197, opened 2025-10-15
ci: add proper npm dist-tag (`alpha` or `latest`) when releasing a new version

sehilyi:
This PR updates the current `release` workflow to check the tag name (whether it includes `alpha` or not) and add a proper dist-tag (either `latest` or `alpha`) when publishing a new version to NPM.

cc @dvdkouril


sehilyi:
If this works, we can remove the instruction about running separate `npm dist-tag` commands in the CONTRIBUTING.md of the `v2` branch

